PR #41, opened 2025-03-31
Use trick by @veykril to reduce codegen size

fasterthanlime:
Const contexts _do_ have access to outer generics :)

https://bsky.app/profile/lukaswirth.bsky.social/post/3lloofn4tfk2d

### DIFF
--- a/shapely-codegen/src/main.rs
+++ b/shapely-codegen/src/main.rs
@@ -38,48 +38,53 @@ fn codegen_tuple_impls(w: &mut dyn Write) -> std::fmt::Result {
 
     // Generate implementations for tuples of size 1 to 12
     for n in 1..=12 {
-        let type_params = (1..=n)
+        let type_params = (0..n)
             .map(|i| format!("T{}", i))
             .collect::<Vec<_>>()
             .join(", ");
 
-        let fields = (0..n)
-            .map(|i| {
-                format!(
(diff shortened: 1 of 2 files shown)


spookyvision:
> Key changes include: (....)

what a bunch of noisy nonsense. Nothing in this A"I" "review" discusses the _actual_ trick, instead it just pollutes the discussion with lengthy gibberish.

> Comments suppressed due to low confidence (2)

so why include them. Further distractions.

fasterthanlime:
> Nothing in this A"I" "review" discusses the _actual_ trick,

That's true, but it did mention all the unrelated changes I snuck in, so 🤷

spookyvision:
the fundamental problem of LLMs not knowing what facts are actively misinforms (at best: confuses) people who aren't already domain experts, I think it's really important to keep the bigger picture in mind here, but if the intended audience of those reviews is "you" specifically then that does not apply. However that's (obviously) not made clear by the copilot comment…

fasterthanlime:
> if the intended audience of those reviews is "you" specifically then that does not apply.

Yeah — just to be completely clear, I explicitly "requested a review from Copilot" (not expecting much, tbh) and I'm going to keep doing so in the next few PRs, mostly out of curiosity! I agree it hasn't been _that_ useful so far.

PR #68, opened 2025-04-10
Add facet-toml crate

tversteeg:
Almost identical to `facet-yaml`, using the awesome `toml_edit` crate, which also doesn't add any `syn` dependency!

I would like to further flesh out this crate. Should I do that in new PRs, or make this one a draft and implement it here?

What are your thoughts in general on adding different data formats?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,6 +205,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "equivalent"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
+
+[[package]]
 name = "facet"
 version = "0.1.7"
 dependencies = [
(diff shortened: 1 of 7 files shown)


fasterthanlime:
My thoughts are that every format will need a champion. If you want to be the one for Tommel, that's absolutely amazing. You can open as many PRs as you want, and I'll try to merge them as quickly as possible because nobody's relying on this project now, so we can move fast. 

I'm thinking of moving away from the YAML parser that I'm using right now because building a document object model ahead of time and then mapping it to structs kind of defeats the purpose. I mean, not really, but we can do so much more. I would like to see things like partial deserialization, streaming deserialization, async visualization, and I don't think any of those are well served by existing parsers. Essentially, my thoughts are to get them working, add extra features, and then make them fast with stuff like SIMD, etc. But for now, just do whatever scratches your itch because this is how we will discover the different factors in making adding support for a bunch of different formats. Thank you so much for your work!

fasterthanlime:
Closing in favor of #73 since I just did some breaking refactors — feel free to keep working on top of main after that!

PR #83, opened 2025-04-10
Remove typeid dependency

fasterthanlime:


lonjil:
Is this sound for general user types? If two crates depend on different versions of the same crate, two different types could have the same fully qualified name, and so hash to the same ScalarId. Sorry if I misunderstood and this is in fact only used for standard types, never user types.

fasterthanlime:
> Is this sound for general user types?

Good point, it's not, let's revert this.

Issue #70, opened 2025-04-10
Ownership move semantics of PokeStruct::set_by_name and OpaqueConst


SimonSapin:
Following up on https://tutut.delire.party/@simon/114311828723276583

This snippet in README felt suspicious:

```rust
        let bar = String::from("Hello, World!");
        poke.set_by_name("bar", OpaqueConst::from_ref(&bar))
            .unwrap();
        // bar has been moved out of
        core::mem::forget(bar);
```

`OpaqueConst` is created from a `&String` shared reference, which i wouldn’t expect it’d be legal to move out of.

The same code is found in a test that does not use `unsafe`:

https://github.com/facet-rs/facet/blob/73a46f072c2d2aaa04dd780086bd7fdb30f131f9/facet-poke/tests/poke_test.rs#L26-L54

When commenting out the `mem::forget` line, `just miri` shows use-after-free when `foo_bar.bar` is accessed after `bar` has been dropped:

<details>

```rust
──── STDERR:             facet-poke::poke_test build_foobar_through_reflection
error: Undefined Behavior: out-of-bounds pointer use: alloc127637 has been freed, so this pointer is dangling
   --> /home/simon/.rustup/toolchains/nightly-2025-04-05-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/raw.rs:138:9
    |
138 |         &*ptr::slice_from_raw_parts(data, len)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ out-of-bounds pointer use: alloc127637 has been freed, so this pointer is dangling
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
help: alloc127637 was allocated here:
   --> facet-poke/tests/poke_test.rs:34:19
    |
34  |         let bar = String::from("Hello, World!");
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: alloc127637 was deallocated here:
   --> facet-poke/tests/poke_test.rs:39:5
    |
39  |     }
    |     ^
    = note: BACKTRACE (of the first span) on thread `build_foobar_th`:
[...]

PR #85, opened 2025-04-10
Mark unsafe spez methods as unsafe, closes #82

### LINKED ISSUE #82: Soundness issues: SEGFAULT in safe code
Running this program on my machine with `cargo run --release` will cause the program to try to access memory at address `0x4141414141414141`.

```rust
use facet::Facet;

#[derive(Facet, Debug)]
struct A {
    x: Vec<u8>,
}

struct B(i64, i64, i64);

impl Default for B {
    fn default() -> Self {
        Self(10, 0x4141414141414141, 20)
    }
}

fn main() {
    let mut a1: Vec<u8> = Vec::new();
    let ptr = facet::OpaqueUninit::new(&mut a1);
    facet_spez::SpezDefaultInPlaceYes::spez_default_in_place(&&facet_spez::Spez(B::default()), ptr);

    println!("{}", a1[0]);
}
```

fasterthanlime:


TethysSvensson:
I'm not sure exactly where you want which invariants to be documented, but at the very least the `OpaqueUninit` also needs to point to something that is compatible with writing a `T` to it.

In my example my `B` type had the same size and alignment as ` Vec<u8>`, and reading that safety comment seems to imply that it would be okay to initialize my `B` on top of a `Vec<u8>`.

Additionally you might also use `facet` to get access to internal fields of a struct where changing it might lead to soundness problems in unrelated code. That should probably also be documented somewhere.

Btw, thanks for making facet. It looks **extremely** promising and I'm really hoping it takes off.

fasterthanlime:
> I'm not sure exactly where you want which invariants to be documented, but at the very least the `OpaqueUninit` also needs to point to something that is compatible with writing a `T` to it.
> 
> In my example my `B` type had the same size and alignment as ` Vec<u8>`, and reading that safety comment seems to imply that it would be okay to initialize my `B` on top of a `Vec<u8>`.
> 
> Additionally you might also use `facet` to get access to internal fields of a struct where changing it might lead to soundness problems in unrelated code. That should probably also be documented somewhere.

I agree with all of that. If you feel like opening a pull request to document these, that would be very welcome. I'm busy dealing with things pointed out in #70 — Which I believe are related but not redundant.

fasterthanlime:
You know what, I think opaque, constant, opaque and everything should carry a shape, for added safety.

PR #87, opened 2025-04-10
Replace `ARCHETYPE` const with `SpezEmpty` type wrapper

Veykril:
From what I can tell, the `ARCHETYPE` shenanigans are only needed for trait specializations where no values actually exist / no values are required to be used. Having to construct a (possibly invalid) dummy type seems unnecessary in those cases given we can just create a type-wrapper `Spez` variant (opposed to a value wrapper one).

One other benefit of this is that we do not have to assume the layout of foreign types, like is done for the `RandomState` impl which will cause this library to break from compiling as soon as std decides to change `RandomState`'s layout in some way.

Open to naming suggestions, changelog will be filled out afterwards.


fasterthanlime:
Oh this is absolutely brilliant, thanks so much.

fasterthanlime:
> Ah, I edited a codegen'd file whoops

Right, so you need to edit `tuples_impls.j2` and then rerun `just codegen`. That's not super documented, sorry

Veykril:
> > Ah, I edited a codegen'd file whoops
> 
> Right, so you need to edit `tuples_impls.j2` and then rerun `just codegen`. That's not super documented, sorry

In your defense, I didn't bother reading (or looking for) any of the docs, I just went straight up source diving 😬

PR #92, opened 2025-04-10
Serialize all numbers in JSON

### LINKED ISSUE #84: `facet-json` does not support converting integers other than `u32` to JSON
After poking around a little bit with `facet-json`, I've built this minimal example:

```rust
use facet::Facet;
use facet_peek::Peek;

#[derive(Facet)]
struct Foo {
	bar: String,
	baz: u32,
}

fn main() {
    let foo = Foo {
		bar: "Hello, World!".to_string(),
		baz: 69,
    };
    let foo = Peek::new(&foo);

    println!("{}", facet_json::to_json_string(foo, true));
}
```

However, this returns `"<unsupported type>"` for `u32`.
Looking at the source code of `facet-json-write`, it seems that it only supports `u64` for
number types, while `facet-json-read` supports all number types.

Is there a reason for this? If not, I'd be happy to quickly write a PR to add support for other numerical types in `facet-json-write`.

silvasch:
Previously, `facet-json-write` only supported `u64` for numbers. This PR adds support for all usigned integers, signed integers and floats.

Fixes #84

### DIFF
--- a/facet-json-write/src/serialize.rs
+++ b/facet-json-write/src/serialize.rs
@@ -40,6 +40,12 @@ fn peek_value_to_json<W: Write>(pv: PeekValue, writer: &mut W) -> io::Result<()>
     } else if pv.shape().is_type::<i128>() {
         let value = unsafe { pv.data().as_ref::<i128>() };
         write!(writer, "{}", value)?;
+    } else if pv.shape().is_type::<f32>() {
+        let value = unsafe { pv.data().as_ref::<f32>() };
+        write!(writer, "{}", value)?;
+    } else if pv.shape().is_type::<f64>() {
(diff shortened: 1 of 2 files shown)


silvasch:
@fasterthanlime After merging with the changes from `main`, the tests fail to compile. I'm not sure why, as the errors come from files I did not touch. Not sure if this PR is the right place to fix this.

fasterthanlime:
> @fasterthanlime After merging with the changes from `main`, the tests fail to compile. I'm not sure why, as the errors come from files I did not touch. Not sure if this PR is the right place to fix this.

I merged some more stuff, rebased on top of main, and now we have #93 — so I'm gonna close this, you can continue work in another branch if you have more json contributions?

PR #99, opened 2025-04-10
Generalize `Facet` array impl to arbitrary lengths

Veykril:

PR #101, opened 2025-04-10
Array generic fixup

fasterthanlime:
Just building on #99 

(btw @Veykril — the changelog is maintained by release-plz, no need to edit it manually)


fasterthanlime:
Closing in favor of #99

PR #105, opened 2025-04-10
Fix deserialization of empty JSON arrays

### LINKED ISSUE #79: facet-json doesn't seem to allow reading a struct with an empty array field
```rust
use facet::Facet;

#[derive(Debug, Facet)]
pub struct MyVec {
    pub inner: Vec<u64>,
}

fn main() {
    let v = MyVec { inner: vec![] };
    let peek = facet_peek::Peek::new(&v);
    let json = facet_json::to_json_string(peek, false);
    dbg!(&json);
    let v: MyVec = facet_json::from_str(&json).unwrap();
    dbg!(v);
}
```

Output:
```
[src/main.rs:12:5] &json = "{\"inner\":[]}"

thread 'main' panicked at src/main.rs:13:48:
called `Result::unwrap()` on an `Err` value: JsonParseErrorWithContext { error: JsonParseError { kind: ExpectedNumber, position: 11 }, input: "{\"inner\":[]}" }
```

It seems to work if the vector is non-empty:
```
[src/main.rs:12:5] &json = "{\"inner\":[123]}"
[src/main.rs:14:5] v = MyVec {
    inner: [
        123,
    ],
}
```

tyilo:
Fixes #79

### DIFF
--- a/facet-json-read/src/parser.rs
+++ b/facet-json-read/src/parser.rs
@@ -280,8 +280,7 @@ impl<'a> JsonParser<'a> {
         }
     }
 
-    /// Expects the start of an array and returns whether there's a first element.
-    /// Returns true if the array has an element, false if the array is empty.
+    /// Expects the start of an array.
     pub fn expect_array_start(&mut self) -> Result<(), JsonParseErrorWithContext<'a>> {
         self.skip_whitespace();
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Merged manually, TYSM!

PR #112, opened 2025-04-10
Use core::net instead of std::net

tyilo:


epage:
These lints have helped a lot in my `no_std` code:
```rust
```

Issue #102, opened 2025-04-10
Using `Shape` to export a stable ABI for Rust structs


samscott89:
### Motivation

The motivation here is I'm working on a Rust debugger, and looking into how the debugger can navigate/manipulate Rust types.

Typically, the only way to reconstruct values from their memory addresses is by using debug information output by the compiler (e.g. in DWARF format), e.g. the struct:

```rust
struct TestStruct {
    field: u64,
}
```

might be output in dwarf info as:

```
0x00000d01:     DW_TAG_structure_type
                  DW_AT_name    ("TestStruct")
                  DW_AT_byte_size       (0x08)
                  DW_AT_accessibility   (DW_ACCESS_private)
                  DW_AT_alignment       (8)

0x00000d09:       DW_TAG_member
                    DW_AT_name  ("field")
                    DW_AT_type  (0x00000cc1 "u64")
                    DW_AT_alignment     (8)
                    DW_AT_data_member_location  (0x00)
                    DW_AT_accessibility (DW_ACCESS_private)
```

However, IMO this has two main limitations:

1. Getting information this way requires adding it to the `rustc` and in a format compatible with DWARF -- these both have broader ecosystem implications that limit how much we can do.
2. Possibly as an example of the above -- dwarf info doesn't tend to include anything about implemented traits, unlike how `facet` exposes some vtable information.

A third limitation that's not currently important for my use case: it might be nice to be able to extract this information even when the debug info is stripped. But in that case I'm not sure you would want to export the shapes either.

### Exploration

The rough idea is to export `TestStruct::SHAPE` as a constant that external programs (my debugger) can read that in, and then use that to reconstruct an instance of `TestStruct` (as an `OpaqueConst` that I can manipulate -- e.g. being able to call `Debug::fmt` on an instance of `TestStruct` would be a win.

To do that, I need to:

1. Export the shape info as a constant (easy)

```rust
#[used]
static TEST_STRUCT_SHAPE: &'static Shape = TestStruct::SHAPE;
```

is sufficient to get this out:

```
❯ nm -pa target/debug/simple-test  | rg TEST_STRUCT_SHAPE
000000010004c2a0 s __ZN11simple_test17TEST_STRUCT_SHAPE17h229f985620a94effE
```

2. Reconstruct the `Shape` in the external process

[...]

PR #114, opened 2025-04-10
Add debugging use case to README.

samscott89:
Added details about the use case discussed in #102.

I tried to provide a small concrete example of how this could be used. The "helper method" I used to get this working is a little gnarly, so I figured it was probably better to omit. But for posterity:

```rust
use std::{fmt, mem::transmute};

use facet::{DebugFn, OpaqueConst};

#[unsafe(no_mangle)]
pub extern "C" fn debug_print_object(object: *const (), debug_fn: *const ()) {
    let object = OpaqueConst::new(object);

    let debug_fn: &Option<DebugFn> = unsafe { transmute(debug_fn) };
    let debug_fn = *debug_fn;

    println!("{:#?}", DebugPrinter { object, debug_fn });
}

struct DebugPrinter {
    object: OpaqueConst<'static>,
    debug_fn: Option<DebugFn>,
}

impl fmt::Debug for DebugPrinter {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        if let Some(debug_fn) = self.debug_fn {
            unsafe { debug_fn(self.object, f) }
        } else {
            write!(f, "<unknown>")
        }
    }
}
```

Make sure this is built with:
```
crate-type = ["cdylib"]
```

And the magic lldb invocation to get this working was roughly:

```bash
(lldb) p (int)dlopen("../facet/target/debug/libfacet_rust_abi.dylib", 10)
(int) -1767225520
(lldb)  p "libfacet_rust_abi.dylib";::debug_print_object(&STATIC_TEST_STRUCT, &TEST_STRUCT_SHAPE->vtable->debug)
TestStruct {
    string_field: "some field I would like to see",
    int_field: 66,
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,83 @@ test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; fini
      Summary [   0.005s] 1 test run: 1 passed, 66 skipped
 ```
 
+## Use cases: augmenting static analysis/debuggers
+
+By default the `Facet` derive macro creates and exports
+a global static variable `{UPPER_CASE_NAME}_SHAPE` referencing the
+`Shape` of the derived `Facet` trait.
+
+Furthermore, `Shape` and all nested fields are `#[repr(C)]`.
+
+This information can be used by external processes (like debuggers) to access the
+layout and vtable data.
+
+For example, suppose we have:
+
+```rust,ignore
+#[derive(Debug, Facet)]
+struct TestStruct {
+    field: &'static str,
+}
+
+static STATIC_TEST_STRUCT: TestStruct = TestStruct {
+    field: "some field I would like to see",
+};
+```
+
+By default, printing this in `lldb` returns the lengthy:
+
+```bash
+(lldb) p STATIC_TEST_STRUCT
+(simple_test::TestStruct) {
+  field = "some field I would like to see" {
+    [0] = 's'
+    [1] = 'o'
+    [2] = 'm'
+    [3] = 'e'
+    [4] = ' '
+    [5] = 'f'
+    [6] = 'i'
+    [7] = 'e'
+    [8] = 'l'
+    [9] = 'd'
+
+  ... (and so on)
+}
+```
+
+However, the `TestStruct::SHAPE` constant is available at `TEST_STRUCT_SHAPE`:
+
+```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -292,6 +292,83 @@ test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; fini
...
+    [6] = 'i'
+    [7] = 'e'
+    [8] = 'l'
+    [9] = 'd'
+
+  ... (and so on)
+}
+```
+
+However, the `TestStruct::SHAPE` constant is available at `TEST_STRUCT_SHAPE`:
```
samscott89, line 340: A small detail I elided for brevity: currently the symbol is still mangled and is actually something like:
```
__ZN11simple_test17TEST_STRUCT_SHAPE17h5e18cd1cf07f11deE
```

However, `lldb` is smart enough to figure this out, and I figure most people who would have this use case could also figure that out.

README.md
```diff
@@ -292,6 +292,83 @@ test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; fini
...
+without needing external methods (but with some additional shenanigans to make `lldb` happy):
+
+```bash
+(lldb) p TEST_STRUCT_SHAPE->vtable->eq
+(core::option::Option<unsafe fn(facet_core::opaque::OpaqueConst, facet_core::opaque::OpaqueConst) -> bool>) {
+  value = {
+    0 = 0x0000000100002538
+  }
+}
+(lldb) p (*((bool (**)(simple_test::TestStruct* , simple_test::TestStruct*))(&TEST_STRUCT_SHAPE->vtable->eq)))(&STATIC_TEST_STRUCT, &STATIC_TEST_STRUCT)
```
samscott89, line 368: Ironically this was to make lldb happy with the type conversion. I think the `Option` around the function pointer throws it a little.

fasterthanlime:
This is fantastic, thanks so much!

You'll want to update the templates in `templates/README.md.j2` for every readme you want to change (the top-level one is copied to `facet`) and run `just codegen`.

Running `just` pretty much does CI on your machine, so there's no surprise when you push :)

fasterthanlime:
Merged manually, thanks so much!

PR #115, opened 2025-04-11
TOML scalar values

tversteeg:
I've implemented most scalar values also implemented by `scalar_impls`.

What do you think about this implementation using `num_traits`? Do you like this, or would you rather have me implement it with a macro generating a function for each type?


fasterthanlime:
> What do you think about this implementation using `num_traits`?

Let's say that it's fine for now because it works, and if we feel like shedding dependencies later, then we can always try to find a way to not depend on it. But I think right now it makes sense.

PR #117, opened 2025-04-11
feat(types): create exhaustive exact scalar type enum

tversteeg:
I came upon this idea when implementing the TOML scalar type parser.
I found that I had a huge list of `.. else if ps.shape().is_type::<$$type$$>() ..` checking all different scalar types.
This creates the problem that it's very hard to verify which type is covered by a parser.

I created an exhaustive `ScalarType` enum with all exact scalar implementations in facet.
This allows someone to match on it when implementing a parser to ensure that all cases are covered.
I created a small example in `facet-yaml/src/lib.rs` to see what this would look like in practice.

I'm not entirely sure about the implementation in the `ScalarDef::builder` since it now overwrites the affinity type.


fasterthanlime:
I think the idea is that at some point we will release 1.0 and after that I would like to avoid breaking changes. So I don't think anything exhaustive is a good idea. However, in peek or poke, for example, I could definitely see some more enums that you can easily match on. Even if you have to have some sort of fallback — I think dealing with everything that looks like a number that fits in a U64, for example, belongs in peek and poke and not individual serializers and deserializers. 

I'm going to be looking at peek and poke more in depth because I believe they don't necessarily need to have many unsafe functions. I don't think deserializers necessarily need to call unsafe interfaces.

PR #118, opened 2025-04-11
Allow json serialization of static string types

tristan:
I understand that supporting references properly is still something to be figured out, but it seemed to me that supporting `'static` references (especially for string-like things) should be possible.

This is a "hammer it until it fits without really understanding the `shape`" implementation, so I wont take offense if I'm missing the mark completely, or if this just gets in the way of planned work.

I think on the surface level, it's reasonably safe to support even without proper ref support, as you still cannot `#[derive(Facet)]` on a type that has lifetime annotations (e.g. `struct Struct<'a>(&'a str)`).

### DIFF
--- a/facet-json-write/src/serialize.rs
+++ b/facet-json-write/src/serialize.rs
@@ -40,6 +40,25 @@ macro_rules! ints {
     };
 }
 
+macro_rules! string {
+    ($pv:expr, $writer:expr, $type:ty) => {
+        if $pv.shape().is_type::<$type>() {
+            let value = unsafe { $pv.data().as_ref::<$type>() };
+            return write!($writer, "\"{}\"", value.escape_debug());
+        }
+    };
+}
+
+macro_rules! strings {
+    ($pv:expr, $writer:expr, $type:ty, $($types:ty),*) => {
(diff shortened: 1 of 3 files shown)


facet-json-write/src/serialize.rs
```diff
@@ -182,6 +198,13 @@ pub fn to_json<W: Write>(peek: Peek<'_>, writer: &mut W, indent: bool) -> io::Re
                             });
                         }
                     }
+                    Peek::Option(popt) => {
+                        if let Some(p) = popt.value() {
+                            peek_value_to_json(p.as_value(), writer)?;
+                        } else {
+                            write!(writer, "null")?;
+                        }
+                    }
```
tristan, line 207: I also snuck in turning optionals into `null` in json-write, as i think it makes sense (but happy to remove it). I also kinda wish I would have called `peek_value_to_json((), writer)?;` to not have `write!(writer, "null")?;` in two places, but it wasn't immediately obvious how to construct a `PeekValue` for `()`, and I got lazy.

fasterthanlime:
Thanks, I think this is a good start!

PR #123, opened 2025-04-11
Fix `no_std` support

epage:
Quick summary
- `facet-core` only worked with `alloc` and not `core` because of `extern crate alloc`
- Some libraries were missing `no_std`
- Some std types were being used through the prelude
- `facet-poke` can support `alloc` without `std`
  - I considered adding an `alloc` feature and a `#[cfg(not(feature = "alloc"))] compile_error!("")` but (1) that would be a breaking change and (2) unsure if that future proofing is worth it
  - I considered doing this to other crates as well, like `facet-json` but (1) that would be a breaking change and (2) was unsure how far to go at this stage of facet

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -29,6 +29,9 @@ nostd:
     cargo check --no-default-features -p facet
     cargo check --no-default-features -p facet-peek
     cargo check --no-default-features -p facet-poke
+    cargo check --no-default-features --features alloc -p facet-core
+    cargo check --no-default-features --features alloc -p facet
+    cargo check --no-default-features --features alloc -p facet-peek
 
 ci:
     #!/usr/bin/env -S bash -euo pipefail
--- a/facet-core/Cargo.toml
+++ b/facet-core/Cargo.toml
@@ -15,5 +15,6 @@ impls = "1.0.3"
 bitflags = "2.4.1"
 
 [features]
-std = []
+std = ["alloc"]
+alloc = []
 default = ["std"]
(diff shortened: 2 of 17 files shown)


fasterthanlime:
> * but (1) that would be a breaking change and (2) unsure if that future proofing is worth it

I would be in favor of that. 

> * considered doing this to other crates as well, like `facet-json` but (1) that would be a breaking change and (2) was unsure how far to go at this stage of facet

Honestly, it seems worth it if you have the time!

fasterthanlime (approved):

epage:
> Honestly, it seems worth it if you have the time!

As you use workspace inheritance for deps and that is [busted](https://github.com/rust-lang/cargo/issues/12162) for `default-features = false`, I'm going to hold off

epage:
> As you use workspace inheritance for deps and that is https://github.com/rust-lang/cargo/issues/12162 for default-features = false, I'm going to hold off

Oh hey, that means `no_std` is still broken because `facet`, `facet-poke`, and `facet-peek` enable default features on `facet-core`

fasterthanlime:
> Oh hey, that means `no_std` is still broken because `facet`, `facet-poke`, and `facet-peek` enable default features on `facet-core`

Alright, let me get that fixed up real quick.

fasterthanlime:
> Alright, let me get that fixed up real quick.

Done in e0e1b8d61d8620d327484949608b6ca1151b8c28

PR #125, opened 2025-04-11
Do not attempt to parse field types, that's not really our job

Veykril:
We don't introspect the types in the macro, so there is little reason to reconstruct parsers for them all, we just need to skip past them to parse the following relevant parts.

With this we should automatically support any type specifications in fields

### DIFF
--- a/facet-derive/src/lib.rs
+++ b/facet-derive/src/lib.rs
@@ -13,6 +13,7 @@ keyword! {
     KDoc = "doc";
     KRepr = "repr";
     KCrate = "crate";
+    KIn = "in";
     KConst = "const";
     KMut = "mut";
     KFacet = "facet";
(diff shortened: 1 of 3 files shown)


fasterthanlime:
I agree with the field types. I think as for visibility, I thought it might be interesting to have it. Because we can parse it.

fasterthanlime:
I _just_ pushed something else but I can resolve those conflicts for you if you want x)

Also, we can worry about visibilities later fwiw — I didn't mean to sign you up for more work.

fasterthanlime:
very complex type used

Honestly clippy has a point there — unsynn makes it easy to just use another struct to parse nested stuff.

(Also, you can run pretty much the same thing as the CI by just doing `just && just miri` in the workspace, just so you know)

Veykril:
Gotcha, I was only running `just test` locally

Issue #69, opened 2025-04-10
facet-derive does not support (type, lifetime, const) parameters


SimonSapin:
It seems that parsing for structs definitions with parameters is not implemented:
```rust
#[derive(Facet)]
struct Example1<T>(T);

#[derive(Facet)]
struct Example2<'a>(&'a str);

#[derive(Facet)]
struct Example3<const N: usize>([u8; N]);
```

My reading of the error message: `expected braced fields, found '<'`

```rust
proc-macro panicked: Could not parse input as struct, tuple struct, or enum: struct Example1 < T > (T) ;
====> Error parsing struct: Some(Unexpected token: expected unsynn::group::BraceGroupContaining<unsynn::container::DelimitedVec<facet_derive::StructField, unsynn::punct::Operator<','>>>, found Punct { char: '<', spacing: Alone, span: SpanData { range: 32..33, anchor: SpanAnchor(EditionedFileId(FileId(0), Edition2024), 9), ctx: SyntaxContextId(4294967036) } } at LineColumn { line: 0, column: 0 })
====> Remaining tokens after struct parsing: struct Example1 < T > (T) ;

proc-macro panicked: Could not parse input as struct, tuple struct, or enum: struct Example2 <'a > (&'a str) ;
====> Error parsing struct: Some(Unexpected token: expected unsynn::group::BraceGroupContaining<unsynn::container::DelimitedVec<facet_derive::StructField, unsynn::punct::Operator<','>>>, found Punct { char: '<', spacing: Joint, span: SpanData { range: 32..33, anchor: SpanAnchor(EditionedFileId(FileId(0), Edition2024), 9), ctx: SyntaxContextId(4294967036) } } at LineColumn { line: 0, column: 0 })
====> Remaining tokens after struct parsing: struct Example2 <'a > (&'a str) ;

proc-macro panicked: Could not parse input as struct, tuple struct, or enum: struct Example3 < const N : usize > ([u8 ; N]) ;
====> Error parsing struct: Some(Unexpected token: expected unsynn::group::BraceGroupContaining<unsynn::container::DelimitedVec<facet_derive::StructField, unsynn::punct::Operator<','>>>, found Punct { char: '<', spacing: Alone, span: SpanData { range: 32..33, anchor: SpanAnchor(EditionedFileId(FileId(0), Edition2024), 9), ctx: SyntaxContextId(4294967036) } } at LineColumn { line: 0, column: 0 })
[...]

PR #130, opened 2025-04-11
feat(facet-derive): Support generic ADTs

Veykril:
Closes https://github.com/facet-rs/facet/issues/69

### DIFF
--- a/facet/tests/derive.rs
+++ b/facet/tests/derive.rs
@@ -289,6 +289,70 @@ fn tuple_struct_field_doc_comment() {
 }
 
 #[test]
+fn record_struct_generic() {
+    #[derive(Clone, Hash, PartialEq, Eq, ::facet::Facet)]
+    struct Blah<'a, T: Facet, const C: usize = 3>
+    where
+        T: core::hash::Hash,
+    {
+        field: core::marker::PhantomData<&'a T>,
+    }
+}
+
+#[test]
+fn tuple_struct_generic() {
+    #[derive(Clone, Hash, PartialEq, Eq, ::facet::Facet)]
+    #[repr(transparent)]
+    struct Blah<'a, T: Facet, const C: usize = 3>(T, core::marker::PhantomData<&'a ()>)
+    where
+        T: core::hash::Hash;
+}
+
+#[test]
+fn unit_struct_generic() {
+    #[derive(Clone, Hash, PartialEq, Eq, ::facet::Facet)]
+    struct Blah<const C: usize = 3>
+    where
+        (): core::hash::Hash;
+}
+
+#[test]
+fn enum_generic() {
+    #[allow(dead_code)]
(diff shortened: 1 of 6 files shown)


facet/tests/derive.rs
```diff
@@ -288,6 +288,13 @@ fn tuple_struct_field_doc_comment() {
     );
 }
 
+#[test]
+fn tuple_struct_generic() {
+    #[derive(Clone, Hash, PartialEq, Eq, ::facet::Facet)]
+    #[repr(transparent)]
+    struct Blah<'a, T: Facet, const C: usize = 3>(T, core::marker::PhantomData<&'a ()>);
```
Veykril, line 295: Fwiw I am not sure whether we want to support lifetimes. I am pretty sure `Facet` implementations on anything involving lifetimes are unsound as it likely allows transmuting lifetimes. 

That is I am already worrying about the impls we have for `&str` and `Cow`.

fasterthanlime: > but I imagine we are likely exposing some safe ones that do type checks under the hood at some point and I believe the moment we do we expose lifetime transmutation in safe code.

We certainly need to find a way to do it. I haven't been able to give it much thought yet.

fasterthanlime: Essentially, here's what I'm thinking. `Peek` and `Poke` have exactly one lifetime. They should be able to manipulate anything that has exactly that lifetime, and that's it, in safe code. Ideally, serializers and deserializers should not have to write unsafe code. 

We should deal with a maximum of one lifetime and not be able to change it at all, and do manual checking, I suppose? It might take us a few tries to find something that works, but I don't want to give up that easily.

Veykril: Right, using an invariant lifetime might work out

fasterthanlime: > using an invariant lifetime might work out

Yeah that's my only goal I think.

facet/tests/derive.rs
```diff
@@ -288,6 +288,70 @@ fn tuple_struct_field_doc_comment() {
...
+//         T: core::hash::Hash,
+//     {
+//         Unit,
+//         Tuple(i32),
+//         Record {
+//             field: T,
+//             phantom: core::marker::PhantomData<&'a ()>,
+//         },
+//     }
+// }
```
Veykril, line 353: There is one limitation right now which is that the generated shadow structs assume all generics are used for each variant, so this currently fails to compile. We'll likely need to filter the generics and where clauses by usages of them within the variant fields. But that's kind of complicated and only syntactic which means it wont even work all the time when more macros are involved.
So I am leaving that out for now

fasterthanlime:
CleanShot 2025-04-11 at 22 21 39@2x

I'm glad the new CI setup is actually being useful here :)

PR #139, opened 2025-04-11
Enums: support only primitive repr and make derive stricter.

samscott89:
The `Default` enum representation is not correct/sound.

As a simple test:
- take the existing `build_enum` test and changing it to using `repr(C)`
- realize the derive macro is broken for `repr(C)`
- after fixing that -- the test fails with `SIGSEGV`

The current approach attempts to guess the discriminant size using the number of variants -- a reasonable idea, but it turns out `#[repr(C)]` enums are based on compiler/platform specifics: https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.c.enum

For now I figured it would be safer to (a) remove this unsafe Default repr and (b) make the derive stricter.

I'll open an issue for whether enum support is feasible in general.


fasterthanlime:
> I'll open an issue for whether enum support is feasible in general.

I've pinged the compiler team about this btw — to try and see if we can get `offset_of` or discriminant stuff stabilized.

Issue #140, opened 2025-04-11
Feasibility of supporting `#[repr(C)]` or `#[repr(Rust)]` enums


samscott89:
Currently `EnumDef` requires an explicit primitive representation for enums to work, and this is enforced by the derive macro (or at least, will be with #139).

Supporting either `#[repr(C)]` or `#[repr(Rust)]` (the default representation) would require understanding the layout of the enum, which is non-trivial.

In the case of `#[repr(C)]`, it might be feasible to do a similar trick with using the shadow struct for the enum variants, but _also_ build a shadow struct for the tag, and use the two to compute the layout of the resultant struct as described here: https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.c.adt

For `#[repr(Rust)]` doing this layout matching seems entirely infeasible. There are the various "niche" optimizations for enums (nice post [here](https://jpfennell.com/posts/enum-type-size/)), and there are no `offset_of` equivalents for enum variants that can get us this information.

I'm not sure if there are any tricks out there to make this work?

(Note: this is not something I'm personally blocked on, just noting it as a potential future issue).

fasterthanlime:
> I'm going to pause on this here though unless there's actually a reasonable desire for this as a feature.

I am definitely interested in that feature!

samscott89:
Updated title + description to scope this to #[repr(Rust)]. Probably blocked on some combination of `size_of` for discriminators and `offset_of` for enum variants.

A-Walrus:
> Probably blocked on some combination of `size_of` for discriminators and `offset_of` for enum variants.

I'm not sure even those would be enough, wouldn't we need to be able to know what the discriminator for each variant is, if and where there is a niche, etc?

samscott89:
Yeah, that sounds right to me.

fee1-dead:
Landed upon this issue because I looked into ways to walk fields of a user ADT and was wondering whether facet would support that (cc #685). (I could also just use [macro-rules-attribute](https://github.com/danielhenrymantilla/macro_rules_attribute-rs) instead, but this might be more fun)

It seems like the current design isn't well-equipped enough to support this, at least not without the `offset_of_enum` [feature](https://github.com/rust-lang/rust/issues/120141) plus the Unsafe Set Enum Discriminants [RFC](https://github.com/rust-lang/rfcs/pull/3727).

[...]

PR #157, opened 2025-04-11
Derive Facet for #[repr(C)] enums

samscott89:
So this was working, and I just merged in the generics stuff but seems like the new facet-reflect is also a bit of a conflict. So I'm going to open this while I resolve in case anything else comes up.

(edit to add: updated)


facet-poke/tests/enum.rs
```diff
@@ -97,3 +97,181 @@ fn build_enum() {
         );
     }
 }
+
+#[test]
```
samscott89, line 101: Both these tests are a full on copy-paste.

I could macroify these if that would be preferable?

facet-peek/src/enum_.rs
```diff
@@ -138,19 +138,20 @@ impl<'mem> PeekEnum<'mem> {
...
         let variant = self.active_variant();
+        let variant_offset = variant.offset;
 
         match &variant.kind {
             VariantKind::Unit => None, // Unit variants have no fields
             VariantKind::Tuple { fields } => {
                 // For tuple variants, find by name
                 let field = fields.iter().find(|f| f.name == field_name)?;
-                let field_data = unsafe { self.value.data().field(field.offset) };
+                let field_data = unsafe { self.value.data().field(variant_offset + field.offset) };
```
samscott89, line 148: I don't love this -- super easy to miss the variant offset and create unsoundness. But it seemed consistent with what there is today.

I _think_ it would be nice if selecting a variant shifted the `self.data` offset appropriately perhaps?

samscott89:
I should say: partially resolves #140

facet-poke/tests/enum.rs
samscott89: Well, in the latest main this test is entirely deleted so I guess that resolves my dilemma

facet-peek/src/enum_.rs
samscott89: Pushed https://github.com/facet-rs/facet/pull/157/commits/c4be0ef988b24af13d6d534bdff18081aa891a58 which I think is an improvement and formalises a little what's going on here -- all our enum variant field offsets are computed relative to the start of the "variant data" section. For `#[repr(u*)]` that variant data offset is 0 and each variant basically adds padding for the discriminant. For `#[repr(C)]` the variant data offset should be fixed and non-zero, but fields in a variant can start from zero.

fasterthanlime:
> So this was working, and I just merged in the generics stuff but seems like the new facet-reflect is also a bit of a conflict. So I'm going to open this while I resolve in case anything else comes up.

Yep sorry, things are moving fast! I'm happy to take on the merge conflict resolution if you want.

samscott89:
> > So this was working, and I just merged in the generics stuff but seems like the new facet-reflect is also a bit of a conflict. So I'm going to open this while I resolve in case anything else comes up.
> 
> Yep sorry, things are moving fast! I'm happy to take on the merge conflict resolution if you want.

It's all good! Should be resolved now. And gave me the excuse to take on a couple of small refactors to make future changes easier

samscott89:
unless more changes come up in which case I will take you up on that...

fasterthanlime:
> unless more changes come up in which case I will take you up on that...

On it!

samscott89:
Also, the `miri` failure seems new -- is that a new test added?

fasterthanlime:
Closing in favor of #163

> Also, the `miri` failure seems new -- is that a new test added?

Nope, apparently installing `color-backtrace` under miri is a bad idea, especially with the btparse backend. I'll make miri required as well (and all the tests, actually)

PR #162, opened 2025-04-11
[with discussion] scalars: add more types, and add info to numbers and floats

niacdoial:
I have a small question about the scalar types in general:
are you only interested in things that "make sense" to a CPU of some sort? or do you include more ...indirect types that still count as scalar on a config file?

specifically, here are a bunch of types, in decreasing amounts of how likely they are to be useful for this project (independently of how hard they would be to describe the layout of)
- complex numbers (I think there are standard complex types in the cstdlib)
- decimal numbers
- datetime/timestamps 
- UUIDs
- fractions, algebraic number types (see https://chadnauseam.com/coding/random/calculator-app)

...also I just realised that the epsilon might not be necessary, since it is in theory computable as $2^{1-\text{MantissaDigits}}$ (MantissaDigits including any implicit leading digit)

### DIFF
--- a/facet-core/src/impls_core/scalar.rs
+++ b/facet-core/src/impls_core/scalar.rs
@@ -406,6 +406,7 @@ static NEGATIVE_INFINITY_F32: f32 = f32::NEG_INFINITY;
 static NAN_F32: f32 = f32::NAN;
 static POSITIVE_ZERO_F32: f32 = 0.0f32;
 static NEGATIVE_ZERO_F32: f32 = -0.0f32;
+static EPSILON_F32: f32 = f32::EPSILON;
 
 // Constants for f64
 static MIN_F64: f64 = f64::MIN;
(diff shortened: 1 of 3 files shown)


fasterthanlime:
> specifically, here are a bunch of types, in decreasing amounts of how likely they are to be useful for this project (independently of how hard they would be to describe the layout of)

I think those could all be useful — datetime/timestamp definitely want, yeah.

I think anything that could have some special handling for serialization, deserialization, a debugger or anything is a good scalar enum variant. Heck, URL should probably be one (or is it URI?)

niacdoial:
and here!
I added a bunch of scalar types (just... not URL, oops. IIRC, URL is a subclass of URI? eh.)
There are two TODOs in those additions you might want to pay attention to and/or bikeshed a little

as for NumberAffinity::epsilon, in the end I think it's better to keep it since it's computed differently for decimals and floats.

niacdoial:
ah ok
(... due to a skill issue on my part, I failed to receive the notification email until this evening)
the way I see it I can either pull it out of the draft now, or try to implement something to wrap around std::'s datetime and systemtime types

I'll do the former, in case you think it's ready to be merged even without this

as for the renamed file, a rebase caught that (somehow? not sure how expected this is)

fasterthanlime:
There's still merge conflicts unfortunately — 

CleanShot 2025-04-14 at 21 29 48@2x

Ping me when it's resolved :)

niacdoial:
> Ping me when it's resolved :)

...
actually reading instructions: 1
me: 0

\*ahem\* @fasterthanlime ping

fasterthanlime (approved):

Issue #148, opened 2025-04-11
Remove dependency toml_edit


fasterthanlime:
We don't want to build entire DOM trees / data structures before deserializing, we want to be parsing directly.

epage:
Some warnings:
- Parsing TOML *requires* that you parse into a DOM of some form for full compliance
  - Keys can come in any order (e.g. `foo.bar.baz = "a"\nhello.world = "10"\nfoo.bar.bob = "b"`)
  - You have to track what keys are explicit vs implicit and error only if two explicit keys are redundant but not two implicit or an implicit and an explicit (e.g. `foo = { bar = { baz = "a" } }\nfoo.bar.bob = "b"` should error iirc)
- I know of four people who have abandoned maintaining a TOML parser, two within the last year.  As far as I know, `toml_edit` is the only one left.

If you decide to write one, I maintain a helper for executing the toml compliance tests, see https://crates.io/crates/toml-test

Its a *very* low priority but I am experiment with writing a new TOML parser.  My hope is that `toml` will depend on it directly and not through `toml_edit` (which will also depend on it).  Its push-parsing / event generation puts very few constraints on callers.  You can see it at https://github.com/epage/toml_edit/tree/parser/crates/toml_parse

I also hope to have a TOML writer like [`json_writer`](https://docs.rs/json-writer/0.4.0/json_writer/) so people don't have to rewrite string escaping rules, deal with "pretty string" detection (and getting it wrong), writing for TOML 1.0 vs 1.1, etc)

tversteeg:
I think it's a good idea to continue implementing the TOML deserializer and later serializer using `toml_edit` first. That way, we can have a fully functioning implementation in a reasonable time-frame with proper tests, allowing the underlying system to be changed and optimized later while still being a functional implementation for the main case of parsing a type from a string.

epage:
FYI toml-rs/toml#856 pulls out of `toml_edit` the minimum logic necessary to write conformant TOML (string escaping, selection of string styles, etc).

epage:
I've now merged `toml_edit` using `toml_parse`.  I've not yet released `toml_parse` but I can do that if someone needs it sooner than later.

Likely the first step is forking the following tests from `toml`
- https://github.com/toml-rs/toml/tree/main/crates/toml/tests/compliance
- https://github.com/toml-rs/toml/tree/main/crates/toml/tests/serde
[...]

PR #168, opened 2025-04-12
TOML enum with unit variant implementation

tversteeg:
I also added the `PokeEnumNoVarient::variant_by_name` and  `PokeEnumNoVarient::contains_variant_with_name` methods to make life a bit easier.

### DIFF
--- a/facet-reflect/src/poke/enum_.rs
+++ b/facet-reflect/src/poke/enum_.rs
@@ -4,7 +4,9 @@ extern crate alloc;
 #[cfg(feature = "alloc")]
 use alloc::boxed::Box;
 
-use facet_core::{EnumDef, EnumRepr, Facet, FieldError, Opaque, OpaqueUninit, Shape, VariantKind};
+use facet_core::{
+    EnumDef, EnumRepr, Facet, FieldError, Opaque, OpaqueUninit, Shape, Variant, VariantKind,
+};
 
 use crate::Guard;
 
@@ -29,6 +31,7 @@ impl<'mem> PokeEnumNoVariant<'mem> {
     pub fn shape(&self) -> &'static Shape {
         self.shape
     }
+
     /// Creates a new PokeEnumNoVariant from raw data
     ///
     /// # Safety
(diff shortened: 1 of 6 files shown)


fasterthanlime:
Btw idk if you've seen #148 but don't stress about it — according to @epage, TOML is one of the more difficult formats to parse — if we remove dependency on `toml_edit` maybe we'll just switch to the more barebones toml parser instead.

facet-toml/src/lib.rs
```diff
@@ -20,175 +22,198 @@ pub fn from_str<T: Facet>(toml: &str) -> Result<T, AnyErr> {
...
+        poke.put(to_scalar::number::<isize>(item)?)
+    } else if shape.is_type::<i128>() {
+        poke.put(to_scalar::number::<i128>(item)?)
+    } else if shape.is_type::<i64>() {
+        poke.put(to_scalar::number::<i64>(item)?)
+    } else if shape.is_type::<i32>() {
+        poke.put(to_scalar::number::<i32>(item)?)
+    } else if shape.is_type::<i16>() {
+        poke.put(to_scalar::number::<i16>(item)?)
+    } else if shape.is_type::<i8>() {
```
fasterthanlime, line 145: Looking at this, I'm thinking that we need methods in PeekValue that can give a non-exhaustive enum. I think the approach was okay, but it just needs to be non-exhaustive. Maybe we need fewer types, like signed and unsigned 64-bit numbers. I guess not many people care about 128-bit-wide numbers. The rest can be bignum, I suppose? And the non-zero stuff, we can actually look inside at the T. We can definitely get rid of a lot of these cases.

facet-reflect/src/poke/enum_.rs
```diff
@@ -146,6 +149,19 @@ impl<'mem> PokeEnumNoVariant<'mem> {
...
         })
     }
+
+    /// Try to find the variant by name.
+    pub fn variant_by_name(&self, name: &str) -> Option<&Variant> {
+        self.def
+            .variants
+            .iter()
+            .find(|variant| variant.name == name)
+    }
```
fasterthanlime, line 159: I think ideally the `facet-reflect` API will evolve to have these methods return "Slots" that can be set.

That's what I was doing initially, but it took me a while to figure out what the right shape of that would be. No pun intended.

facet-toml/src/lib.rs
tversteeg: Do you want it in facet-core like in #117, or do you want an enum that's exclusive to facet-reflect?

fasterthanlime: > or do you want an enum that's exclusive to facet-reflect?

I think that one makes the most sense. It should be on PeekValue and PokeValue — facet-core is lower-level.

PR #173, opened 2025-04-12
feat(reflect): add `ScalarType` enum

tversteeg:
I hope this is what you had in mind.

I removed all `NonZero` and `(i|u)128` variants.

### DIFF
--- a/facet-reflect/src/peek/value.rs
+++ b/facet-reflect/src/peek/value.rs
@@ -8,7 +8,6 @@ use crate::{Peek, ScalarType};
 pub struct PeekValue<'mem> {
     data: OpaqueConst<'mem>,
     shape: &'static Shape,
-    scalar_type: Option<ScalarType>,
 }
 
 impl core::fmt::Display for PeekValue<'_> {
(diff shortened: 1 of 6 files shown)


facet-reflect/src/scalar.rs
```diff
@@ -0,0 +1,217 @@
...
+    Ipv4Addr,
+    /// `core::net::Ipv6Addr`.
+    Ipv6Addr,
+    /// `facet_core::typeid::ConstTypeId`.
+    ConstTypeId,
+}
+
+impl ScalarType {
+    /// Infer the type from a shape definition.
+    pub fn try_from_shape(shape: &'static Shape) -> Option<Self> {
```
tversteeg, line 60: Should this be an `Option<Self>` or panic if it's not a scalar?

If it should panic, it would also mean that it's practically exhaustive.

facet-reflect/src/peek/value.rs
```diff
@@ -61,7 +62,13 @@ impl<'mem> PeekValue<'mem> {
     /// `data` must be initialized and well-aligned, and point to a value
     /// of the type described by `shape`.
     pub(crate) unsafe fn unchecked_new(data: OpaqueConst<'mem>, shape: &'static Shape) -> Self {
-        Self { data, shape }
+        let scalar_type = ScalarType::try_from_shape(shape);
```
tversteeg, line 65: Should I calculate the type here or in the getter itself?

Doing it here means it will be calculated at most once, but also when it's not used.
Doing it in `scalar_type` means it will be calculated every time it's called, but not when it's used.

fasterthanlime: I vote "in the getter"

facet-reflect/src/scalar.rs
fasterthanlime: Option is the right call!

tversteeg:
@fasterthanlime missed one implementation, it can be merged now!

PR #184, opened 2025-04-12
Implement Facet for BTreeMap + minor simplifications for HashMap

simonask:
Implementation based on the one for HashMap.

Not sure where to add tests. 🤔

### DIFF
--- a/facet-core/src/_trait/impls/btree_impl.rs
+++ b/facet-core/src/_trait/impls/btree_impl.rs
@@ -1,0 +1,195 @@
+use core::alloc::Layout;
+
+use alloc::collections::{BTreeMap, VecDeque};
+
+use crate::{
+    ConstTypeId, Def, Facet, MapDef, MapIterVTable, MapVTable, MarkerTraits, Opaque, OpaqueConst,
+    Shape, ValueVTable,
+};
+
+struct BTreeMapIterator<'mem, K> {
+    map: OpaqueConst<'mem>,
+    keys: VecDeque<&'mem K>,
+}
+
+unsafe impl<K, V> Facet for BTreeMap<K, V>
+where
+    K: Facet + core::cmp::Eq + core::cmp::Ord + 'static,
+    V: Facet + 'static,
+{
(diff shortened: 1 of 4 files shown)


facet-core/src/_trait/impls/btree_impl.rs
```diff
@@ -0,0 +1,198 @@
...
+                            write!(f, "}}")
+                        })
+                    }
+
+                    builder =
+                        builder.default_in_place(|target| unsafe { target.put(Self::default()) });
+                    builder = builder
+                        .clone_into(|src, dst| unsafe { dst.put(src.as_ref::<BTreeMap<K, V>>()) });
+
+                    if K::SHAPE.vtable.eq.is_some() && V::SHAPE.vtable.eq.is_some() {
```
simonask, line 85: The impl for HashMap does this, but I guess there's not really any reason to go through the type-erased Eq implementation for K, right?

fasterthanlime: Oh the key equality comparison might be completely unecessary here, yeah — might want to patch the HashMap implementation as well.

simonask: Cool, I patched HashMap as well, and did it for the `hash` vtable fn. It's kind of scary without tests though - I'd be happy to add some, but I'm not really sure where to put them. 😅

sample/Cargo.toml
```diff
@@ -4,4 +4,5 @@ publish = false
 edition = "2024"
 
 [dependencies]
-facet = { version = "0.1.17", path = "../facet" }
+# This crate has publish=false, so it doesn't need to specify a version.
+facet = { path = "../facet" }
```
simonask, line 8: Totally shoehorned this in to get the codegen test to pass, happy to revert it.

fasterthanlime: I ended up doing the same somewhere, no worries

fasterthanlime (approved):
Thanks! I'm merging this quickly so I have to deal with the merge conflicts later myself!

PR #186, opened 2025-04-12
Fix `unchecked_set` docs about errors

mkeeter:
`unchecked_set` and `unchecked_set_by_name` don't actually check field shapes, so let's remove that from the docstring.


mkeeter:
The CI failure is also happening on `main`, so I don't think it's related to this PR 😅

fasterthanlime:
> The CI failure is also happening on `main`, so I don't think it's related to this PR 😅

Yeahh. It'll be fixed b/c another PR is being merged.

`unchecked_set` is dying after my next PR, but let's merge this in the meantime.

PR #188, opened 2025-04-12
Rewrite facet-reflect for safety

fasterthanlime:
In particular this:

  * Only exposes a safe API for Peek/Poke
  * Provides `Slot` for safe partial structure initialization
  * Add supports for validating invariants


fasterthanlime:
Even though this comments out `facet-json`, `facet-toml`, `facet-yaml`, and `facet-msgpack` from the workspace, I'm still going to merge it because the previous interface was unsafe. Now `facet-reflect` is entirely safe and we can rebuild those crates on top of it. I'm merging it before any of the contributions drift any further. 

`facet-args` and `facet-urlencoded` are there as examples of the new `Wip` API, and so are the tests. There is no in-place mutation API anymore for the time being (ie. the non-uninit "Poke"). It is an enormous footgun and we will have to think about how to deal with invariants when you're able to modify every field independently. 

This closes #190 by letting `Wip` hold the lifetime of the shortest reference it contains. I think this will deserialize from borrowed strings.

Code coverage is going down for the time being.

facet-reflect/Cargo.toml
```diff
@@ -12,16 +12,21 @@ categories = ["development-tools", "rust-patterns"]
...
 [features]
 default = ["std"]
-std = ["alloc"]
+std = ["alloc", "indexmap"]
 alloc = []
 
 [dependencies]
+facet-ansi = { path = "../facet-ansi" }
 facet-core = { path = "../facet-core", version = "0.5.3" }
+indexmap = { version = "2.9.0", optional = true }
```
epage, line 21: This prevents `Wip` being used in non-std contexts

Is the dual nature of ordered and indexable by custom type required? Or is it a nicety, like order preserving?

Options
- Switch to `BTreeMap` in `alloc` contexts
- Pull in a custom hash implementation to still use `indexmap` (but then you can't deactivate that dep in `std` contexts)
- Implement a custom "index map" using `Vec` (e.g. clap does this with [`flat_map`](https://github.com/clap-rs/clap/blob/master/clap_builder/src/util/flat_map.rs) but the motivation was for binary size)

fasterthanlime: > Or is it a nicety, like order preserving?

It's a correctness thing, because we need to be able to drop fields in the right order, but I don't think we actually need an index map for that. 

I'm also not exactly sure what the drop order should be for partially initialized structs, to be honest??

PR #191, opened 2025-04-12
fix git hook portability for nix

niacdoial:
notes:
- nixos does not have /bin/bash, but it does have /bin/sh and /usr/bin/env
- ...I hope the "command" function is indeed portable to all sh implementations

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,14 @@
 <picture>
 <source srcset="https://github.com/facet-rs/facet/raw/main/static/logo-v2/logo-only.webp">
 <img src="https://github.com/facet-rs/facet/raw/main/static/logo-v2/logo-only.png" height="35" alt="Facet logo - a reflection library for Rust">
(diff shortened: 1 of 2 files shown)


fasterthanlime:
I'm not sure this PR contains the right changes?

fasterthanlime:
I pushed a fixed domain that changes the hooks to only use just, but not use bash.

niacdoial:
...indeed, this is not what I wanted to push. what the heck.
I guess changing files in .git doesn't get committed properly, my bad. Guess I'll have to learn how those hooks were set up in the first place.

(ironically enough, the changes pushed were generated by the on-push hook because my fork has a different name :|  )

fasterthanlime:
> Guess I'll have to learn how those hooks were set up in the first place.

They're in `.cargo-husky`, see https://crates.io/crates/cargo-husky

PR #193, opened 2025-04-13
Deserialize enum in TOML

tversteeg:
Not everything is covered yet, which I will add later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,8 +224,11 @@ name = "facet-toml"
 version = "0.3.0"
 dependencies = [
  "facet",
+ "facet-ansi",
  "facet-core",
  "facet-reflect",
+ "facet-testhelpers",
+ "log",
  "num-traits",
  "toml_edit",
 ]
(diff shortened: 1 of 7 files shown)


fasterthanlime:
I'll hold off on reviewing this until #188 lands — it closes a bunch of unsoundness and should make it easier to write format libs.

fasterthanlime:
Okay yeah I'm focused on bringing `facet-json` back from the dead for now. I suspect `facet-toml` will need to be rewritten to take advantage of the iterative approach at some point, too.

tversteeg:
> Okay yeah I'm focused on bringing `facet-json` back from the dead for now. I suspect `facet-toml` will need to be rewritten to take advantage of the iterative approach at some point, too.

Ok, I'll do that after you've implemented `facet-json`, that way I'll have a reference.

I've updated this PR with the newest changes, the push/pop `Wip` system is a big improvement over the previous system and very ergonomic!

tversteeg:
I've updated this PR with the newest changes, the push/pop `Wip` system is a big improvement over the previous system and very ergonomic!

PR #199, opened 2025-04-13
Minor updates to Readme and other docs

DavidAntliff:
I saw a few minor things that could be fixed - nothing significant though.

I assumed `nt` is a private alias to `cargo nextest`?

I think [this](https://github.com/facet-rs/facet/blob/046ca7eccaba9890a0b5f6c3eaf8f065a4ad6b20/CONTRIBUTING.md?plain=1#L22) might be incorrect:

> Similarly, don't edit `tuples_impls.rs`, edit the `tuples_impls.rs.j2` next to
it, and then run `just codegen` — the code is in `facet-codegen/`

There don't seem to be any .j2 templates on this branch. I haven't changed this in this PR, though.

Also, aside, the result of `just codegen` is dependent on the name of the directory that the developer has checked the project out into. For example, if the dev chooses `facet.git` instead of `facet` then the output of `just codegen` will be significantly different. Might be worth mentioning this somewhere, maybe `facet/CONTRIBUTING.md`? Or minor changes to `just codegen` to use a fixed path?

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,7 +10,7 @@ happily do it. I will be happier clarifying ahead of time than undoing your stuf
 
 In general, give a high-level overview of your plan and ask for consent before making any edits. Thanks.
 
-Don't edit any `README.md` files directly — edit `templates/README.md.j2` and run `just codegen` to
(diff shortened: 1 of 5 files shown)


fasterthanlime:
Done in #201 — thanks so much for flagging this.

PR #215, opened 2025-04-14
Remove unneeded cfg

mkeeter:
`impls_alloc` is already gated by `feature = "alloc"` at the module level.


fasterthanlime:
Yes! Meant to remove those but forgot.

Issue #210, opened 2025-04-14
Handling "opaque" smart pointers


mkeeter:
I've got a thin wrapper around a smart pointer, which I would like to derive `Facet` on:

```rust
/// Owned handle for a standalone math tree
#[derive(Clone, Debug)]
pub struct Tree(Arc<TreeOp>);
```

Unfortunately, I can't derive `Facet` on the inner `TreeOp`.  It's both recursive (#203) _and_ includes foreign types which don't derive `Facet` (in this case, `nalgebra` matrices).

However, I _also_ don't need reflection into the `TreeOp`, so I don't actually care about the `SmartPointerDef::pointee` field; I just want to set values of type `Tree` in other structs using reflection.  The `pointee` shape isn't load-bearing for this use case.

I'm wondering whether Facet should have a strategy for these kind of opaque pointers.

Here's a two-step approach that I'm considering:

1) Make `SmartPointerDef::pointee` an `Option<&'static Shape>`.  This would let people manually implement `Facet` without having to [lie about the `pointee`](https://github.com/mkeeter/fidget/blob/e30323f29518f179da4f42636b665ef209cbe232/fidget/src/core/context/tree.rs#L368-L369).  At this point, it will always be `Some(..)` for code using the derive macro.
2) Then, add a `#[facet(opaque)]` annotation to the derive macro, which makes the pointee `None` instead

I've got code doing (1) already, and would be willing to work on (2) if you think it's a good strategy.  Otherwise, happy to hear other thoughts!

fasterthanlime:
Does solving #203 help at all?

PR #218, opened 2025-04-14
Make `SmartPointerDef::pointee` optional

mkeeter:
This is a partial fix for #210: it allows people to manually implement `Facet` for "opaque" smart pointers, where we don't care about introspecting the shape of the pointed-to type.

(There will be a follow-up PR to add a `#[facet(opaque)]` syntax to the derive macro, but that one is messier)

I'm open to adjectives other than "opaque", which is used elsewhere in the codebase with a slightly different meaning.  Looking through a list of synonyms, the only one that I like is "veiled"; other suggestions are welcome.

### DIFF
--- a/facet-core/src/types/smartptr.rs
+++ b/facet-core/src/types/smartptr.rs
@@ -13,8 +13,8 @@ pub struct SmartPointerDef {
     /// vtable for interacting with the smart pointer
     pub vtable: &'static SmartPointerVTable,
 
-    /// shape of the inner type of the smart pointer
-    pub pointee: &'static Shape,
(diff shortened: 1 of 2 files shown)


fasterthanlime:
> I'm open to adjectives other than "opaque", which is used elsewhere in the codebase with a slightly different meaning. Looking through a list of synonyms, the only one that I like is "veiled"; other suggestions are welcome.

Tbh the `Opaque` elsewhere in the codebase should just be `Ptr`.

Have you considered "Concealed", "Covert", "Incognito"?

PR #226, opened 2025-04-15
Add `#[facet(opaque)]` annotation for `Arc<T>`

### LINKED ISSUE #210: Handling "opaque" smart pointers

mkeeter:
This closes #210 by letting `#[derive(Facet)]` structures contain an `Arc<T>`, where `T` does _not_ implement `Facet`.

It defines a new type `pub struct OpaqueArc<T>(pub alloc::sync::Arc<T>)`, which implements `Facet` for **all** `T`, and has `pointee: None` in its `SmartPointerDef`.

Then, if there's a `#[facet(opaque)]` annotation, the macro wraps the member in `OpaqueArc` in the closure that's used to extract its `SHAPE`.

This required changing the closure signature in `shape_of` to return a value rather than a reference, since we can't forge a reference to an `OpaqueArc` that isn't in the struct.  I _think_ this is okay; the closure is only used for type inference, and isn't actually called.

I added a unit test for building such a structure:
```rust
    #[derive(Copy, Clone, Debug, Eq, PartialEq)]
    pub struct NotDerivingFacet(u64);

    #[derive(Facet)]
    pub struct Handle(
        #[facet(opaque)]
        std::sync::Arc<NotDerivingFacet>
    );

    #[derive(Facet)]
    pub struct Container {
        inner: Handle,
    }
```

It's now possible to poke a `Handle` into a WIP `Container`.

Note that it's _not_ trivially possible to poke an `Arc<NotDerivingFacet>` into a WIP `Handle`; to do so, the user has to manually wrap it in `OpaqueArc`.  I don't love this, but don't a strategy for avoiding it yet.

--------------------------------------------------------------------------------

I'm open to iteration if you have feedback, e.g. where and how to document this functionality.

### DIFF
--- a/facet-core/src/impls_alloc/smartptr.rs
+++ b/facet-core/src/impls_alloc/smartptr.rs
@@ -1,8 +1,8 @@
 use core::alloc::Layout;
 
 use crate::{
-    ConstTypeId, Def, Facet, KnownSmartPointer, PtrConst, SmartPointerDef, SmartPointerFlags,
-    SmartPointerVTable, value_vtable,
+    ConstTypeId, Def, Facet, KnownSmartPointer, Opaque, PtrConst, SmartPointerDef,
+    SmartPointerFlags, SmartPointerVTable, value_vtable,
 };
 
 unsafe impl<T: Facet> Facet for alloc::sync::Arc<T> {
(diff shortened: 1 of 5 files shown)


mkeeter:
Uhhhh I misclicked and didn't mean to close this; I think I've got a fix in the new branch.

mkeeter:
I'm now seeing the correct commit / changes – everyone loves eventual consistency?

fasterthanlime (approved):
I'm guessing Arc went first b/c it's the one you needed? But we'll need others eventually right?

mkeeter:
Yup, that was my plan!

This is another advantage of the `Opaque<T>` type: we can add `impl Facet<T> for Opaque<OtherPointerType<T>>` as needed, which is cleaner than the dedicated `OpaqueArc<T>` from the previous implementation.

fasterthanlime:
> which is cleaner than the dedicated `OpaqueArc<T>` from the previous implementation.

Ah yes, that's why I meant to ask a question. Good good, carry on!

PR #227, opened 2025-04-15
fix(def): add `PeekMap` def getter

lmmx:
Please can `PeekMap` have a def getter like `PeekList` and friends? :slightly_smiling_face: 

```bash
louis 🌟 ~/lab/facet/facet/facet-reflect/src/peek $ grep -r 'fn def'
list.rs:    pub fn def(&self) -> ListDef {
option.rs:    pub fn def(self) -> OptionDef {
smartptr.rs:    pub fn def(&self) -> &SmartPointerDef {
struct_.rs:    pub fn def(&self) -> &Struct {
enum_.rs:    pub fn def(self) -> EnumDef {
```

I am writing a proof of concept deserialisation function and this appeared as a cargo build blocker, I hope it's not a problem to add.


fasterthanlime (approved):
LGTM!

PR #246, opened 2025-04-15
remove PtrMut::write

### LINKED ISSUE #241: PtrMut::write doesn't work as advertised
https://github.com/facet-rs/facet/blob/16d28673a75af688b6ecfcb2bbbb92ebf8e10b99/facet-core/src/ptr.rs#L265-L271

This only copies one byte, which is almost certainly not what's intended. See the following test for an example.
There is also no way to fix this given what `PtrMut` and `PtrConst` are intended to do, since they don't have access to any type information.

Fortunately this doesn't seem like it's being used (no compile errors after removing it!), so I would opt to just remove it and take the breaking change. 

```rust
#[test]
fn test_ptr_write() {
    let mut bytes = [0u8; 16];
    let other_bytes = [0xffu8; 16];

    unsafe {
        let ptr = PtrMut::new(&raw mut bytes);
        let src = PtrConst::new(&raw const other_bytes);

        ptr.write(src);
    }

    assert_eq!(bytes, other_bytes); // fails, only bytes[0] == 0xff, all other bytes are 0
}
```

RustyYato:
This function doesn't work as intended, and isn't used anywhere so let's get rid of it

fixes #241

### DIFF
--- a/facet-core/src/ptr.rs
+++ b/facet-core/src/ptr.rs
@@ -253,20 +253,4 @@ impl<'mem> PtrMut<'mem> {
     pub unsafe fn replace<T>(self, value: T) -> Self {
         unsafe { self.drop_in_place::<T>().put(value) }
     }
-
-    /// Copies data from the source pointer to this location
-    ///
-    /// # Safety
-    ///
-    /// - The destination pointer must be properly aligned and point to allocated memory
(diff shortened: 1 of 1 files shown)


fasterthanlime (approved):
🔥

PR #254, opened 2025-04-16
feat(json): add Ariadne dependency for errors [closes #251]

### LINKED ISSUE #251: feat(json): Use miette or ariadne for error reporting
Let's find out what this could look like

lmmx:
PR to replace the manual error markup with Ariadne:

- #251

Marks the span correctly for the StringAsNumber error type

Screenshot from 2025-04-16 14-44-42


zesterer:
Hello :) Thanks for using ariadne! Just a minor comment: I would recommend you do one of the following:

1) Avoid testing the *exact* output of ariadne in integration tests: we don't making promises about the exact output as part of our semver guarantees, so you might find that your tests start breaking due to minor formatting changes between patch releases, say.

2) Pin the exact crate version for tests to remove the chance of accidental breakage. Using the following should be sufficient:

```toml
[dev-dependencies]
ariadne = "=0.5.1"
```

Of course, if you notice any actual regressions in the quality of the diagnostic output, I'd definitely appreciate a bug report :grinning:

facet-json/src/deserialize.rs
```diff
@@ -54,40 +56,54 @@ pub enum JsonErrorKind {
...
+                    self.pos + 1
+                } else {
+                    self.input.len()
+                };
+                (self.pos, span_end)
             }
+        };
+
+        let mut report = Report::build(ReportKind::Error, (source_id, span_start..span_end))
+            .with_message(format!("Error at {}", self.path.fg(Color::Yellow)));
```
zesterer, line 78: You might want to add `.with_config(Config::new().with_index_type(IndexType::Byte))`.

Ariadne currently defaults to character indices: this is a misstep from the past that's going to be fixed soon.

lmmx: Ah wow thanks for reviewing! Opened #257 to track this :slightly_smiling_face:

PR #269, opened 2025-04-17
chore(reflect): remove unnecessary `matches!` followed by the same `let`

tversteeg:
To be used instead of this horrific incantation: https://github.com/facet-rs/facet/pull/259/files#diff-7c313d0af4bf7f525a9b8e4f98285f8c778fdbc88dc9de2c759c2efeaea29070R98-R101

### DIFF
--- a/facet-reflect/src/wip/mod.rs
+++ b/facet-reflect/src/wip/mod.rs
@@ -864,7 +864,7 @@ impl<'a> Wip<'a> {
         }
     }
 
-    /// Puts the default value in the currrent frame.
+    /// Puts the default value in the current frame.
     pub fn put_default(mut self) -> Result<Self, ReflectError> {
         let Some(frame) = self.frames.last_mut() else {
             return Err(ReflectError::OperationFailed {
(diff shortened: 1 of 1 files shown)


fasterthanlime:
> this horrific incantation

Haha, yeah. The thing is, you'll notice when doing an iterative approach — at some point you don't necessarily know if the option's contents are going to be Some or None yet and you have to allocate for Some for the next value parse, so..

Anyway, yeah put_none is a nice addition — just gotta get Miri to pass.

tversteeg:
> > this horrific incantation
> 
> Haha, yeah. The thing is, you'll notice when doing an iterative approach — at some point you don't necessarily know if the option's contents are going to be Some or None yet and you have to allocate for Some for the next value parse, so..
> 
> Anyway, yeah put_none is a nice addition — just gotta get Miri to pass.

I've quickly renamed it to `push_none` and actually push the value because that's more consistent with the rest right? Or should it still be `put_none`?

fasterthanlime:
> Or should it still be `put_none`?

Depends — do you need to `pop` after it? Adding a leak test is nice but let's also add "happy path" tests for option building.

By the way, `put_default` works here :) So.. not entirely sure `push_none` is really needed after all.

tversteeg:
> > Or should it still be `put_none`?
> 
> Depends — do you need to `pop` after it? Adding a leak test is nice but let's also add "happy path" tests for option building.

No, I don't think popping is necessary.

> By the way, `put_default` works here :) So.. not entirely sure `push_none` is really needed after all.

Oh yes, of course, I hadn't considered that, in that case I'll convert this PR to remove the unnecessary `matches!`.

PR #282, opened 2025-04-17
fix(utf8): deserialise UTF-8 encoded field names (closes #277)

### LINKED ISSUE #277: Field rename Unicode support
My hunch is there's something failing to do with field name parsing, but it's hard to say where without a precise repro

I ran through ThouCheese's tricky test case rename `@#$%^&€` (via #138) and it looks like it causes a panic when it reaches `UnknownField`'s message function call based on the backtrace...

https://github.com/facet-rs/facet/blob/3c6dc8fbdcf9a87f969f7b40aa529ca2db44a9ae/facet-json/src/deserialize.rs#L50

I decided to check each of the chars individually to try to find the problematic one(s). ([Repro here](https://github.com/lmmx/facet-validate/blob/97628130692e543640cd65c259dd18f46e14496e/src/main.rs))

The following are all OK:

- `@@@`
- `#`
- `##`
- `$$$`
- `%%%`
- `^^^`
- `&&&`

Not OK:

- `€`
- `😄`

I checked the codegen and the FieldAttribute seems fine:

```rust
                    .attributes(&[::facet::FieldAttribute::Rename("@#$%^&€")])
```

I expect it'll be in the... parsing -- **edit** I think it's here specifically:

https://github.com/facet-rs/facet/blob/6fda19d334906ceba6f4a3adb4808abc1243dab8/facet-json/src/deserialize.rs#L318-L340

`c as char` treats each byte as an entire Unicode scalar value.

For a multibyte UTF‑8 sequence like 🏀 ([0xF0, 0x9F, 0x8F, 0x80]) that produces hot garbage chars ("ð\u{9f}\u{8f}\u{80}").  The parser later can’t find a struct field called that, so it builds `JsonErrorKind::UnknownField` and the `unwrap()` panics.

lmmx:
- **fix(utf8): properly handle UTF-8 encoding in field names; add tests**

### DIFF
--- a/facet-json/src/deserialize.rs
+++ b/facet-json/src/deserialize.rs
@@ -48,6 +48,7 @@ impl<'input> JsonParseErrorWithContext<'input> {
             JsonErrorKind::NumberOutOfRange(n) => format!("Number out of range: {}", n),
             JsonErrorKind::StringAsNumber(s) => format!("Expected a string but got number: {}", s),
             JsonErrorKind::UnknownField(f) => format!("Unknown field: {}", f),
+            JsonErrorKind::InvalidUtf8(e) => format!("Invalid UTF-8 encoding: {}", e),
         }
     }
 }
@@ -65,6 +66,8 @@ pub enum JsonErrorKind {
     StringAsNumber(String),
     /// An unexpected field name was encountered in the input.
     UnknownField(String),
+    /// A string that could not be built into valid UTF-8 Unicode
+    InvalidUtf8(String),
 }
 
 #[cfg(not(feature = "rich-diagnostics"))]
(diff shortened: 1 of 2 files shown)


facet-json/src/deserialize.rs
```diff
@@ -329,15 +332,33 @@ pub fn from_slice_wip<'input, 'a>(
...
                                 }
                                 _ => {
+                                    bytes.push(c);
                                     pos += 1;
-                                    value.push(c as char);
                                 }
                             }
                         }
+
+                        // Convert collected bytes to string at once
```
fasterthanlime, line 351: ohhhhhhh

PR #283, opened 2025-04-17
Update derive macro to compute repr(C) offsets at derive-time.

samscott89:


fasterthanlime (approved):
Wonderful. That was fast!

PR #290, opened 2025-04-18
Allow manually defining enum discriminants

### LINKED ISSUE #284: Facet derive not support this kind of enum
This code panic.

```rust
#[repr(u8)]
#[derive(Facet)]
enum Test {
  A = 0,
}
```

facet version 0.9

A-Walrus:
Fixes #284.

see the reference for how discriminants are assigned
https://doc.rust-lang.org/reference/items/enumerations.html#implicit-discriminants

BTW there is a lot of duplication between processing primitive enums and c-style enums. we should probably merge the code somewhat at some point


fasterthanlime:
> BTW there is a lot of duplication between processing primitive enums and c-style enums. we should probably merge the code somewhat at some point

👍

fasterthanlime (approved):
LGTM on a cursory look

Issue #143, opened 2025-04-11
Support skip_serializing(_if)


ThouCheese:
In order to control whether or not we actually serialise a field, it'd be good to support an attribute like `skip_serializing_if`.

```rust
#[test]
fn test_skip() {
    #[derive(Debug, PartialEq, Eq, facet_derive::Facet)]
    struct Struct {
        always: Option<i32>,
        #[facet(skip_serializing_if = Option::is_none)]
        sometimes: Option<i32>,
    }

    let test1 = Struct {
        always: None,
        sometimes: None,
    };
    let test1_peek = facet_peek::Peek::new(&test1);
    let test1_str = r#"{"always": null}"#;
    assert_eq!(test1_str, facet_json::to_json_string(test1_peek, false));
    assert_eq!(test1, facet_json::from_str(test1_str).unwrap());

    let test2 = Struct {
        always: Some(1),
        sometimes: Some(2),
    };
    let test2_peek = facet_peek::Peek::new(&test2);
    let test2_str = r#"{"always": 1,"sometimes": 2}"#;
    assert_eq!(test2_str, facet_json::to_json_string(test2_peek, false));
    assert_eq!(test2, facet_json::from_str(test1_str).unwrap());
}
```

ThouCheese:
Wow this is done now, closing!

PR #296, opened 2025-04-19
Implement the skip serializing attribute

ThouCheese:
We do be wanting to deal with #143


facet-core/src/types/field.rs
```diff
@@ -54,6 +54,10 @@ pub struct FieldBuilder {
 pub enum FieldAttribute {
     /// Marks field as containing sensitive information
     Sensitive,
+    /// Skip serializing this field.
+    SkipSerializing,
+    /// Skip serializing this field if the function identified by `.0` evaluates to true.
+    SkipSerializingIf(&'static str),
```
ThouCheese, line 60: I think we need to extract the function pointer or at least a `Box<Fn(&StructName) -> bool>` at this point instead of just the function name

facet-json/tests/integration/write.rs
```diff
@@ -136,3 +136,27 @@ fn test_field_rename_serialization() {
...
+        goodbye: "world",
+    };
+    let json = facet_json::to_string(&test_struct1);
+    assert_eq!(json, r#"{"hello":"monde"}"#);
+
+    #[derive(Debug, PartialEq, Clone, Facet)]
+    struct Salutations(&'static str, #[facet(skip_serializing)] &'static str);
+    let test_struct2 = Salutations("groetjes", "wereld");
+    let json = facet_json::to_string(&test_struct2);
+    assert_eq!(json, r#"{"0":"wereld", "1":"wereld"}"#);
```
ThouCheese, line 161: Tuple structs seem to be serializing as `{"0": "objects"}`, whereas serde does an array.

ThouCheese:
Writing down some of my thoughts so I don't forget

facet-core/src/types/field.rs
fasterthanlime: No BoxFn plz — yes to function pointers, like in all the other vtables

Veykril, line 60: Wouldn't it make more sense for these to be encoded in `Arbitrary` somehow, having `facet-json` (and other serializers) parse that our of arbitrary? Otherwise this now ties serialization behavior into facet-core.

ThouCheese: I just followed the logic of `#[facet(rename = "...")]` (`Rename` down below), which does the exact same thing

ThouCheese: So if someone writes `#[facet(skip_serialing_if = Vec::is_empty)]` the type in here should be `&fn(Vec<T>) -> bool`. I don't think we already know the type `Vec<T>` here however.

Veykril: Well rename is a more general concept, not specific to serialization.

fasterthanlime: I’m open to suggestions, but I think it would be hard to do arbitrary attributes that are actually symbols? We can try and figure out some syntax I suppose

ThouCheese:
Fun times were had by all, in order to get the skip_serializing_if function I need to type-erase it to an `fn(* const ()) -> bool` style function pointer. Then I need to call that while serializing to figure out whether a field should be included or not. This unfortunately conflicts with the beautiful deny(unsafe) that facet-json just received.

facet-json/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 #![warn(missing_docs)]
 #![warn(clippy::std_instead_of_core)]
 #![warn(clippy::std_instead_of_alloc)]
-#![deny(unsafe_code)]
+// #![deny(unsafe_code)]
```
ThouCheese, line 5: This how you tell a quality PR

fasterthanlime: Nah we're just going to move that to facet-reflect :)

facet-core/src/types/field.rs
fasterthanlime: nope, `fn(OpaquePtr) -> bool` — and we generate the glue in the derive macro — same as `default`

facet-json/tests/integration/write.rs
fasterthanlime: Yeah let's file an issue for that

facet-derive-emit/src/lib.rs
```diff
@@ -172,12 +173,20 @@ pub(crate) fn gen_struct_field(
                             let value = attr_str[equal_pos + 1..].trim().trim_matches('"');
                             attribute_list
                                 .push(format!(r#"::facet::FieldAttribute::Rename({:?})"#, value));
+                        } else if key == "skip_serializing_if" {
+                            let value = attr_str[equal_pos + 1..].trim();
+                            attribute_list.push(format!(
+                                r#"::facet::FieldAttribute::SkipSerializingIf(unsafe {{ ::std::mem::transmute({value} as fn(&{field_type}) -> bool) }})"#,
```
ThouCheese, line 179: Is a `transmute` from `field_type` to `PtrConst` safe?

facet-core/src/types/field.rs
```diff
@@ -57,7 +59,7 @@ pub enum FieldAttribute {
     /// Skip serializing this field.
     SkipSerializing,
     /// Skip serializing this field if the function identified by `.0` evaluates to true.
-    SkipSerializingIf(fn(*const ()) -> bool),
+    SkipSerializingIf(fn(PtrConst) -> bool),
```
fasterthanlime, line 62: the bad news is: this 100% needs to be an `unsafe fn` because we're passing it a pointer!

but the unsafe code for calling it should live in `facet-core` or `facet-reflect` or something — not in `facet-json`.

Just like we do for `Wip::set_default`  for example—in core and reflect we have unsafe code, we have unsafe function pointers, and we have safe wrappers on top of that, which we can then use from deserializers.

fasterthanlime (changes requested):

facet-core/src/types/field.rs
ThouCheese: Alright sorted! I've also typedefed this function like the rest of them are, this is now a `SkipSerializingIfFn`. I moved the logic of "which field are and aren't included in serializing" into facet-reflect, which removes a bit of duplication and moves the unsafety into that crate.

fasterthanlime:
@ThouCheese you can run `just msrv` locally to iterate on fixing the msrv check, just fyi!

fasterthanlime:
I think the write tests belong in separate `write/XXX.rs` files now @ThouCheese if you want to do that last fix

fasterthanlime:
nvm I'm gonna resolve the conflicts meself

fasterthanlime:
CleanShot 2025-04-20 at 22 43 44@2x

nooooo :( we do rebase here

ThouCheese:
Oops, we don't love merge commits?

fasterthanlime:
> Oops, we don't love merge commits?

The royal we doesn't. I'm looking at how to recover but if we can't, then we can't

ThouCheese:
Squashing maybe?

fasterthanlime:
> Squashing maybe?

Yeah, `git merge --squash` from a clean branch did the trick.

fasterthanlime:
Cleaned up in #342

PR #305, opened 2025-04-19
feat(derive): Support facet(transparent) attr on containers

fasterthanlime:


facet-derive-emit/src/lib.rs
```diff
@@ -152,6 +152,9 @@ pub(crate) fn gen_struct_field(
                             .trim_end_matches('"') // FIXME: that's pretty bad
                     ));
                 }
+                FacetInner::Transparent(_) => {
+                    // Not applicable on fields; ignore.
```
lffg, line 156: Why not panic here?

fasterthanlime: Yeah let's file that as an issue — it should either be supported or yield a compile error. Can you open an issue for it?

lffg: Yes, of course! If you prefer, I could open a PR to put a panic in there so that we get a compile error. Or would you prefer another approach?

PR #307, opened 2025-04-19
Impl `Facet` for `Rc<T>`

axelkar:
Also added tests in facet-reflect

### DIFF
--- a/facet-core/CHANGELOG.md
+++ b/facet-core/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Other
+
+- Impl `Facet` for `Rc<T>` ([#307](https://github.com/facet-rs/facet/pull/307))
+
 ## [0.10.0](https://github.com/facet-rs/facet/compare/facet-core-v0.9.1...facet-core-v0.10.0) - 2025-04-19
 
 ### Added
--- a/facet-core/CHANGELOG.md
+++ b/facet-core/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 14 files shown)


axelkar:
How should I deallocate the `PtrConst`, as instructed by `NewIntoFn`'s documentation? `std::mem::forget`?

Should I add `into_smart_pointer` etc. functions to `facet::Def`?

FYI I'm trying to get the coverage up by adding tests for the vtable

fasterthanlime:
I think you can look at Arc<T> for guidance on both of these! Yes to into_smart_pointer

axelkar:
I found this comment on a `strong` field: `shape of the corresponding non-atomic pointer, if this pointer is atomic.`. I think it's incorrect so I removed it.

Naming is inconsistent in `SmartPointerVTableBuilder`, not sure what to do there and I don't think this PR will have a major version bump.

fasterthanlime:
> Naming is inconsistent in `SmartPointerVTableBuilder`,

All breaking changes are on the table now so please make things consistent :) I don't worry too much, the cargo-semver-checks will take care of that for us.

facet-core/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `into_*` functions to `Def`
```
fasterthanlime, line 12: this is automatically updated by release-plz btw

facet-core/src/types/mod.rs
```diff
@@ -262,7 +262,7 @@ impl Shape {
 #[non_exhaustive]
 // this enum is only ever going to be owned in static space,
 // right?
-#[allow(clippy::large_enum_variant)]
+#[expect(clippy::large_enum_variant)]
```
fasterthanlime, line 265: 👍

fasterthanlime (approved):
LGTM, thanks for the contribution!

fasterthanlime:
Rebased in #331

PR #308, opened 2025-04-19
chore: add Nix flake

axelkar:
Introduced a development environment managed by [Nix](https://nixos.org). More info [here](https://nix.dev/#what-can-you-do-with-nix). It can be hooked up to direnv with [nix-direnv](https://github.com/nix-community/nix-direnv/) and the `use flake` command.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,42 @@
+{
+  "nodes": {
+    "flake-compat": {
+      "locked": {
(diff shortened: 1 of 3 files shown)


fasterthanlime:
I have personally moved away from nix, but I don’t mind this being here so I’m merging it anyway

PR #309, opened 2025-04-19
Don't allocate strings in facet-json deserialization unless necessary

Veykril:


fasterthanlime:
Not directly related, but I was thinking of using alternative allocators because we have to do a bunch of allocations in WIP

PR #311, opened 2025-04-19
feat(toml): pretty error reports & enum fix

tversteeg:
Sorry for being inactive for a few days! I'm also currently working on adding proper error handling with ariadne similar to the JSON crate.

### DIFF
--- a/facet-toml/src/error.rs
+++ b/facet-toml/src/error.rs
@@ -4,7 +4,7 @@ use core::ops::Range;
 
 #[cfg(feature = "rich-diagnostics")]
 use ariadne::{Color, Label, Report, ReportKind, Source};
-use facet_core::{Shape, TypeNameOpts};
+use facet_core::Shape;
 use facet_reflect::ReflectError;
 
 /// Any error from deserializing TOML.
@@ -35,12 +35,10 @@ impl<'input> TomlError<'input> {
                 rust_type,
                 reason,
             } => {
-                let rust_type_name = TypeNameWriter(rust_type);
-
                 if let Some(reason) = reason {
-                    format!("Can't parse type '{rust_type_name}' from '{toml_type_name}': {reason}")
+                    format!("Can't parse type '{rust_type}' from '{toml_type_name}': {reason}")
                 } else {
-                    format!("Can't parse type '{rust_type_name}' from '{toml_type_name}'")
+                    format!("Can't parse type '{rust_type}' from '{toml_type_name}'")
                 }
             }
             TomlErrorKind::ExpectedType { expected, got } => {
@@ -48,13 +46,10 @@ impl<'input> TomlError<'input> {
             }
             TomlErrorKind::UnrecognizedType(r#type) => format!("Unrecognized type '{type}'"),
             TomlErrorKind::UnrecognizedScalar(scalar_type) => {
-                format!(
-                    "Unrecognized Rust scalar type '{}'",
-                    TypeNameWriter(scalar_type)
-                )
+                format!("Unrecognized Rust scalar type '{scalar_type}'",)
             }
             TomlErrorKind::InvalidKey(field) => {
-                format!("Invalid Rust key '{}'", TypeNameWriter(field))
+                format!("Invalid Rust key '{field}'")
             }
             TomlErrorKind::ExpectedFieldWithName(name) => {
                 format!("Expected field with name '{name}'")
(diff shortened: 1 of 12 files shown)


tversteeg:
Added similar error handling to facet-json:

image

facet-toml/src/error.rs
```diff
@@ -1,25 +1,175 @@
...
+    /// Tried parsing a single value as a struct with multiple fields.
+    ParseSingleValueAsMultipleFieldStruct,
+}
+
+/// Wrap the `Shape` in a type so we can write it's type name directly with display.
+struct TypeNameWriter(&'static Shape);
+
+impl core::fmt::Display for TypeNameWriter {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.write_type_name(f, TypeNameOpts::default())
```
tversteeg, line 173: @fasterthanlime I feel this was a bit cumbersome to find out, do you mind if I add a `type_name_to_string()` or similar named to `Shape`?

facet-reflect/src/error.rs
```diff
@@ -1,7 +1,7 @@
 use facet_core::{Characteristic, EnumDef, Field, FieldError, Shape};
 
 /// Errors that can occur when reflecting on types.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
tversteeg, line 4: I also sneakily added this in here, it's very useful to be able to compare what kind of error you have.

facet-toml/src/error.rs
fasterthanlime: > do you mind if I add a `type_name_to_string()` or similar named to `Shape`?

I do mind, `to_string()` allocates x)

But you can just do `{shape}` in error messages, since it implements Display.

facet-reflect/src/error.rs
fasterthanlime: Sure, that's okay!

facet-toml/src/error.rs
tversteeg: If I do `{shape}` I get the whole shape description, which is quite noisy for a single line. I only want to know the type name.

fasterthanlime:
I'm reworking JSON (again) in #317 — there's going to be conflicts here unfortunately haha.

I have a new approach based on `try_from` for numbers that I think you might like — I feel like the TOML & JSON crates are in healthy competition to become as good as one another :)

facet-toml/src/error.rs
tversteeg: You're always one step ahead of me, I was only looking at the `{shape:?}` implementation, thanks!

tversteeg:
> I'm reworking JSON (again) in #317 — there's going to be conflicts here unfortunately haha.

No problem at all, always happy to see new improvements that make the implementation easier to write and understand.

> I have a new approach based on `try_from` for numbers that I think you might like — I feel like the TOML & JSON crates are in healthy competition to become as good as one another :)

Ah, so basically what I did [here](https://github.com/facet-rs/facet/pull/311/files#diff-34dca0920c9372ee7259ae768d100cafd720adc1e8c07610ba0f2af103c4d858R33-R54) without `NumCast`? If that's the case, I can completely drop the `num-traits` dependency, can't wait!

fasterthanlime:
> Ah, so basically what I did [here](https://github.com/facet-rs/facet/pull/311/files#diff-34dca0920c9372ee7259ae768d100cafd720adc1e8c07610ba0f2af103c4d858R33-R54) without `NumCast`? If that's the case, I can completely drop the `num-traits` dependency, can't wait

Basically yes — the matrix is incomplete, unfortunately but as a rule, if something can be useful for more than 1 parser, it's better if it lives in -reflect or -core rather than -toml/json etc. It'll make it easier to cover more formats!

fasterthanlime:
> I'm also currently working on adding proper error handling with ariadne similar to the JSON crate.

Re this: see #251 — I think long-term, miette is better, it's designed for diagnostics with multiple spans etc.

I'm going to experiment in JSON first, but if you don't want churn _there too_ then don't do too much work with ariadne I guess x)

fasterthanlime:
It would be nice to have future PRs be smaller! But I understand it can be hard due to timezone differences etc.

tversteeg:
> It would be nice to have future PRs be smaller! But I understand it can be hard due to timezone differences etc.

I agree, if you prefer I can also create sub-PRs that rebase into a main-PR with all the unmerged changes for TOML.

fasterthanlime:
> I agree, if you prefer I can also create sub-PRs that rebase into a main-PR with all the unmerged changes for TOML.

No, let’s not spend time splitting this one, I did big PRs for JSON too, sometimes you have to!

fasterthanlime (approved):
Only skimmed, but LGTM!

PR #314, opened 2025-04-19
Make tuples serialize into lists rather than objects

### LINKED ISSUE #312: Tuples being serialised as objects
When we serialize tuples we will probably want their representation to be an array rather than an object.

```rust
#[test]
fn test_serialize_tuple() {
    let test_tuple1 = ("groetjes", 3);
    let json = facet_json::to_string(&test_tuple1);
    assert_eq!(json, r#"["groetjes", 3]"#);
    
    #[derive(facet::Facet)]
    struct TestTuple(i32, String, bool);
    let test_tuple2 = TestTuple(3, "aaa".to_string(), true);
    let json = facet_json::to_string(&test_tuple2);
    assert_eq!(json, r#"[3, "aaa", true]"#);
}
```

ThouCheese:
To close #312. Only does serialising for now

### DIFF
--- a/facet-json/src/serialize.rs
+++ b/facet-json/src/serialize.rs
@@ -31,8 +31,17 @@ pub fn peek_to_writer<W: Write>(peek: &Peek<'_>, writer: &mut W) -> io::Result<(
 
 /// The core serialization function
 fn serialize<W: Write>(peek: &Peek<'_>, writer: &mut W) -> io::Result<()> {
+    use facet_core::{
+        Struct,
+        StructKind::{Tuple, TupleStruct},
+    };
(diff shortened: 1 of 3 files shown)


fasterthanlime:
I think it’s a good idea to merge only the serialization part for now 😌 mark it as ready to review if you agree!

also note that I am rewriting the JSON implementation once again, but as long as you add tests, then we can be sure there is no regression

ThouCheese:
Yep yep good idea! I will add some deserialization tests as well that I will mark with skip

fasterthanlime:
Rebased in #332

PR #331, opened 2025-04-20
Impl `Facet` for `Rc<T>`

fasterthanlime:
Rebased version of #307


axelkar:
Did you remove the changelog entries? The commit messages don't really line up. Thanks for the rebase!

fasterthanlime:
> Did you remove the changelog entries? The commit messages don't really line up. Thanks for the rebase!

I did — there's going to be conflicts because release-plz generates it automatically so good commit messages is all we got. I can edit them back in before release — let me do that right now.

fasterthanlime:
> let me do that right now.

Well, I can't because I don't want to regress `facet-json` but here's what it will look like:

https://github.com/facet-rs/facet/commit/12c07cc8aecfa27d08dbbae71dd98a90cbed2582#diff-b51ab0186bc9b11a6e67cd1bc34e0899cb6744e6fb393191d0eab0fab11cd6a1

Is that good enough or do you want to make some edits before it goes out?

I know the changelog is pretty rough for now, I think I'll get higher standards when things settle down a little bit, there's a lot of research still going on at this stage,

axelkar:
I just thought it'd be good to add
```
Add `into_*` functions to `Def`
```
to the changelog, but it's not a major change :+1:

PR #345, opened 2025-04-21
Actually run compile test for skip_serializing_if

ThouCheese:
This code was elegantly nicked from `facet-reflect`


ThouCheese:
I noticed a format inside a format, so I went over the code base and replaced them all where possible. Fairly few of them were in actual hot paths, most of them tests, so no big gains are to be had

facet-derive-emit/tests/compile_tests.rs
```diff
@@ -0,0 +1,176 @@
+use std::fs;
```
fasterthanlime, line 1: I'd like to extract this to `facet_testhelpers` at some point, instead of having it be duplicated.

PR #350, opened 2025-04-21
Implement `Facet` for (subset of) function pointers

Veykril:
I've set the arity to a low 6 and only added `Rust` and `C` abis to not blow things up too much.


fasterthanlime:
Well, it's a good thing this'll only be built once for large projects x) just like tuples.

Veykril:
Ye ... this can result in a massive build matrix so I imagine that this might want some feature gating to have certain ABIs / bigger arities opt-in 😅

fasterthanlime:
CleanShot 2025-04-21 at 10 37 20@2x

Woopsie

Veykril:
Interesting, I suppose miri discards closure identity or something

fasterthanlime:
> Interesting, I suppose miri discards closure/function identity or something?

If you're asking the question I think you're asking, then yes: comparing function pointers/closure addresses if moot in miri, it doesn't do what you want ever. Which I guess is a warning that, in release builds, you really shouldn't be doing that anyway because functions might get duplicated/deduplicated, whatever.

PR #352, opened 2025-04-21
feat(toml): include reflect path in error

tversteeg:


fasterthanlime:
idk if you saw I made a nice macro for it in facet-json — this isn't super final, I'm sure we'll find a better way eventually — returning the `Wip` with the error is I'm not sure the best way.

specifically, if reflection fails, we lose the Wip (because it must be deallocated) — it's not returned as part of Err variants in `Wip` methods. If that makes sense.

tversteeg:
> idk if you saw I made a nice macro for it in facet-json — this isn't super final, I'm sure we'll find a better way eventually — returning the `Wip` with the error is I'm not sure the best way.

I copied your macro here! It is indeed nice, reducing a lot of code duplication.

> specifically, if reflection fails, we lose the Wip (because it must be deallocated) — it's not returned as part of Err variants in `Wip` methods. If that makes sense.

That makes sense. But it would still be possible to extract the path before the deallocation right?

fasterthanlime:
> But it would still be possible to extract the path before the deallocation right?

Yup! Although I'm not in love with the fact that the path is just a string — if we want to keep it that way, let's make sure it's standardized across the whole ecosystem (and then it can also be used as a "selector" when doing partial deserialization, etc.)

tversteeg:
> Yup! Although I'm not in love with the fact that the path is just a string — if we want to keep it that way, let's make sure it's standardized across the whole ecosystem (and then it can also be used as a "selector" when doing partial deserialization, etc.)

Maybe make it a custom type, something like [key in toml_edit](https://docs.rs/toml_edit/latest/toml_edit/struct.Key.html)? That way, it could just implement `Display` for default nice path as-is, but still allow deserializers to choose their own representation.

facet-toml/src/to_scalar.rs
```diff
@@ -142,9 +155,12 @@ pub(crate) fn put_from_str<'input, 'a>(
                 got: item.type_name(),
             },
             item.span(),
+            wip.path(),
         )
     })?;
 
+    // TODO: only generate if actually error
+    let path = wip.path();
```
epage, line 163: Is `wip.path()` meaningful to the user?

At least with serde, the Rust data model can deviate from the Serde data due to features like `untagged`, `flatten`.  What `toml_edit` does in its serde code is to add the table keys to the error as its unwinding the stack on errors.

PR #353, opened 2025-04-21
chore(deps): update dependencies

tversteeg:
Not sure how frequent you want these types of commits.


fasterthanlime:
> Not sure how frequent you want these types of commits.

Once per week is probably good enough — we could set up dependabot :)

fasterthanlime:
> I think Renovate is better in that case, you can configure it to bundle dependency updates, which dependabot can't do AFAIK.

Works for me — if someone reading this wants to set up renovate, file an issue and assign yourself to it!

PR #358, opened 2025-04-22
Stub out facet-kdl

fasterthanlime:
As discussed with @thelostlambda in https://github.com/kdl-org/kdl-rs/pull/83#issuecomment-2820480711

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,6 +237,20 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


TheLostLambda:
Thanks @fasterthanlime ! I should be able to throw together something in the next couple of days to give people a first look at! Even just doing something like differentiating between named and positional arguments (properties and arguments in KDL) will be farther than we've ever gotten with `serde`!

Issue #190, opened 2025-04-12
Soundness issue: Can create dangling reference using safe code


tyilo:
```rust
use facet::Facet;
use facet_reflect::PokeUninit;

#[derive(Debug, Facet)]
struct Foo<'a> {
    s: &'a str,
}

fn main() {
    let (poke, _guard) = PokeUninit::alloc::<Foo>();
    let poke = poke.into_value().typed::<Foo>().unwrap();
    let v = {
        let s = "abc".to_string();
        let foo = Foo { s: &s };
        poke.put(foo)
    };
    dbg!(v.as_peek());
}
```

fasterthanlime:
facet-reflect is undergoing a complete rewrite, Ir love some eyes on #188 rather than the current API if possible!

tyilo:
I think the important thing is just to add a `T: 'mem` bound everywhere there is a generic `T` in the poke types.

fasterthanlime:
> I think the important thing is just to add a `T: 'mem` bound everywhere there is a generic `T` in the poke types.

It was a bit more complicated than that, but it's now tested in `facet-reflect/tests/wip/compile_tests.rs` — solved in #188

tyilo:
This still seems to be a problem:

```rust
use facet::Facet;
use facet_reflect::{ReflectError, Wip};

#[derive(Debug, Facet)]
struct Foo<'a> {
    v: &'a str,
}

fn main() -> Result<(), ReflectError> {
    let foo = {
        let s = String::from("abc");

        Wip::alloc::<Foo>()
            .field_named("v")?
            .put(s.as_str())?
            .pop()?
            .build()?
            .materialize::<Foo>()?
    };
    dbg!(foo);

    Ok(())
}
```

fasterthanlime:
> This still seems to be a problem:

Any ideas?

tyilo:
It seems like `materialize` has the opposite lifetime bound of what it needs. It currently has `T: 'a`, but I think what is needed is `'a: T`, which I don't know how to express in Rust.

tyilo:
I think it can be solved by adding an associated type `SelfWithLifetime` to `Facet`:

```rust
trait Facet: Sized {
    type SelfWithLifetime<'a>: Facet + 'a;
    ...
}
```

and `materialize` can then return `T::SelfWithLifetime<'a>` instead of `T`.

Example implementations of the associated type:
```rust
unsafe impl Facet for String {
    type SelfWithLifetime<'a> = String;
[...]

PR #371, opened 2025-04-22
Add (unsoundness proving) lifetime variance tests for facet-reflect

Veykril:
cc https://github.com/facet-rs/facet/issues/190


Veykril:
Closing this as https://github.com/facet-rs/facet/pull/372 has cherry picked it. CI refused to run here I think the CI setup does not like you double pushing

PR #372, opened 2025-04-22
Draft: Invariance

tyilo:
WIP #190

### DIFF
--- a/facet-reflect/src/wip/mod.rs
+++ b/facet-reflect/src/wip/mod.rs
@@ -313,7 +313,7 @@ impl<'a> Wip<'a> {
     }
 
     /// Allocates a new value of type `S`
-    pub fn alloc<S: Facet>() -> Self {
+    pub fn alloc<S: Facet<'a>>() -> Self {
         Self::alloc_shape(S::SHAPE)
     }
 
@@ -745,7 +745,7 @@ impl<'a> Wip<'a> {
     ///
     /// * `Ok(Self)` if the value was successfully put into the frame.
     /// * `Err(ReflectError)` if there was an error putting the value into the frame.
-    pub fn put<T: Facet + 'a>(self, t: T) -> Result<Wip<'a>, ReflectError> {
+    pub fn put<T: Facet<'a>>(self, t: T) -> Result<Wip<'a>, ReflectError> {
         let shape = T::SHAPE;
         let ptr_const = PtrConst::new(&t as *const T as *const u8);
         let res = self.put_shape(ptr_const, shape);
@@ -754,7 +754,7 @@ impl<'a> Wip<'a> {
     }
 
     /// Checks if the current frame is of type `T`.
-    pub fn current_is_type<T: Facet + 'static>(&self) -> bool {
+    pub fn current_is_type<T: Facet<'static>>(&self) -> bool {
         self.frames
             .last()
             .is_some_and(|frame| frame.shape == T::SHAPE)
--- a/facet-reflect/src/wip/mod.rs
+++ b/facet-reflect/src/wip/mod.rs
@@ -283,15 +283,12 @@ pub struct Wip<'a> {
     istates: FlatMap<ValueId, IState>,
 
     /// lifetime of the shortest reference we hold
-    phantom: PhantomData<&'a ()>,
+    phantom: PhantomData<*mut &'a ()>,
 }
 
 impl<'a> Wip<'a> {
     /// Puts the value from a Peek into the current frame.
-    pub fn put_peek<'mem>(self, peek: crate::Peek<'mem>) -> Result<Wip<'mem>, ReflectError>
-    where
-        'a: 'mem,
-    {
+    pub fn put_peek(self, peek: crate::Peek<'a>) -> Result<Wip<'a>, ReflectError> {
         self.put_shape(peek.data, peek.shape)
     }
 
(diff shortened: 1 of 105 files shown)


tyilo:
@fasterthanlime Maybe you can take a look at fixing the enum derive implementation? I don't think I really know what's going on with the code...

fasterthanlime:
> @fasterthanlime Maybe you can take a look at fixing the enum derive implementation? I don't think I really know what's going on with the code...

Happily! Tomorrow 😴

facet-reflect/src/wip/mod.rs
```diff
@@ -283,15 +283,12 @@ pub struct Wip<'a> {
     istates: FlatMap<ValueId, IState>,
 
     /// lifetime of the shortest reference we hold
-    phantom: PhantomData<&'a ()>,
+    phantom: PhantomData<*mut &'a ()>,
```
Veykril, line 286: nit but I would define this as follows (and do this elsewhere as well)
```suggestion
    phantom: PhantomData<fn(&'a ()) -> &'a ()>,
```
Expressing variance with function pointers it a bit more generic, `fn(a)` = contravariant, `fn() -> a` covariant, `fn(a) -> a` invariant

facet/tests/derive.rs
```diff
@@ -288,7 +288,7 @@ fn tuple_struct_field_doc_comment() {
 #[test]
 fn record_struct_generic() {
     #[derive(Clone, Hash, PartialEq, Eq, ::facet::Facet)]
-    struct Blah<'a, T: Facet, const C: usize = 3>
+    struct Blah<'a, T, const C: usize = 3>
```
Veykril, line 291: I assume this errored but we should ideally support `Facet` bounds here

tyilo: Yeah, the problem is that
```rust
#[derive(::facet::Facet)]
struct Foo<'a, T: Facet<'a>> {
    field: core::marker::PhantomData<&'a T>,
}
```

expands to

```rust
unsafe impl<'facet, 'a, T: Facet<'a>> ::facet::Facet<'facet> for Foo<'a, T>
where
    'a: 'facet,
    'facet: 'a,
    T: ::facet::Facet<'facet>,
{
    ...
}
```

which causes the following errors:

```
error[E0283]: type annotations needed: cannot satisfy `T: Facet<'a>`
 --> facet/tests/facet_bound.rs:3:10
  |
3 | #[derive(::facet::Facet)]
  |          ^^^^^^^^^^^^^^
  |
note: multiple `impl`s or `where` clauses satisfying `T: Facet<'a>` found
 --> facet/tests/facet_bound.rs:3:10
  |
3 | #[derive(::facet::Facet)]
  |          ^^^^^^^^^^^^^^
  = note: this error originates in the derive macro `::facet::Facet` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0283]: type annotations needed: cannot satisfy `T: Facet<'a>`
 --> facet/tests/facet_bound.rs:3:10
  |
3 | #[derive(::facet::Facet)]
  |          ^^^^^^^^^^^^^^
  |
note: multiple `impl`s or `where` clauses satisfying `T: Facet<'a>` found
 --> facet/tests/facet_bound.rs:3:10
  |
3 | #[derive(::facet::Facet)]
  |          ^^^^^^^^^^^^^^
note: required by a bound in `Foo`
 --> facet/tests/facet_bound.rs:4:19
  |
4 | struct Foo<'a, T: Facet<'a>> {
  |                   ^^^^^^^^^ required by this bound in `Foo`
  = note: this error originates in the derive macro `::facet::Facet` (in Nightly builds, run with -Z macro-backtrace for more info)
```

tyilo: Seems like this has already been discussed in https://github.com/serde-rs/serde/issues/2418

So maybe we don't want to support this?

Veykril: Ye it seems unlikely we can make that work

fasterthanlime: > So maybe we don't want to support this?

Yeah let's drop it for now — we can surface it to the compiler team if they're interested

fasterthanlime:
@tyilo Let me know if you need me to do anything about this PR — I know you asked about the enum generation code. Is that still something that needs to be done?

tyilo:
> @tyilo Let me know if you need me to do anything about this PR — I know you asked about the enum generation code. Is that still something that needs to be done?

Yeah, the derive for enums has been broken for some cases which I think would be nice to fix if possible. Could you take a look at that?

<hr>

We probably also want to rename `'facet` to `'__facet` to make it less likely to clash with existing lifetimes.

fasterthanlime:
> Yeah, the derive for enums has been broken for some cases which I think would be nice to fix if possible. Could you take a look at that?

I'm just adding random test cases and seeing what breaks. Is that one of the things you were talking about? Do you have failing tests anywhere I can look at?

Rustc says you cannot use generic parameters from outer items, use of generic parameters from outer item.

fasterthanlime:
Yeah, okay, mostly it's enums with a C representations that are broken.

tyilo:
> > Yeah, the derive for enums has been broken for some cases which I think would be nice to fix if possible. Could you take a look at that?
> 
> I'm just adding random test cases and seeing what breaks. Is that one of the things you were talking about? Do you have failing tests anywhere I can look at?
> 
> Rustc says you cannot use generic parameters from outer items, use of generic parameters from outer item.

Mostly thinking about the tests are commented out.

fasterthanlime:
Closing in favor of #373 — which has commits from both you and I @tyilo — if there are fixes to make we'll make them in a follow-up PR.

PR #373, opened 2025-04-23
Invariance (now with enums!)

fasterthanlime:
This builds on #372 to also bring in enum support.

I'm inclined to merge this as soon as possible. I'm just going to ask a review from @Veykril and @tyilo for the last few commits — We can always do cleanups in follow-ups.


Veykril (approved):
lgtm!

Reviewed only your commits here, but I think I reviewd tylio's parts already so

PR #375, opened 2025-04-23
Test enum `#[repr(C, int)]`

tyilo:


fasterthanlime (approved):

fasterthanlime:
At some point we might want to generate those tests, there's so many combinations!

PR #376, opened 2025-04-24
Implement Facet for references and pointers

### LINKED ISSUE #355: Implement `Facet` for `&T` and `&mut T` if `T: Facet`
Seems reasonable and would allow you to derive `Facet` for a struct having `&T` and `&mut T` fields.

tyilo:
Fixes #355

### DIFF
--- a/facet-core/src/impls_core/refptr.rs
+++ b/facet-core/src/impls_core/refptr.rs
@@ -5,10 +5,8 @@ use crate::{
     ValueVTable,
 };
 
-// TODO: Also implement for `T: !Sized`
-// - Conflicts with `&[T]`, `&str` and `&Path`
 // TODO: Also implement for `T: !Facet<'a>`
-unsafe impl<'a, T: Facet<'a>> Facet<'a> for &'a T {
+unsafe impl<'a, T: Facet<'a> + ?Sized> Facet<'a> for &'a T {
     const SHAPE: &'static Shape = &const {
         Shape::builder()
             .id(ConstTypeId::of::<Self>())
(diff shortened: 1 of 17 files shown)


facet-core/src/impls_core/refptr.rs
```diff
@@ -0,0 +1,201 @@
...
+                        builder = builder.debug(|value, f| {
+                            let v = unsafe { value.get::<Self>() } as *const &mut T;
+                            let v = unsafe { v.read() };
+                            unsafe {
+                                (T::SHAPE.vtable.debug.unwrap_unchecked())(PtrConst::new(v), f)
+                            }
+                        });
+                    }
+
+                    // TODO: More functions
```
fasterthanlime, line 96: I know half the world is saying that we're giving up our humanity by using them, but LLMs are actually very good at generating those because it's basically pattern matching — Sonnet 3.5/3.7 are great at it, GPT-4.1 is doing okay too, the Gemini models are buggy in Zed most of the time so I can't really use them :/

tyilo:
@fasterthanlime What are the guarantees for the implementations of a type's vtable?

Is it on a best-effort basis or do we need to guarantee that `T::SHAPE.vtable.debug` is `Some` if and only if `T: Debug`?


It is possible to have trait implementations only for `&T` but not for `T`, which we can't detect in the generic `&T` implementation:
```rust
#[derive(Facet)]
struct S;

impl Debug for &S {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "&S")
    }
}
```

fasterthanlime:
> It is possible to have trait implementations only for `&T` but not for `T`, which we can't detect in the generic `&T` implementation:

Oof yeah I see what you mean.

I'm not sure how to deal with that because our whole specialization scheme is based on references.

I guess personally I would not care if there are missing implementations — as in, if `&T` implements `Debug` (but `T` doesn't), and `<&T>::SHAPE` doesn't have `Debug`. Does that makes sense?

We need to pick something that is internally consistent and kind of forget about the things we cannot possibly support on Rust stable for now.

fasterthanlime:
Is there anything preventing merging this, aside from the fact that it currently has merge conflicts?

tyilo:
> Is there anything preventing merging this, aside from the fact that it currently has merge conflicts?

I think we can do that if we are happy with the potential missing implementations/marker traits.

We could consider making these have 3-states: implemented, not implemented, unknown, but we can also postpone that decision to later.

tyilo:
We probably want to fix these before merging:
- Impls for pointers
- Split `RefPtr` into two different `Def`'s: For formats you would normally only support references anyway.

fasterthanlime:
Done in #462 / #527

PR #379, opened 2025-04-24
Use declarative macros for implementing the `Facet` trait for function pointers & tuples

42triangles:
Progress:
- [x] Function pointers
  - [X] Original implementation
  - [x] Refactor macro to be linear instead of quadratic
  - [x] Check build timings
  - [x] Check the vtable macro invocation reg. the `FIXME` that was in the generating code before
- [ ] ~~Tuples~~ (will follow in a separate PR)


42triangles:
If compile times for the macro are a concern it could be refactored such that it pulls the next parameter per invocation, implements the trait, and then repeats until there are no parameters left to pull, instead of what it currently does (which is shrinking the parameter list which requires `O(n^2)` macro invocations since it copies the whole argument list from the right to the left except for the last one each iteration - but with a max parameter count of 6 this isn't a lot)

fasterthanlime:
Impressive! Can we keep the tuples one for a separate PR?

fasterthanlime:
> If compile times for the macro are a concern

Did you actually benchmark that? I feel like we could do cargo build timings for that.

Parsing the soup of tokens we used to generate isn't free either 🙃

Veykril:
Wasn't the entire point for codegenerating these to avoid using macros in favor of less compile times?

fasterthanlime:
> Wasn't the entire point for codegenerating these to avoid using macros in favor of less compile times?

No, for tuples it was too hard to do with declarative macros, at least that was my opinion.

I think you just followed suit for function pointers.

I don't see why declarative macros would be more slower or more expensive for this — but that's something we can easily measure, right?

42triangles:
> No, for tuples it was too hard to do with declarative macros, at least that was my opinion.

Taking a closer look at it, it should be pretty much the same and... fine - not *the nicest*, but doable (essentially `(T0.0, T1.1, T2.2,)` parsed as `($($element:ident.$access:tt,)+)`?)

> At least there's no lexing involved x)

Fair! Also less file I/O I guess

42triangles:
Build timings are fine, it's all within error bars - a maximum difference of 0.1s.
On my (very noisy, because it's the thing I'm using right now with browser open & everything, but you could potentially see differences better because it's much slower) laptop: 5.9s for the declarative macro, 6.0s for the generated file
On my PC (which isn't running much and has enough cores to spare): 1.4s for both

42triangles:
Expanding `value_vtable!` actually *DOES* make a difference (measured with the same commands):

| device | calling `value_vtable!` | manually expanded (and shortened) |
|-|-|-|
| laptop (slow, noisy) | 5.7s | 5.0s |
| PC | 1.4s | 1.3s |

Considering this is only *twelve* `impl`s (six per ABI) where `value_vtable!` was called, I believe that manually expanding `value_vtable!` in `facet/facet-core/src/impls_core/scalar.rs` could actually be worth it too!

42triangles:
Extra note: In my expansion of `value_vtable!` I left `drop_in_place` out since `fn` pointers don't have any drop code anyways

fasterthanlime:
Alright it sounds like expanding value_vtable is worth it!

Also, I'm now thinking that... we don't necessarily need to go through the builder pattern for implementations that are internal to facet-core, right? Only for the derive macro.

fasterthanlime (approved):

fasterthanlime:
LGTM, I requested a review from @Veykril — if that's 👍  too then let's merge this.

Veykril (approved):

PR #383, opened 2025-04-24
Fix `Peek` lifetimes

Veykril:
This is going to be annoying to fix

The lifetimes as written are incorrect, easpecially `Wip::peek` returns `Peek` that is tied to our invariant lifetime, while it needs to be tied to the `Wip` as we could drop it before the `Peek` causing UB

### DIFF
--- a/facet-reflect/src/wip/heap_value.rs
+++ b/facet-reflect/src/wip/heap_value.rs
@@ -7,10 +7,10 @@ use facet_core::{Facet, PtrConst, PtrMut, Shape};
 use owo_colors::OwoColorize as _;
 
 /// A type-erased value stored on the heap
-pub struct HeapValue<'a> {
+pub struct HeapValue<'facet_lifetime> {
     pub(crate) guard: Option<Guard>,
     pub(crate) shape: &'static Shape,
-    pub(crate) phantom: PhantomData<*mut &'a ()>,
+    pub(crate) phantom: PhantomData<&'facet_lifetime ()>,
 }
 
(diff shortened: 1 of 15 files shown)


Veykril:
Alright, nothing more lifetimes can't fix.

facet-reflect/src/wip/heap_value.rs
```diff
@@ -24,14 +24,14 @@ impl Drop for HeapValue<'_> {
     }
 }
 
-impl<'a> HeapValue<'a> {
+impl<'facet_lifetime> HeapValue<'facet_lifetime> {
     /// Returns a peek that allows exploring the heap value.
-    pub fn peek(&self) -> Peek<'a> {
+    pub fn peek(&self) -> Peek<'_, 'facet_lifetime> {
```
Veykril, line 29: This here specifically dropped the lifetime association of `&self`.

Veykril:
Btw I just realized we probably want to replace the `*mut &'a ()` symbol soup with the `fn(&'a ()) -> &'a ()` symbol soup as the former impacts auto-traits, making the types `!Send` and `!Sync`

fasterthanlime:
> Btw I just realized we probably want to replace the `*mut &'a ()` symbol soup with the `fn(&'a ()) -> &'a ()` symbol soup as the former impacts auto-traits, making the types `!Send` and `!Sync`

Ah yep, luckily there's a central place to do so now :) (BGP)

fasterthanlime (approved):

PR #385, opened 2025-04-24
Add support for `!Sized` types

tyilo:
We probably need to change the definition of `Def::Slice` to be an actual slice (`[T]`) and not a slice refence (`&[T]`).

Ties into #376


tyilo:
Maybe we should merge this first and then work on changing the impls for `&str` and `&[T]` in #376

fasterthanlime:
> Maybe we should merge this first

Sounds good to me!

fasterthanlime (approved):

PR #386, opened 2025-04-24
Use declarative macros for tuple implementations of Facet

42triangles:
- [X] Base implementation
- [x] As far as I can tell, the `Characteristic` check isn't correct (there should be two, one for `Debug` and one for `PartialEq`, not a single one for `Eq`, unless there's something I'm missing which is very possible)
- [x] Implement `default_in_place`, `partial_ord`, `ord`, `hash`, `drop_in_place`, `clone_into`

Further changes:
* `debug` now outputs `(T0,)` instead of `(T0)`
* Some stuff is slightly less short to make it fit the macro, but should be fully equivalent (tuples always have trailing commas in the generated, last & first elements aren't treated differently anymore (`MarkerTraits::all().intersection(T0::SHAPE.vtable.marker_traits)`, `eq` now has a trailing `true` instead of directly returning from the last one)

Build performance appears to stay the same:
|device|generated file|declarative macro|
|-|-|-|
|laptop (noisy, slow) | 4.4s | 4.3s |
| PC | 1.2s | 1.2s |

(it appears it's slightly faster to build if I run `cargo clean; cargo build --timings; cp ../target/cargo-timings/cargo-timing.html cargo-timing-$version.html` in `facet/facet-core`, I assume because it's building less things in parallel, hence the different numbers from #379)


facet-core/src/impls_core/tuple.rs
```diff
@@ -32,2490 +29,214 @@ pub fn write_type_name_list(
...
+                                eq = true
+                            });
+                        }
+
+                        if Characteristic::PartialOrd.all(elem_shapes) {
+                            builder = builder.partial_ord(|a, b| impl_facet_for_tuple! {
+                                ord on ($($elems.$idx,)+),
+                                partial_ord(a, b),
+                                eq = Some(Ordering::Equal)
+                            });
```
fasterthanlime, line 174: these are so clean 😭

please add yourself to the facet-core authors (@veykril you too) in follow-up PRs

fasterthanlime (approved):

PR #387, opened 2025-04-24
Work on `facet-kdl`

TheLostLambda:
Sorry this isn't yet in a mergeable state — I've had limited time this week and I'm going to need to pester @zkat to find a way to make `miette-derive` an optional dependency of `kdl-rs` I think... That's currently tripping `just absolve` as it's quite `syn`ful.

I'll focus on getting that super basic test passing, then I'll make sure it passes CI so it can be merged ASAP!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -245,6 +245,8 @@ dependencies = [
  "facet-core",
  "facet-reflect",
  "facet-testhelpers",
+ "indoc",
+ "kdl",
  "log",
  "num-traits",
  "owo-colors",
@@ -420,6 +422,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "indoc"
+version = "2.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4c7245a08504955605670dbf141fceab975f15ca21570696aebe9d2e71576bd"
+
+[[package]]
 name = "insta"
 version = "1.43.0"
 source = "git+https://github.com/mitsuhiko/insta#34196da9ee1a33df6d9e0a57a47c5692e3c5de8f"
@@ -427,6 +435,18 @@ dependencies = [
  "console",
  "once_cell",
  "similar",
+]
+
+[[package]]
+name = "kdl"
+version = "6.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "12661358400b02cbbf1fbd05f0a483335490e8a6bd1867620f2eeb78f304a22f"
+dependencies = [
+ "miette",
+ "num",
+ "thiserror",
+ "winnow 0.6.24",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


TheLostLambda:
> I'm going to need to pester @zkat to find a way to make miette-derive an optional dependency of kdl-rs I think...

I reckon the keep-syn-out-of-the-dep-tree rule is pretty strict for `facet` stuff, @fasterthanlime ? `kdl-rs` is relying on a derive macro for rich and helpful parse errors currently, though I've also seen that `ariadne` is what's being used in this repository.

How massively difficult is moving from `syn` to `unsyn`? Would you like all of the `facet-*` crates to use `ariadne`?

zkat:
@TheLostLambda there was a convo on mastodon about how I would tbh totally accept a patch to `miette-derive` to port it to `unsyn`.

In the meantime, removing the `derive` feature from the miette dependency in `kdl-rs` should be relatively easy. Implementing `Diagnostic` by hand for the two structs that use it should be trivial.

TheLostLambda:
Keeping this thread updated with upstream work, a PR (https://github.com/zkat/miette/pull/436) has been submitted to remove `syn` from `miette` which `kdl-rs` depends on — this resulted in a ~4.6X compile-time improvement (https://github.com/zkat/miette/issues/435), and makes it possible to remove `syn` from `kdl-rs` next!

fasterthanlime:
> this resulted in a ~4.6X compile-time improvement ([zkat/miette#435](https://github.com/zkat/miette/issues/435)), and makes it possible to remove `syn` from `kdl-rs` next!

Holy shit yesss

TheLostLambda:
Slightly more modest gains for `kdl-rs`, but it builds ~23% faster in debug and doesn't depend on `syn` anymore :) https://github.com/kdl-org/kdl-rs/pull/125#issuecomment-2833296161

Now I can get back to pulling it into `facet-kdl`!

TheLostLambda:
@fasterthanlime This is still largely a mess, but I'm getting closer to something that I can then tidy up an merge as an extraordinarily minimal working demo!

In the meantime, I figured I'd throw a question your way (if you run out of other really difficult things to think about):
```rust
// QUESTION: I don't know when this would be particularly good practice, but it could be nice if `facet` shipped
// some sort of macro that allowed libraries to rename the Facet trait / attributes? This might make it clearer
// what's going on if you're ever mixing several `Facet` libraries that all use different arbitrary attributes? I
// just think that `#[kdl(child)]` would be a lot clearer than `#[facet(child)]` if, say, you also wanted to
// deserialize from something like XML? Or command-line arguments? Those would also need attributes, e.g.
// `#[facet(text)]` or `#[facet(positional)]`, and I think things would be a lot clearer as `#[xml(text)]` and
// `#[args(positional)]`. If, however, it's far too evil or hard to implment something like that, then arbitrary
// attributes should be given "namespaces", maybe? Like `#[facet(kdl, child)]` or `#[facet(xml, text)]`?
//
// Overall I think this is a hard design question, but I do think it's worth considering how several `facet` crates
// relying on arbitrary attributes should interact...
```

fasterthanlime:
> ```rust
> // QUESTION: I don't know when this would be particularly good practice, but it could be nice if `facet` shipped
> // some sort of macro that allowed libraries to rename the Facet trait / attributes? This might make it clearer
> // what's going on if you're ever mixing several `Facet` libraries that all use different arbitrary attributes? I
> // just think that `#[kdl(child)]` would be a lot clearer than `#[facet(child)]` if, say, you also wanted to
> // deserialize from something like XML? Or command-line arguments? Those would also need attributes, e.g.
> // `#[facet(text)]` or `#[facet(positional)]`, and I think things would be a lot clearer as `#[xml(text)]` and
> // `#[args(positional)]`. If, however, it's far too evil or hard to implment something like that, then arbitrary
> // attributes should be given "namespaces", maybe? Like `#[facet(kdl, child)]` or `#[facet(xml, text)]`?
> //
> // Overall I think this is a hard design question, but I do think it's worth considering how several `facet` crates
> // relying on arbitrary attributes should interact...
> ```

This is absolutely a design question, and the answer is, I think, for attributes that have the same meaning across parsers or deserializers or libraries, then they should be global attributes. For things that only make sense for one thing, then they should be namespace attributes. If they're going to be there differently for different formats, then we should have the ability to namespace them. We should have some sort of precedence, right? So if you have something like `facet(kdl, rename = "blah")` then it has precedence over `facet(rename = "blah")`

Library specific attributes are just open right now, so for now... I don't know how to deal with them yet. I would like to get a KDL demo merged in so we can move ahead with some of the larger fundamental changes we want to do, and then we can increment on the actual design.  Breakage is not an issue; we're going to make breaking changes for a couple more months, most likely.

fasterthanlime:
I'm commandeering this PR to make sure it's merged before a big restructure — hope that's fine! I'll get it to a mergeable state in another branch — even if it's nonfunctional, I just want to sure we carry the code with us.

fasterthanlime:
For the future, @TheLostLambda — please don't do merge commits. Always rebase.

You can use this as an opportunity to learn https://jj-vcs.github.io/jj/latest/, it's always fun! But yeah, no merge commits. I want all PRs to be cleanly applied to main. You never know what hides in merge commits...

fasterthanlime:
Merged (in progress) as part of #469 — to avoid code rot as we move things around.

TheLostLambda:
Thanks for handling this @fasterthanlime ! Good to know about merge commits and I was already looking to play around with Jujutsu, so it's the perfect opportunity!

I should have some more progress that I'll put in a new PR over the next couple of days :)

I've learned a lot from working on this repo already!!!

Issue #381, opened 2025-04-24
Generic way to push smartpointers with wip


tversteeg:
Would it be possible/reasonable to have something similar like `wip.push_some()` for smart pointers, maybe `push_smart_pointer()`. I assume most consumers of the facet-reflect crate will not want to care about the implementation details of different smart pointers.

If this is not possibly, how would you recommend them to be handled when using `Wip<'a>`? I have been trying some things for about an hour, but I can't really figure out a proper solution to support them.

fasterthanlime:
I think the way I would go about it would be... `.field()` or `.push()` or whatever builds a `T` in storage somewhere (a special "SmartPointer" kind of frame), and on `.pop()` the smart pointer is initialized with that `T`

because we can't really build an `Arc<MaybeUninitialized<T>>` and then start writing to `T` right? we have to have that temp storage, just like we have for K/V pairs when filling a map, for example.

tversteeg:
> because we can't really build an `Arc<MaybeUninitialized<T>>` and then start writing to `T` right? we have to have that temp storage, just like we have for K/V pairs when filling a map, for example.

Why can't we do that? As with `wip.push_some()`, wouldn't you always want to initialize the smart pointer when deserializing some data format?

fasterthanlime:
> Why can't we do that?

Ah, I guess https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.get_mut is a thing.. I thought writing into the `T` of an `Arc<T>` would be UB, I'm not sure!

fasterthanlime:
Fixed in #390 — you can just push the inner type now.

PR #390, opened 2025-04-25
🧨 (big one) Introduce `Shape::inner` for (Utf8PathBuf, NonZero<u8> etc.) along with try_from_inner, try_into_inner, borrow_inner etc.

fasterthanlime:
I'm basically trying to come up with a clean solution to this type of errors:

There's an error where the deserializer faces-json is expecting an enum or a string and it's getting a shape of type MyString which is a transparent wrapper over strings.

Where MyString is a newtype over the other.


tversteeg:
This is great! Do you think some of the smart pointer types, as discussed in #381 could also fit in?

fasterthanlime:
> Do you think some of the smart pointer types, as discussed in #381 could also fit in?

Maybe it would be good to merge smart pointers and other transparent types?? Let's find out.

PR #403, opened 2025-04-25
fix(reflect): make rename an actual rename instead of an alias

tversteeg:
Without this change the new test fails.

### DIFF
--- a/facet-json/tests/integration/read/rename.rs
+++ b/facet-json/tests/integration/read/rename.rs
@@ -467,3 +467,28 @@ fn test_field_rename_missing_required_error() {
     #[cfg(not(miri))]
     assert_snapshot!(e.to_string());
 }
+
+/// Rename to verify it's not an accidental alias
+#[test]
+fn test_field_rename_not_alias() {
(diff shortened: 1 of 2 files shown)


fasterthanlime:
I will review when the tests are successful!

facet-json/tests/integration/read/rename.rs
```diff
@@ -467,3 +467,28 @@ fn test_field_rename_missing_required_error() {
...
+
+        #[facet(rename = "c")]
+        b: String,
+    }
+
+    let json = r#"{"b":"focus group 1","c":"focus group 2"}"#;
+
+    let result: ABTesting = match from_str(json) {
+        Ok(s) => s,
+        Err(e) => panic!("Error deserializing JSON: {}", e),
```
fasterthanlime, line 489: Note that returning an error (having the test return `eyre::Result<()>`) gives us a nice display implementation of our JSON errors for free, so I would recommend that for future changes.

fasterthanlime (approved):

PR #405, opened 2025-04-25
fix: jsonschema doctest breakage

lmmx:
- Close #404

PR #406, opened 2025-04-25
test(jsonschema): repair doc-test by using facet main crate directly instead of sub-crates

### LINKED ISSUE #404: [BUG] Doctests broke in v0.16.0 (`facet-jsonschema` lacks `facet` crate dev dep)
This can be fixed by adding the workspace `facet` crate to the dev deps, like in `facet-yaml`, `facet-kdl`, `facet-toml`

tversteeg:
Strangely enough, this actually improved cold build times on my machine from 2.4 secs to 2.1.

Shall I apply this everywhere?

Closes #404.

If you take a look at this, please also first take a look at #405.

### DIFF
--- a/facet-jsonschema/Cargo.toml
+++ b/facet-jsonschema/Cargo.toml
@@ -20,9 +20,7 @@ keywords = [
 categories = ["development-tools", "encoding"]
 
 [dependencies]
-facet-core = { version = "0.16.0", path = "../facet-core" }
-facet-derive = { version = "0.16.0", path = "../facet-derive" }
-facet-reflect = { version = "0.16.0", path = "../facet-reflect" }
+facet = { version = "0.16.0", path = "../facet" }
 
 [dev-dependencies]
 serde_json = "1.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -251,9 +251,7 @@ dependencies = [
 name = "facet-jsonschema"
 version = "0.16.0"
 dependencies = [
- "facet-core",
- "facet-derive",
- "facet-reflect",
+ "facet",
  "schemars",
  "serde_json",
 ]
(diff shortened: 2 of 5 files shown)


lmmx:
It looks like your way shaves 5 seconds off but the test job is notably longer, I suspect as it's uncached? Perhaps you could trigger it again and it'd run faster?

fasterthanlime:
`facet` is fine as a dev dep, it's just never fine as a regular dep (of libraries like these).

Removing it as a regular dependency is what caused this, I believe.

fasterthanlime:
> Shall I apply this everywhere?

@epage warned specifically against this so I'm going to go with no :(

facet-jsonschema/Cargo.toml
```diff
@@ -20,9 +20,7 @@ keywords = [
 categories = ["development-tools", "encoding"]
 
 [dependencies]
-facet-core = { version = "0.16.0", path = "../facet-core" }
-facet-derive = { version = "0.16.0", path = "../facet-derive" }
-facet-reflect = { version = "0.16.0", path = "../facet-reflect" }
```
epage, line 25: > @epage warned specifically against this so I'm going to go with no :(

The difference here is that `facet-jsonschema` depends on the derive while the other cases do not.  A worthwhile question is how important using the derive is to this crate but I would consider jsonschema lower priority than cases like json.

I do wonder what made the difference in build times.  This should have made the build a little more serial but have no other impact.  I suspect it has to do with `cargo hack` and either it added more cases for some reason or it built things in an order that had better reuse of cache with feature activations (unsure how much of that gets thrown out for local builds or reused).

epage:
> repair doc-test by using facet main crate directly instead of sub-crates

That could also be solved by having a dev-dependency on `facet`.

epage:
Oh that was done in #405.

PR #415, opened 2025-04-25
Serialize struct formed enums more consistently

ThouCheese:


facet-json/src/deserialize.rs
```diff
@@ -504,7 +501,10 @@ pub fn from_slice_wip<'input: 'facet, 'facet>(
                                 Some((index, variant)) => {
                                     trace!("Variant {} selected", variant.name.blue());
                                     reflect!(variant(index));
-                                    transparent = true;
+                                    // An enum is transparent only if it
+                                    // 1. Is a tuple variant, i.e. MyEnum::A(2), not MyEnum::A { a: 3 }
+                                    // 2. Has one field, i.e. not MyEnum::A(2, 3).
+                                    transparent = crate::variant_is_transparent(&variant);
```
ThouCheese, line 507: I think I need to do something more here? Because we need to add a new level of nesting (`{` and `}`) if the current enum is not transparent right?

fasterthanlime:
@ThouCheese I'm rebasing this and taking a look

fasterthanlime:
Here's a helpful incantation to help debug this:

CleanShot 2025-04-26 at 11 24 13@2x

Copy-pastable:

```
RUST_BACKTRACE=full nrc json_read_struct_variant -F facet-reflect/log
```

Issue #408, opened 2025-04-25
Remove all authors


fasterthanlime:
cf. https://github.com/facet-rs/facet/commit/538ebc80540850f66d27b416f9f1f04bd7873ece#r155889635

The field is immutable. Metadata is better surfaced through github, etc.

PR #433, opened 2025-04-26
fix(docs) Remove reference to old process

sinon:
## Why?

https://github.com/facet-rs/facet/issues/408 removed using the `Cargo.toml` metadata to highlight contributors but `DEVELOP.md` still referenced old process

## What?

Remove incorrect documentation, I couldn't find a existing issue about what will replace this so opted for full removal instead of replacing with `# TODO` / issue link.


fasterthanlime:
Makes sense, thanks!

PR #434, opened 2025-04-26
Expose a typed VTable API, reduce unsafe code

Veykril:


facet-core/src/impls_alloc/vec.rs
```diff
@@ -28,38 +28,34 @@ where
                             }
                         })
                         .default_in_place(|target| unsafe { target.put(Self::default()) })
-                        .clone_into(|src, dst| unsafe { dst.put(src.get::<Self>()) });
+                    // FIXME: THIS IS VERY WRONG
+                        .clone_into(|src, dst| unsafe { dst.put(core::ptr::read(src)) });
```
Veykril, line 32: All `clone_into` implementations for non-copy types are wrong. All they do is memcopy which means they can easily cause double drops, aliasing and more fun stuff for collections.

I didn't fix it in this PR yet, so this just preserves the wrong implementations

fasterthanlime: I'm not overly surprised. A lot of this was done pretty carelessly by myself a while back. Thank you for taking this on so much.

fasterthanlime (approved):
Looks super neat, thanks for thatr!

PR #440, opened 2025-04-26
Revert changes to pre-commit hook introduced in 4589057

gennyble:
Noticed this when trying to commit for #438. I just deleted the `pre-commit` then. Looks like this was changed here, https://github.com/facet-rs/facet/commit/45890573d564859e49d1d0916d36af40acb07eb0, and was a find-replace error? `facet-dev` does not currently have any mention of "precommit" anywhere.

### DIFF
--- a/.cargo-husky/hooks/pre-commit
+++ b/.cargo-husky/hooks/pre-commit
@@ -1,2 +1,2 @@
 #!/usr/bin/env sh
-cargo run -p facet-dev precommit
+cargo run -p facet-dev -- generate


fasterthanlime:
My bad, I meant to run `cargo run -p facet-dev -- generate`

That's what we should run as precommit

fasterthanlime:
Also, fair warning, `cargo husky` only sets up hook the First time you run cargo test — you might have to `cp .cargo-husky/hooks/* .git/hooks/` yourself after fixing this

fasterthanlime:
Woops, I was 1 minute too fast, sorry @gennyble !! Thanks for raising this — it's fixed in main.

gennyble:
> Woops, I was 1 minute too fast, sorry @gennyble !! Thanks for raising this — it's fixed in main.

No problem, I was too slow, haha!

Is it worth testing the hooks in CI? That sounds silly but walking into a new codebase and a git-hook failing is jarring. I can try and write a test.

fasterthanlime:
I'm going to say it's not worth it because we have enough contributors that it would be surfaced almost immediately

fasterthanlime:
But on the other hand, it's kind of harmless so if you wanna add a test, I'll merge it!

PR #452, opened 2025-04-26
json: Support facet(flatten), make serializer iterative

ThouCheese:
I think there will be more work to be done on this but:

1. Switches the serializer from a recursive one to an iterative one, allowing us to serialize infinitely deeply nested objects.
2. Add attributes to enum variants
3. Support flattening structs and enums into structs


ThouCheese:
Meh I want it to be less code

fasterthanlime:
I commandeer this PR! 🏴‍☠️  (I'm going to make test pass, rebase it, and merge it)

PR #458, opened 2025-04-27
refactor(toml)!: setup structure for serialize implementation

tversteeg:
Moved the serialize and deserialize implementations to different directories, similar to facet-json.

Sorry for the big PR, but this actually contains very little substantive changes, most is just the moving of files.


fasterthanlime:
Yeah, those refactors are useful, so no worries.  :)

fasterthanlime (approved):

facet-toml/src/serialize.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[cfg(feature = "alloc")]
+use alloc::string::String;
+
+use facet_core::{Def, Facet, FieldAttribute};
+use facet_reflect::{Peek, ScalarType};
+use toml_write::TomlWrite;
+
+/// Implement writing TOML values to types that allow being written to.
+pub trait TomlSerialize: TomlWrite {
```
epage, line 13: I see little value in putting `TomlWrite` in the public API or in offering this through an extension trait

tversteeg: I thought it would be useful so that it is automatically implemented for all types that `TomlWrite` also implements. But since the types are fairly restricted, maybe there's indeed not much value.

PR #460, opened 2025-04-27
feat(args) - Handle multi word args

sinon:
### Problem

Using common casing for multi-word arguments results in error. As the mismatch in common kebab-case used for arguments does not find the snake_case Facet field.

### What

Normalize the keys for non-short named arguments to snake case. This initial work makes 2 assumptions:
- GNU/Unix kebab-casing is the initial norm
- Facet derived struct will be built with snake_cased field names

I will hopefully tackle making these less implicit assumption by aligning with the json crate on how it is tackling this casing aspect

Note: This is stacked on: https://github.com/facet-rs/facet/pull/459 both PRs are so small that I don't think this complicates review.

### DIFF
--- a/facet-args/src/lib.rs
+++ b/facet-args/src/lib.rs
@@ -4,6 +4,9 @@
 #![forbid(unsafe_code)]
 #![doc = include_str!("../README.md")]
 
+extern crate alloc;
+use alloc::borrow::Cow;
+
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Yeah, it looks like GitHub didn't like this tact branch. That's my bad, I should have handled it differently!

PR #461, opened 2025-04-27
Add support for rename_all and arbitrary attributes on variants

### LINKED ISSUE #417: Wishlist: support attributes on structs and enums
I am working on adding ASN.1 support to Facet, starting with BER/DER. Progress is good so far. Like any future XML encoding (which will definitely also benefit from this), a small amount of metadata is required in order to encode tag-length-value bytes correctly. For the purposes of this issue, just ignore the invariants on `IA5String`.

The following example works nicely for setting the field's tag based on an attribute:
```rust
// FooQuestion ::= SEQUENCE {
//   trackingNumber INTEGER,
//   question       IA5String
// }
#[derive(Debug, Facet, PartialEq, Eq)]
struct FooQuestion {
    tracking_number: i8,
    #[facet(asn1 = "[UNIVERSAL 22]")]
    question: String,
}
```

This isn't yet possible for enum variants, only for the fields of the enum variant:
```rust
// This doesn't work (yet!)

// FooAnswer ::= CHOICE {
//   someanswer [0] SEQUENCE {
//     trackingNumber INTEGER,
//     answer IA5String,
//   }
//   rejection [1] SEQUENCE {
//     trackingNumber INTEGER,
//   }
// }
#[derive(Debug, Facet, PartialEq, Eq)]
enum FooAnswer {
    #[facet(asn1 = "[0]")] // Can't do this...
    SomeAnswer {
        tracking_number: i8,
        #[facet(asn1 = "[UNIVERSAL 22]")]
        answer: String,
    },
    #[facet(asn1 = "[1]")] // ...which is a pain!
    Rejection {
        tracing_number: i8,
    },
}
```

It also means that simple cases like unit structs have to be wrapped in a newtype pattern if they need to be tagged:
```rust
// This works, but it isn't very pretty...

// [3] NULL
#[derive(Debug, Facet, PartialEq, Eq)]
struct MyAsn1NullTypeWrapper(#[facet(asn1 = "[3]")] MyAsn1NullType);

#[derive(Debug, Facet, PartialEq, Eq)]
struct MyAsn1NullType;

// ...this doesn't work at all

// [3] NULL
#[derive(Debug, Facet, PartialEq, Eq)]
#[facet(asn1 = "[3]")]
struct MyAsn1NullType;
```

RoastVeg:
Fixes #417

### DIFF
--- a/facet-core/src/types/enum_.rs
+++ b/facet-core/src/types/enum_.rs
@@ -67,6 +67,9 @@ pub struct Variant {
     /// Discriminant value (if available). Might fit in a u8, etc.
     pub discriminant: i64,
 
+    /// Attributes set for this variant via the derive macro
+    pub attributes: &'static [VariantAttribute],
+
     /// Fields for this variant (empty if unit, number-named if tuple).
     /// IMPORTANT: the offset for the fields already takes into account the size & alignment of the
     /// discriminant.
@@ -87,6 +90,7 @@ impl Variant {
 pub struct VariantBuilder {
     name: Option<&'static str>,
     discriminant: Option<i64>,
+    attributes: &'static [VariantAttribute],
     fields: Option<StructDef>,
     doc: &'static [&'static str],
 }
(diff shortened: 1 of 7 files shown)


fasterthanlime:
Merged in #467

Issue #436, opened 2025-04-26
Rationale behind `Def` enum nature and its stability


h33p:
First of all, what a delight of a crate!

# Background

I am working on a [trait-based plugin system with language interop](https://github.com/h33p/cglue) and am looking to migrate from `abi_stable` into something that is more minimal and elegant for the purposes of comparing type layout compatibility. In addition, having something that works better than `cbindgen` for generating bindings would be great (as `cbindgen` parses the source tree, and is generally quite fallible).

A while back I wrote a [simple PoC for compile time type information](https://github.com/h33p/ctti), and since `facet` takes the same idea of lifting type information into an associated const, but has fleshed it out into a whole ecosystem, I figured it would make sense to build atop of it.

Now, what I want to build is equivalent of `abi_stable`'s [`check_layout_compatiblity`](https://docs.rs/abi_stable/latest/abi_stable/abi_stability/fn.check_layout_compatibility.html). For that, I need to express underlying type layout in a ABI-stable manner, so that I can pass 2 type layouts coming from different shared libraries (potentially compiled with different Rust versions). In the future, I may choose to do a `cbindgen` equivalent and spit out `C`/`C++` code directly from everything that touches a `Facet`-deriving type.

However, after checking `facet`'s internals, it appears that making a layout that can remain stable across versions of both Rust and the type-checker itself, does not seem to be as straightforward as I thought it would be.

# `Def` enum

Looking at the `Def` enum:

```rust
#[non_exhaustive]
#[repr(C)]
pub enum Def {
    Scalar(ScalarDef),
    Struct(StructDef),
    Enum(EnumDef),
    Map(MapDef),
    List(ListDef),
    Array(ArrayDef),
    Slice(SliceDef)),
    Option(OptionDef),
    SmartPointer(SmartPointerDef),
    FunctionPointer(FunctionPointerDef),
}
```

My first impression is that it's oddly arbitrary. If we followed the [Rust reference](https://doc.rust-lang.org/reference/types.html), `Def` would look like so:

```rust
// Exhaustive!
#[repr(C)]
pub enum Def {
    Primitive(PrimitiveDef),
    Sequence(SequenceDef),
    User(UserDef),
    Function(FunctionDef),
    Pointer(PointerDef),
}
``` 

[...]

Issue #455, opened 2025-04-27
Non-`'static` `Shape`s


djkoloski:
I was looking at facet to get an idea of what it does and why it's unique. A cool property of schema-oriented systems is that type schemas can be loaded at runtime. I'd be interested in experimenting with runtime schemas, which in facet would be `Shape`s created at runtime. It looks like `Shape` currently uses exclusively `'static` references, but it could just be parameterized over a lifetime to allow for this. Is that something you'd be interested in seeing the code impact of? I don't want to jumpscare you with a draft PR that changes such a fundamental part of the library. And thanks for all your hard work!

fasterthanlime:
Note that #462 is another large-scale change to facet and will definitely impact any exploratory work you do there @djkoloski

fasterthanlime:
Okay, so this is in main now (cc @djkoloski) — I still think we need to do a little "lifetime review", but it should allow some interesting stuff, I think. We need to come up with a little demo for it!

PR #462, opened 2025-04-27
Rework type information (Def)

h33p:
This is related to #436, but merging may not necessarily close it. Seems also related to #376.

The primary purpose of this work is to lift type information _as it is_, rather than speculating on the exact semantics of each type, at a case-by-case basis. The benefits primarily come in the following forms:

- Defining types in accordance to the [type system](https://doc.rust-lang.org/reference/types.html) makes it easier to reason about the types.
- We enable a stable layout for verifying type information (`abi_stable` functionality).
- We enable code generation based on exact type representation (`cbindgen` functionality).
- More optimal `Shape` recursion - we are able to store `&Shape` in all Kinds, but `Kind::Pointer`.
  - This enables more Shape-related work to be done at const time!

There are several broad TODOs left:

- [ ] Design considerations.
- [ ] Extract the type-semantic information from `Def` and put it under `Semantics`.
  - [ ] Intermediate merge here?
- [ ] Switch from `Def` to `Kind` and `Semantics` in all dependents.
- [ ] Remove `Def`.
- [ ] Final cleanup (e.g. rename remaining `...Def` to `...Kind`, and change as many `fn() -> &Shape` fields to `&Shape`, docs, tests).

## Design considerations

Before proceeding further, there are several design and detail considerations I'd need some input on:

- [ ] In `UserKind`, do we need to extract the `#[repr(align(N))]`, if we can get alignment by other means (`ShapeLayout`)?
- [ ] Do we need to include `bits` in `IntegerKind`/`FloatKind`? The information is already computable from the layout.
- [ ] Do we split special (built-in) types, e.g. `NonZero` away from `User`, and represent them explicitly?
- [ ] Should we de-embed discriminant from field offsets within enums (by having separate offset-to-union value)? This would allow us to express `#[repr(Rust)]` enums, and allow full inspection of null-pointer-optimized Rust enums, like `Option<T>` without option-specific code.
- [ ] Do we assign `Shape::inner` on raw pointers?
- [ ] Tuples cannot have attributes or docs attached, should it contain `&[&Shape]` instead?
- [ ] Is it okay to expect `Unsized` shape vtables take the parent pointer (pointer to the wide pointer of the unsized object)?

## Extracting type-semantic information

I considered 2 approaches:

- Extracting semantics from `Def` into enum `Semantics`/`Sem`.
- Putting `Semantics` as an opaque object, and providing a getter for concrete semantics object.

I initially thought the latter could be cool for custom functionality, e.g. generating custom wrappers for a type in a hypothetical `C++` bindings generator, but then I figured through #417 we could embed arbitrary type-bound values on a `Shape`.

Extracting semantics from `Def` is arguably simpler, and wouldn't sacrifice existing functionality.


fasterthanlime:
@h33p I have yet to take a look but the description looks promising! 

Does this overlap with https://github.com/facet-rs/facet/issues/455 at all?

h33p:
> Does this overlap with https://github.com/facet-rs/facet/issues/455 at all?

I am touching types that reference Shapes, so there is, but not much beyond that (I don't see how moving `fn() -> &'a Shape` to `&'a Shape` would change anything in a fundamental capacity).

fasterthanlime:
Just answering some of the design questions:

> * [ ] In `UserKind`, do we need to extract the `#[repr(align(N))]`, if we can get alignment by other means (`ShapeLayout`)?

No, there's no need to duplicate the information.

> * [ ] Do we need to include `bits` in `IntegerKind`/`FloatKind`? The information is already computable from the layout.

I think it was just being fancy and not knowing where to put somethings — no need to include them.

> * [ ]  Do we split special (built-in) types, e.g. `NonZero` away from `User`, and represent them explicitly?

I think that's what I was trying to get at with affinity: that we should be able to tell that they're those special types, but if you want to represent them as they are, then they should have that `repr(transparent)` wrapper on top. And we need that for deserialization anyway.

> * [ ]  Should we de-embed discriminant from field offsets within enums (by having separate offset-to-union value)? This would allow us to express `#[repr(Rust)]` enums, and allow full inspection of null-pointer-optimized Rust enums, like `Option<T>` without option-specific code.

I'm confused about the claim that we can express `#[repr(Rust)]` enums — As far as I can tell, in stable Rust, it's a no-go. I'm also not clear how de-embedding the discriminant helps with that, but let's maybe tackle that in a follow-up PR since I think it's a complicated topic. 

> * [ ]  Do we assign `Shape::inner` on raw pointers?

I think you mentioned somewhere, and your instinct is right, that pointers are the only way that we can have recursive types, and so that's the only place where we need to have indirection through a function. 

> * [ ]  Tuples cannot have attributes or docs attached, should it contain `&[&Shape]` instead?

That's a fair point, but having their offsets is very convenient for reflection. So, even though you can technically get the offset of every tuple field by looking at the alignment of every shape inside of it, I would still like to have those statically computed. 

> * [ ]  Is it okay to expect `Unsized` shape vtables take the parent pointer (pointer to the wide pointer of the unsized object)?

I think so, it should be documented but I think so.

---

Because it is such a large change, I think we might proceed the way I did other large changes before, which is to merge something that changes the core and comment out all the other crates from the workspace at first. And then keep merging PRs that uncomment and port crates roughly in that order. First core, then derive, then reflect, then pretty, then JSON, and then the other formats. 

Hopefully, porting reflect will unlock the other formats pretty immediately? There's probably little reason for it to change.

tversteeg:
> Hopefully, porting reflect will unlock the other formats pretty immediately? There's probably little reason for it to change.

The removal of `Def` will break all format crates since they all directly depend on it.

I can port over facet-toml, and maybe some of the small format crates if needed.

h33p:
Thanks for your input! Most of these are fairly self-explanatory, I just wanted to circle back to 

> I think that's what I was trying to get at with affinity: that we should be able to tell that they're those special types, but if you want to represent them as they are, then they should have that repr(transparent) wrapper on top. And we need that for deserialization anyway.

Currently, `NonZero<$type>` is defined as follows:

```rust
                    .kind(Kind::User(UserKind {
                        repr: Repr {
                            base: BaseRepr::Transparent,
                            packed: false,
                            align: None,
                        },
                        variant: UserVariant::Struct(StructDef {
                            kind: StructKind::TupleStruct,
                            fields: &const {
                                [Field::builder()
                                    .name("0")
                                    // TODO: is it correct to represent $type here, when we, in
                                    // fact, store $type::NonZeroInner.
                                    .shape(|| <$type>::SHAPE)
                                    .offset(0)
                                    .flags(FieldFlags::EMPTY)
                                    .build()]
                            },
                        }),
                    }))
```

This is a bit inconsistent, since the underlying type is not a `usize`, for instance, but an internal implementation detail - [`NonZeroUsizeInner`](https://doc.rust-lang.org/src/core/num/nonzero.rs.html#74). This is not per say about semantics, but actual type representation - we can only stabily represent internal structure of `NonZero` in the following cases:

- Arbitrarily decide to map it to the underlying primitive (diversion).
- Capture unstable [`nonzero_internals`](https://doc.rust-lang.org/src/core/num/nonzero.rs.html#27) (unstable).
- Have a dedicated `Kind::Lang(LangKind::NonZero(<$type>::SHAPE))` branch.

> I'm confused about the claim that we can express #[repr(Rust)] enums — As far as I can tell, in stable Rust, it's a no-go. I'm also not clear how de-embedding the discriminant helps with that, but let's maybe tackle that in a follow-up PR since I think it's a complicated topic.

There's a special case:

```rust
enum Enum<T> {
    Something(T),
    Nothing,
}
```

See - "null pointer optimization" [here](https://doc.rust-lang.org/nomicon/repr-rust.html).

If all-zero is not a valid representation for `T`, then `size_of::<Enum<T>>() = size_of::<T>()` (and there's no discriminant). This is great for `Option<&T> <--> *const T` interop. But you're right, we can deal with this later, and you're also right, de-embedding the discriminant wouldn't help much here (plus, [offset_of_enum](https://doc.rust-lang.org/nightly/unstable-book/language-features/offset-of-enum.html) returns offsets with the discriminant).

> I think you mentioned somewhere, and your instinct is right, that pointers are the only way that we can have recursive types, and so that's the only place where we need to have indirection through a function.

This is a bit different, since `inner` is a separate field in `Shape`. For `&mut T` I already assign `inner`, since that's good for ser/des, but raw pointers may not always point to valid dereferenceable data, hence my hunch is against filling in the `inner` (both raw and reference pointers still point to the inner type through `Kind`).

> Because it is such a large change, I think we might proceed the way I did other large changes before, which is to merge something that changes the core and comment out all the other crates from the workspace at first. And then keep merging PRs that uncomment and port crates roughly in that order. First core, then derive, then reflect, then pretty, then JSON, and then the other formats.

Sounds good! I will work out the `Semantics` part and comment out the rest of the crates. Do we want this called `Semantics` or `Sem` to make it as short as `Def` was?

fasterthanlime:
Alright, my new position on non-zero is that I don't know what the right thing is to do. The null pointer optimization thing makes sense for me. As for recursive types, I'm just scared you will run into the same limitations of `const fn` that I did and realize why we needed the indirection through a function in the first place, but if you don't, then that's great. 

> Sounds good! I will work out the `Semantics` part and comment out the rest of the crates.

We can name it `Sem`, that's fine — because we had a few large PRs pending, I created a bunch of milestones I'd like to get merged before I start "blocking other contributions because of the restructure":

  * https://github.com/facet-rs/facet/milestones

I'm almost done with the `Flatten` one, and I'll get started with the "common basis for facet-json and facet-msgpack" right after — the aim is to make that big restructure less painful down the line.

Are you fine doing exploratory work in the meantime, waiting for my go to try & rebase/redo the work on top of main?

h33p:
Thank you for the input! I'll get on to work, rebasing is not a problem from my end.

fasterthanlime:
> Thank you for the input! I'll get on to work, rebasing is not a problem from my end.

First milestone is done, I'll ping you again when the "common base for serialization/deserialization" is done :)

h33p:
Pushed new changes!

- Rework `Def` into `Sem`
  - Remove some `Sem` variants
    - Have no vtable or additional description - `Struct`, `Enum`, `FucntionPointer`.
    - Can match against `Kind::Sequence(Slice)` and `Kind::Sequence(Array)`.
  - Maintain certain `Shape` getters
    - k, v in `MapSem` (added extra indirection here)
    - t in `ListSem`
    - t in `Slice` (removed indirection since this is now unsized)
      - [ ] Should we switch slice vtable to be `len`, and `as_ptr`, and then compute element pointer from those and the target shape? This would be orthogonal to `Kind::Sequence(Array)`
    - Pointees and stuff in `SmartPointerSem`
  - Rename `StructKind` to `StructVariant`.
  - Remove `&[T]` and `&str` `Facet` impls.
  - Switch `&Path` to `Path` for `Facet` impl.
  - Swap `fn() -> &Shape` to `&Shape` in `Field`.
  - For `Option<T>`, where `T` is NPO, I express the internal enum of `Option` with `EnumRepr::RustNPO`.
    - Might want to embed `nullable` field to a `Shape`, instead of relying upon `size_of` heuristic. But that could be left for later. I can also remove this code for now.

Have some questions:

- [ ] Should we switch `ListSem` and `SliceSem` VTables to have `as_ptr` method, instead of `get_item_ptr`?
  - Manual element address calculations are now necessary on `Kind::Sequence(Array)` anyways. It would put all slice-like in-line.
- [ ] Ergonomics of `Kind`. `Def::Struct(sd)` becomes `Kind::User(UserKind { variant: UserVariant::Struct(sd), .. })`. We could lift special getters for `struct/enum/union/etc.` within `Kind`. Is there a more preferred way?
- [ ] Naming is hard. `Kind...` clashes with existing `StructKind`.
  - Currently renamed `StructKind` to `StructVariant`, but there's still `UserVariant` that embeds `StructKind`/`EnumKind`/etc. Confusing.
  - Should `Kind` be renamed to `Type`? Then reference `ty: Type` on `Shape`. Any alternatives?

fasterthanlime:
> Should we switch `ListSem` and `SliceSem` VTables to have `as_ptr` method, instead of `get_item_ptr`?
> * Manual element address calculations are now necessary on `Kind::Sequence(Array)` anyways. It would put all slice-like in-line.

Sure.

> * Ergonomics of `Kind`. `Def::Struct(sd)` becomes `Kind::User(UserKind { variant: UserVariant::Struct(sd), .. })`. We could lift special getters for `struct/enum/union/etc.` within `Kind`. Is there a more preferred way?

I don't know if facet-core should do the heavy lifting or if facet-reflect should. No preferred way. I agree it's a bit verbose now 😅

> Naming is hard. `Kind...` clashes with existing `StructKind`.
> * Currently renamed `StructKind` to `StructVariant`, but there's still `UserVariant` that embeds `StructKind`/`EnumKind`/etc. Confusing.
> * Should `Kind` be renamed to `Type`? Then reference `ty: Type` on `Shape`. Any alternatives?

Can't see any alternatives for now — follow your heart for now, maybe we'll find something better later? We're mega-breaking things anyway.

h33p:
Addressed most of the things, the core changes are pretty much done.

- Renamed `Kind` to `Type`.
- Embedded `Repr` in `StructType`/`EnumType`/`UnionType` to remove `UserSubtype`.
  - Rename enum's `repr` to `enum_repr`.

Question:

Perhaps `Def` doesn't need to die after all? The main change is removal of `Struct` and `Enum` variants. The rest is roughly the same. I can rename `Sem` back to `Def`, if that were to be more convenient.

fasterthanlime:
> Question:
> 
> Perhaps `Def` doesn't need to die after all? The main change is removal of `Struct` and `Enum` variants. The rest is roughly the same. I can rename `Sem` back to `Def`, if that were to be more convenient.

I guess so? Whatever makes it easy to port the format crates over to the new structure is very welcome.

I'm merging facet-serialize in pretty soon: https://github.com/facet-rs/facet/pull/480 — but it's still pretty rough around the edges.

Ideally there'd be a similar facet-deserialize crate, I'm not quite sure how to approach it yet.

fasterthanlime:
I need to take a serious look at those changes. I've mostly been following the discussion so far, but I haven't looked at the code, so I have no idea how close this is to being mergeable, especially given format crates. But I think I'm done cleaning up other areas of the code, except for coming up with a facet-deserialize crates to lessen the pain of migrating to the new structure even more. 

@h33p When you look at facet-reflect, do you see anything that couldn't be done via the new facet core structure? Because if not, then we can get that PR to a mergeable state without causing too much trauma on the format side of things.

On your end, once this is merged, what is the next step towards stable ABI and whatnot? Do you want any of this to be part of the facet organization? Do you want to develop this completely separately? What do you want the communication lines to look like after this is merged?

h33p:
> When you look at facet-reflect, do you see anything that couldn't be done via the new facet core structure? Because if not, then we can get that PR to a mergeable state without causing too much trauma on the format side of things.

I just had a look at reflect today. What I found was that various vtable entries were not being forwarded by references, so I had to introduce the following change:

- In order to forward trait impls in pointer vtables, we add `const VTABLE: &'static ValueVTable` within `Facet`.
  - `ShapeBuilder` must now be associated with `T: Facet`, which ensures correct `vtable` being set on the `SHAPE`. This might need to be changed when we add non-static Shapes, but for now it is a hard requirement.

In addition, I renamed `Sem` back to `Def`.

Aside from that, there are currently some errors that need to be cleaned up:

```
$ cargo test -p facet-reflect
<...>
failures:
    wip::option_leak::wip_option_testleak1
    wip::option_leak::wip_option_testleak2
    wip::put_into_tuples::test_put_into_tuples

test result: FAILED. 115 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

But these seem to merely be edge cases rather than fundamental design flaws. Not 100% sure if I'm the best person to correct the `facet-reflect` code, though.

> On your end, once this is merged, what is the next step towards stable ABI and whatnot?

I have set up an experiment crate that I shall work on. It will probably pull out and reimplement a structure that holds `Type+ShapeLayout` equivalents, and filter out other fields, because I don't believe maintaining ABI stability on `Facet` trait itself makes much sense (too easy to break + we currently can't derive facet on facet-core structures).

> Do you want any of this to be part of the facet organization? Do you want to develop this completely separately?

I don't mind either way, in fact, I would probably prefer working in-tree. This project being proactive with accepting contributions really helps :)

> What do you want the communication lines to look like after this is merged?

Not entirely sure what you mean by this. If I were to maintain `facet-layout` (that's the name I came up with for it, subject to change), then getting tagged on a issue/PR would be sufficient. At the same time, I'm not entirely sure if I can keep up with the pace at all times. In ideal world, there shouldn't be much need for `facet-layout` to change on `facet-core` changes. There are of course, many additional features to work on later down the line (e.g. [abi_stable's prefix types](https://docs.rs/abi_stable/latest/abi_stable/docs/prefix_types/index.html) equivalent).

fasterthanlime:
So, am I understanding right that you're happy with the current state of this PR? And if I were to take it over, rebase it, and fix up facet-reflect, merge it, and then invite you to the org, you would be able to then develop facet-layout in-tree, and we'd communicate via GitHub issues. Is that correct? 

If you need more time or want to get more things in, that's fine. But I'm kind of eager to get this merged just to avoid further drifting :) Anything that can be left to a follow-up PR, as long as it's not another huge breaking change, is, well, better left to a follow-up PR!

h33p:
> So, am I understanding right that you're happy with the current state of this PR? And if I were to take it over, rebase it, and fix up facet-reflect, merge it, and then invite you to the org, you would be able to then develop facet-layout in-tree, and we'd communicate via GitHub issues. Is that correct?

Sounds good! I have rebased the latest changes now, all that's left is fixing facet-reflect :)

> If you need more time or want to get more things in, that's fine. But I'm kind of eager to get this merged just to avoid further drifting :) Anything that can be left to a follow-up PR, as long as it's not another huge breaking change, is, well, better left to a follow-up PR!

I think all this should be pretty much ready now!

fu5ha:
Exciting PR!

fasterthanlime:
I've commandeered this PR — it's squashed in #527, and I'm going to attempt to fix the other crates built on top.

fasterthanlime:
Merged in #527

PR #479, opened 2025-04-29
Fix clone_into functions

### LINKED ISSUE #435: `clone_into` implementations are wrong
See https://github.com/facet-rs/facet/pull/434, they all do memcopies which is incorrect for `!Copy` types.

fasterthanlime:
Closes #435


fasterthanlime:
Wow that was easier that expected.

PR #481, opened 2025-04-29
pretty: support camino

ppentchev:
Hi,

First of all, thanks again for your work on Facet!

So this is my first attempt at a contribution: make facet-pretty display Utf8PathBuf and Utf8Path values using their implementation of the Display trait.

Please bear in mind that I have no idea what I'm doing, and basing those types' vtables on String and `&str` may very well have broken something or killed the performance of something else that I have no idea about :) So feel free to point out anything that might go wrong or be pessimized as a result of those changes; it's fine if you don't accept them at all.

Thanks again, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/facet-core/src/impls_camino.rs
+++ b/facet-core/src/impls_camino.rs
@@ -57,7 +57,8 @@ unsafe impl Facet<'_> for Utf8PathBuf {
             ))
             .vtable(
                 &const {
-                    let mut vtable = value_vtable!((), |f, _opts| write!(f, "Utf8PathBuf"));
+                    let mut vtable =
+                        value_vtable!(Utf8PathBuf, |f, _opts| write!(f, "Utf8PathBuf"));
                     vtable.parse =
                         Some(|s, target| Ok(unsafe { target.put(Utf8Path::new(s).to_owned()) }));
                     vtable.try_from = Some(try_from);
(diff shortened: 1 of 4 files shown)


facet-core/src/impls_camino.rs
```diff
@@ -57,7 +57,7 @@ unsafe impl Facet<'_> for Utf8PathBuf {
             ))
             .vtable(
                 &const {
-                    let mut vtable = value_vtable!((), |f, _opts| write!(f, "Utf8PathBuf"));
+                    let mut vtable = value_vtable!(String, |f, _opts| write!(f, "Utf8PathBuf"));
```
fasterthanlime, line 60: Any reason this can't just be `Utf8PathBuf`? I think this is what it should've been from the beginning.

(Same remark for `Utf8PathBuf` below)

ppentchev: Good question! As I mentioned, I have no idea what I'm doing, so I tried the first thing that came to mind, and then it didn't even cross my mind that I could use the actual type (I thought this was some kind of "base this on that type that we already know about"). Thanks for setting me straight :) Fixed in https://github.com/facet-rs/facet/pull/481/commits/3f08c50c877ac38c2dd1ab43fb357856f3e4dcb4

fasterthanlime:
@ppentchev Wonderful, thanks for the contribution & the quick follow-up :)

PR #485, opened 2025-04-30
refactor(toml): use facet-serialize

tversteeg:
Really liking this implementation so far! Do you have any suggestions about how to handle custom attributes in the future?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -558,6 +558,7 @@ dependencies = [
  "facet",
  "facet-core",
  "facet-reflect",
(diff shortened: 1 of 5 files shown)


fasterthanlime:
> Do you have any suggestions about how to handle custom attributes in the future?

Do you have specific attributes in mind? I think in the future the serialize API will pass. I don't know. I think most attributes should be at the facet level and then for scoped attributes, I guess they should be passed by facet-serialize to the Serialize impl? I'm not sure yet.

Issue #451, opened 2025-04-26
[facet-json] Deserialise empty tuple


lmmx:
I contributed a PR for deserialisation of tuples yesterday but couldn't get the empty tuple (unit type) to be handled correctly.

It's covered by tests which are ignored (see err and read::tuple submodules of facet-json::tests::integration)

Also N.B. comments from Amos on the PR here {link}

PR #489, opened 2025-04-30
feat(empty-tuple): parse unit type as empty tuple

lmmx:
- This fulfils #451 following preparation in #398
- The tests that were ignored are now active, and supplemented with further tests for a range of nesting levels
  - Please note tests are labelled with a readable name, plus a shorthand for legibility where "u" means "unit" and "i" means integer, so `uGP:i` means "expected unit grandparent but got [string to deserialise] with an integer". This was useful when writing the tests and hopefully when interpreting them, it can be removed if it adds noise.

These tests have brought to light what may be considered incorrect behaviour, or may make sense (to the Rust parser): the 0-tuple "grandparent" `(((),),)` is constructed from the 0-tuple "parent" JSON string "[[]]"

```rust
// (uGGP:uP,uGP:uP,uP:uP) Not unit (great grandparent, grandparent, parent) --> unit parent
// (i.e. there is an implicit default, the value is filled without being present)
// NOTE: Silently coerces to the type
/// Expect a 1-tuple[1-tuple[0-tuple]] but it's a 1-tuple[0-tuple]
#[test]
fn test_deserialize_tuple_empty_nested_flexible() -> Result<()> {
    // Expect a 1x-nested 0-tuple - yup it's a 1x-nested 0-tuple
    let _ok: ((),) = from_str("[[]]")?; // Correct
    // Expect a 2x-nested 0-tuple - coerced from a 1x-nested 0-tuple
    let _ok: (((),),) = from_str("[[]]")?; // Should be from "[[[]]]" but OK
    // Expect a 3x-nested 0-tuple - coerced from a 1x-nested 0-tuple
    let _ok: ((((),),),) = from_str("[[]]")?; // Should be from "[[[[]]]]" but OK

    Ok(())
}

// (uGGP:uP) Not unit great grandparent --> unit parent
// (as for uGP:uP case)
// NOTE: Silently coerces to the type
#[test]
fn test_deserialize_tuple_empty_nested_2x_flexible() -> Result<()> {
    // Expect a 2x-nested 0-tuple - yup it's a 2x-nested 0-tuple
    let _ok: (((),),) = from_str("[[[]]]")?; // Correct
    // Expect a 3x-nested 0-tuple - coerced from a 2x-nested 0-tuple
    let _ok: ((((),),),) = from_str("[[[]]]")?; // Should be from "[[[[]]]]" but OK
    // Expect a 4x-nested 0-tuple - coerced from a 2x-nested 0-tuple
    let _ok: (((((),),),),) = from_str("[[[]]]")?; // Should be from "[[[[[]]]]]" but OK

    Ok(())
}
```

I initially considered this a bug, but upon realising it is not a fixed mistake but rather is modifiable by the type annotation then I consider it perhaps to be a built in flexibility of the Rust parser itself so have moved it from `err.rs` to the `read/tuples.rs` test (and accordingly no snapshot, there is no error to record).

**Status**: I would say this is feature complete but for the localisation of the errors:

Screenshot from 2025-04-30 17-21-19

- Note both the token/position and the error message are incorrect in some cases

As per previous PRs, I'm happy to forego that for future contrib


fasterthanlime:
Thanks for these.

I'm mentally preparing for `facet-deserialize` and I don't know A) how feasible it is to share code between facet-json and others, B) what it's going to look like.

Have you been giving it some thought too? I'm wondering if we could get that behavior right "once and for all" in the common code and let the format-specific crates deal with the idiosyncracies of individual.. well, formats.

lmmx:
Only briefly, I did consider whether there was rework here among the formats earlier but I'm only glancingly familiar. The dispatch approach introduced by ThouCheese in 6e1d479 that replaced all the macros seems like a step towards that state machine direction though.

I think all the formats would want to know the path and the position, so error diagnostics could be abstracted, it occurred to me that for debugging purposes it'd be useful to "see" what was going onto that stack (oh: I see now the facet-kdl crate is pretty TODO!)

Hmm, will see if there's anything I can contribute towards facet-toml to get more of an intuition for the common factors

- The error kinds look like they could be made common
- Maybe a common notion of a 'field' and parts like its opener/key/preamble/separator/closer would be common concepts

If we did that we might then find a lot of code got removed saying things like "the error occurred at this path" as it'd just be implied that the state machine would have been at that part when it bailed (even if we're not calling it `bail!` any more)

I don't see a stack being used in the TOML deserialisation though, I don't know if that's substantively different... Maybe a format could be a trait

fasterthanlime:
> I don't see a stack being used in the TOML deserialisation though

Yeah, the timeline was kind of:

  * I rewrote the facet-json deserializer to be iterative
  * @ThouCheese rewrote the facet-json serializer to be iterative
  * I made the iterative facet-json serializer "generic" so other crates could use it
  * I ported the facet-msgpack serializer to it
  * @tversteeg ported the facet-toml serializer to it

The next steps would be porting all the existing format crates to it, adding what's missing in the process, and then coming up with a `facet-deserializer` approach based on the current facet-json deserializer, then porting all the format crates to that.

the idea is that there would be a lot of effort duplication Between different format crates since there's a lot of shared work, I would imagine, and common problems. Also, doing all that work now makes it less painful to rework the foundations of facet-core, like is being done in #462 

As for facet-kdl, I merged it before it was ready (It's currently publish = false) — There was a bunch of prerequisite work regarding syn and the KDL crate itself that @TheLostLambda tackled first.

Does that make sense?

fasterthanlime:
(@lmmx also please note the miri tests are failing right now so I can't merge this)

TheLostLambda:
Very serendipitous timing here — I was planning on working some more on things tonight and wanted to re-write things to be more iterative / stack driven (as opposed to the `facet-toml` nested functions I was copying before).

@fasterthanlime What is the status of `facet-json`s deserializing at the moment? Is that an architecture you're happy with? If so, I can try to copy that for `facet-kdl`, then, with two very different formats taking the same approach, it might become clearer what code can be abstracted away and shared via a future `facet-deserialize`?

Just to be the most helpful I can be (since I'm still a bit tight on time), would you rather I work on deserialization, or should I try to get serialization working with `facet-serializer`?

lmmx:
Oh yes sorry, got distracted reading about the Strategy pattern and doodling a general solution :nerd_face: Will fix CI now

Was wondering if you might break out traits for the error handling (i.e. split up the "format trait") along the lines of

```rust
/// Convert inner error into the format-native one without exposing Span/Path plumbing to the call-site.
pub trait ErrorPolicy {
    type Error;
    fn reflect(&self, e: ReflectError, path: &str, span: Span) -> Self::Error;
    fn custom(&self, kind: impl Into<Self::Error>) -> Self::Error;
}

/// Knows how to look up whichever field/variant name in the format-specific representation
/// (token stream, toml_edit::Item, yaml_rust2::Event, ...).
pub trait Navigator {
    type Cursor;                 // JSON: &mut Tokenizer,  TOML: &'doc Item
    fn enter_field(&self, cur: &mut Self::Cursor, name: &str) -> bool;
    fn skip_value(&self, cur: &mut Self::Cursor) -> Result<(), ()>;
    // (+any more shared operations)
}
```

held by the deserialisation context:

```rust
pub struct Cx<'facet,'input,E,N> {
    wip:   Wip<'facet>,
    input: &'input [u8],
    span:  Span,
    path:  String,
    err:   E,   // impl ErrorPolicy
    nav:   N,   // impl Navigator
}
```

Could then see how things like recovering from an error and aggregating multiple errors might use those traits: if the ErrorPolicy allows recovery, then use the Navigator's skip_value to move past the bad value and keep going

TheLostLambda:
Far from an expert here, but https://github.com/zesterer/chumsky (same author as [ariadne](https://github.com/zesterer/ariadne)) has thought a lot about error recovery strategies and collecting multiple errors during one pass / parse!

fasterthanlime:
> @fasterthanlime What is the status of `facet-json`s deserializing at the moment? Is that an architecture you're happy with? If so, I can try to copy that for `facet-kdl`, then, with two very different formats taking the same approach, it might become clearer what code can be abstracted away and shared via a future `facet-deserialize`?

@TheLostLambda That's an option. You could also just directly go for facet-deserialize and try to apply it to KDL — But you mentioned having limited time, so maybe focusing on facet-serialize and applying it to KDL since it already exists, might make more sense!

I don't think there's anything useless you can do at that point, if that's any comfort

TheLostLambda:
>  maybe focusing on facet-serialize and applying it to KDL since it already exists, might make more sense!

Nice! I'll start with this then! It should generally help me get my bearings and nail down the attribute syntax too :)

PR #492, opened 2025-05-01
test(facet-msgpack): test cases/snapshots

lmmx:
Here are tests, snapshots, and 'inactive' tests (in a 'drafts' subfolder) for facet-msgpack, taken from the facet-json crate

- For reference if it helps, there's a Python `msgpack` library you can produce these hex bytes with:

```py
import msgpack
import binascii

def print_bytes(data):
    # Pack the data into MessagePack format
    packed = msgpack.packb(data)
    hex_representation = binascii.hexlify(packed).decode('utf-8')
    print(f"Original data: {data}")
    
    # Print hex representation with spaces between bytes
    formatted_hex = ' '.join([hex_representation[i:i+2] for i in range(0, len(hex_representation), 2)])
    print(f"Hex: {formatted_hex}")
    
    # Print raw byte array
    print(f"Bytes: {[hex(b) for b in packed]}")
    
    # Print decimal values of bytes
    print(f"Decimal: {list(packed)}")
    print("-" * 50)
```

e.g.

```py
>>> print_bytes({"Y": ["hello", True]})
Original data: {'Y': ['hello', True]}
Hex: 81 a1 59 92 a5 68 65 6c 6c 6f c3
Bytes: ['0x81', '0xa1', '0x59', '0x92', '0xa5', '0x68', '0x65', '0x6c', '0x6c', '0x6f', '0xc3']
Decimal: [129, 161, 89, 146, 165, 104, 101, 108, 108, 111, 195]
--------------------------------------------------
```


fasterthanlime:
> * For reference if it helps, there's a Python `msgpack` library you can produce these hex bytes with:

You know what? With UV and things like that, I think I'd rather depend on Python for correctness tests than equivalent Rust packages. Feel free to bring in Python into our CI later on if you want to.

Issue #382, opened 2025-04-24
Wishlist: support subcommands in facet-args


ppentchev:
Hi,

First of all, thanks A LOT for writing and maintaining Facet! It is, should I say, extremely refreshing in its simplicity and straightforwardness!

Now, you probably guessed it, but it would be great if facet-args could somehow parse enums as subcommands, ideally also with the next (or rather, the rest) of the command-line arguments being parsed as parameters for the specific enum variant. Something like this, maybe:

``` rust
/// What to do?
#[derive(Debug, Facet)]
#[repr(u8)]
enum CliCommand {
    /// Display program feature information and exit.
    // #[facet(name("features"))] or determined automagically from the stringified variant name
    Features,
}

/// Program name or something
#[derive(Debug, Facet)]
struct Cli {
    /// What to do?
    #[facet(positional)]
    cmd: CliCommand,
}
```

Of course, this is totally a wishlist item... and maybe one day I myself may get 'round to figuring out how `Wip` works so I can try to make it recurse into the enum in that specific way...

Thanks again for all you're doing for Facet and also everything else Rust, and keep up the great work!

G'luck,
Peter

fasterthanlime:
That is absolutely in-scope, thanks for filing it :) And feel free to experiment with it yourself! I will need it for myself at some point but have to prioritize what I tackle first right now :)

fasterthanlime:
Woop this was done a while ago! See the showcase:  https://facet.rs/guide/showcases/args/#subcommands

PR #502, opened 2025-05-03
feat(args): Basic enum support

sinon:
Related: https://github.com/facet-rs/facet/issues/382

### DIFF
--- a/facet-args/src/lib.rs
+++ b/facet-args/src/lib.rs
@@ -10,7 +10,7 @@ use alloc::borrow::Cow;
 mod error;
 
 use error::{ArgsError, ArgsErrorKind};
-use facet_core::{Def, Facet, FieldAttribute};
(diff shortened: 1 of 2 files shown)


sinon:
> Again, I haven't really reviewed this because it's still in draft mode but.. maybe you didn't mean it to be? CleanShot 2025-05-05 at 12 39 03@2x

Sorry for the slow response, have been out of commission with flu. I had intended on using this PR to also add handling for struct-like and tuple-like enum variants but that has proved more involved with needing to re-work the parsing so that figuring out positional considers the nested structure these enums introduce. Will hopefully have a bit of time to focus on this on the weekend but happy to get first item of handling the unit-like case merged.

fasterthanlime:
No rush! I'm happy to keep this open until you feel better :)

PR #506, opened 2025-05-04
feat(toml)!: implement most of serialization

tversteeg:
I was struggling for quite a while with this one, the problem I ended up having was that there was not enough information when some part was finished, such as a field. Which required me to build a complicated stack to keep track of how many items have been processed. That's why I've extended facet-serialize with optional operations.

The method previous method of using toml_write directly was also impossible to use because the order of fields in TOML determines the resulting table structure.

Also once again sorry for the big PR, I couldn't find a meaningful way to split this. I thought about splitting just the serialize part, but there weren't many tests that applied to my use case, so that would be a lot of extra work.

I still need to do proper error handling in the future.


fasterthanlime (approved):
Yep, that's what I wanted from you! Extending facet-serialize so that we don't have to keep extensive state on the side of the serializer is exactly what I had in mind. Thanks for that.

PR #510, opened 2025-05-05
chore(deps): cleanup unused dependecies with cargo-machete

tversteeg:
Maybe we can somehow embed cargo-machete in the pipeline? It's a very useful tool.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -240,31 +240,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam-deque"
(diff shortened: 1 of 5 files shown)


fasterthanlime:
I agree about doing something like `-Wunused-crate-dependencies` in CI (more precise than cargo-machete).

I think the KDL deps were there "because we're going to need them", so I have no strong opinions on whether they stay or go.

Good catch on rayon though, I think that may have been used in `dev` at some point.

PR #511, opened 2025-05-05
test(toml): add wide benchmark

tversteeg:
The nested benchmark doesn't work because the toml crate itself can't handle it.

Results running on my PC:

```
     Running benches/compare_serde.rs (target/release/deps/compare_serde-2dac0ce75db18f24)
Timer precision: 20 ns
compare_serde                fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ Wide - facet_deserialize  67.88 µs      │ 150.6 µs      │ 89.17 µs      │ 88.16 µs      │ 100     │ 100
├─ Wide - facet_serialize    33.48 µs      │ 85.46 µs      │ 34.12 µs      │ 35 µs         │ 100     │ 100
├─ Wide - serde_deserialize  33.78 µs      │ 86.08 µs      │ 34.07 µs      │ 35.03 µs      │ 100     │ 100
╰─ Wide - serde_serialize    23.79 µs      │ 36.13 µs      │ 24.18 µs      │ 24.74 µs      │ 100     │ 100
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -534,6 +534,7 @@ name = "facet-toml"
 version = "0.21.0"
 dependencies = [
  "ariadne",
+ "codspeed-divan-compat",
  "eyre",
  "facet",
  "facet-core",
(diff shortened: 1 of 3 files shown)


fasterthanlime:
Oh that's going to be interesting because serde always has to go through a "DOM" and we might not have to, eventually...

fasterthanlime (approved):

PR #513, opened 2025-05-05
fix(toml): cleanup implementation & support unit type

tversteeg:


fasterthanlime (approved):
Have you been keeping an eye on the `facet-deserialize` efforts? I wonder how much duplicated work there is going on between format-specific crates and the generic thing.

I'll ping you again when my branch passes all the JSON tests.

PR #514, opened 2025-05-05
fix(reflect): add missing scalar types

tversteeg:
Nothing fancy, I just thought this would read better.


fasterthanlime:
I have not looked at the details, but I trust you. I will check out the while thing later before tagging a new release :)

PR #522, opened 2025-05-06
KDL serialization experiments

TheLostLambda:
Opening up a draft PR to ask a couple of design questions before making some changes to `facet-serialize`!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -796,7 +796,7 @@ dependencies = [
 [[package]]
 name = "kdl"
 version = "6.3.3"
-source = "git+https://github.com/TheLostLambda/kdl-rs.git?branch=free-of-syn#963eb2571a734a800e88ea0548c2da738893e7a6"
+source = "git+https://github.com/kdl-org/kdl-rs.git#439aa63bfc595432e6dc6b0fdd708c364628ab18"
 dependencies = [
  "miette",
  "num",
--- a/facet-kdl/Cargo.toml
+++ b/facet-kdl/Cargo.toml
@@ -20,7 +20,7 @@ default = ["std"]
 log = "0.4.27"
 facet-core = { version = "0.20.0", path = "../facet-core", default-features = false }
 facet-reflect = { version = "0.18.2", path = "../facet-reflect", default-features = false }
-kdl = { git = "https://github.com/TheLostLambda/kdl-rs.git", branch = "free-of-syn" }
+kdl = { git = "https://github.com/kdl-org/kdl-rs.git" }
 
 [dev-dependencies]
 facet = { path = "../facet" }


TheLostLambda:
A more actionable question for @fasterthanlime , would you like methods like 
```rs
fn serialize_field_name(&mut self, name: &'static str) -> Result<(), Self::Error>;
```
to be ***changed*** to methods that always take a slice of the arbitrary attributes:
```rs
fn serialize_field_name(&mut self, name: &'static str, attributes: &'static [&'static str]) -> Result<(), Self::Error>;
```
or should I ***add*** new method and keep the old one like this:
```rs
fn serialize_field_name(&mut self, name: &'static str) -> Result<(), Self::Error>;
fn serialize_field_name_with_attributes(&mut self, name: &'static str, attributes: &'static [&'static str]) -> Result<(), Self::Error>;
```

Having default implementations like this could mean that you just need to implement one of the two methods, but it would also make it possible to implement neither of them and end up with code that crashes (presumably) from a stack overflow:
```rs
fn serialize_field_name(&mut self, name: &'static str) -> Result<(), Self::Error> {
    serialize_field_name_with_attributes(self, name, &[])
}

fn serialize_field_name_with_attributes(&mut self, name: &'static str, attributes: &'static [&'static str]) -> Result<(), Self::Error> {
    serialize_field_name(self, name)
}
```

I'm leaning towards just one version of the function that always takes the attributes, since `start_object` already takes an `len: Option<usize>` (setting a precedent for always providing that potentially unneeded information) and the two-version approach would be annoying without the default implementations, but a dangerous foot-gun with them.

TheLostLambda:
Can confirm, the two-default-impls-approach leads to a stack overflow — it looks like the last time a potential solution to this was even discussed was pre-1.0: https://github.com/rust-lang/rfcs/issues/628

fasterthanlime:
> to be _**changed**_ to methods that always take a slice of the arbitrary attributes:

We can change it but I'd rather not use arbitrary attributes for core/blessed crates like facet-kdl — I consider attributes like `attr` / `child` official/blessed and we should use Flags (cheaper to look up when serializing/deserializing) for them!

TheLostLambda:
Ah, I see! I missed those! So just to ensure we're on the same page, you'd like me to add to these?

https://docs.rs/facet/latest/facet/struct.FieldFlags.html

And then change the method to be something like:
```rs
fn serialize_field_name(&mut self, name: &'static str, flags: &FieldFlags) -> Result<(), Self::Error>;
```

I'll need a couple more than are there currently, I think, so I'd be editing `FieldFlags` a bit!

fasterthanlime:
> Ah, I see! I missed those! So just to ensure we're on the same page, you'd like me to add to these?
> 
> https://docs.rs/facet/latest/facet/struct.FieldFlags.html
> 
> And then change the method to be something like:
> 
> ```rust
> fn serialize_field_name(&mut self, name: &'static str, flags: &FieldFlags) -> Result<(), Self::Error>;
> ```
> 
> I'll need a couple more than are there currently, I think, so I'd be editing `FieldFlags` a bit!

Yes, that sounds good!

TheLostLambda:
You're welcome to! I've been busier than intended, unfortunately... I might have the chance to give it a go this weekend, but I can just reopen the PR if so!

PR #525, opened 2025-05-08
Add `PeekListLike` abstraction and fix serialize and `get_item_ptr` for arrays

Akida31:
Lists, arrays and slices are very similar, so this abstracts over `Peek`s of them.

I wasn't sure if I should just modify `facet_reflect/src/peek/list.rs`
since lists may have additional behaviour in the future different from
slices and arrays.

This additionally fixes `get_item_ptr` vtable entry for arrays, since it didn't use
`index` (and just returned the pointer to the original array).


fasterthanlime:
Looking good, thanks for your contribution!

PR #526, opened 2025-05-08
refactor(deserialize)!: make deserialize format stateful

tversteeg:
I think it will be quite hard to implement most deserializers without any state between the passes.


fasterthanlime:
> I think it will be quite hard to implement most deserializers without any state between the passes.

Mhhh yeah I went back and forth on this but the idea was that facet-deserialize should be the one to maintain state — have you already ran into stoppers in facet-toml re: this?

PR #529, opened 2025-05-09
feat: support core::ops::Range

not-my-profile:
I haven't used Facet much so please review this carefully, I'm likely to have missed something. I'd like to use Facet to build a debugger for the parser I'm building and it's using `core::ops::Range<usize>` so I'd like `Range` to implement `Facet` so that I can derive `Facet` for my struct that's containing the range.

My parser is also using  `core::iter::Peekable` so I'd also like to implement Facet for it but I'm not sure what do about its private fields? Omit them?

### DIFF
--- a/facet-core/src/impls_core/ops.rs
+++ b/facet-core/src/impls_core/ops.rs
@@ -1,0 +1,64 @@
+use crate::{
+    ConstTypeId, Def, Facet, Field, Shape, StructDef, StructKind, VTableView, ValueVTable,
+};
+use core::{alloc::Layout, mem};
+
+unsafe impl<'a, Idx: Facet<'a>> Facet<'a> for core::ops::Range<Idx> {
+    const SHAPE: &'static crate::Shape = &const {
+        Shape::builder_for_sized::<Self>()
+            .type_params(&[crate::TypeParam {
+                name: "Idx",
+                shape: || Idx::SHAPE,
+            }])
+            .id(ConstTypeId::of::<Self>())
+            .layout(Layout::new::<Self>())
+            .vtable(
+                &const {
+                    let mut builder = ValueVTable::builder::<Self>().type_name(|f, opts| {
+                        write!(f, "Range")?;
+                        if let Some(opts) = opts.for_children() {
+                            write!(f, "<")?;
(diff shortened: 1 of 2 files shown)


facet-core/src/impls_core/ops.rs
```diff
@@ -0,0 +1,64 @@
...
+                    });
+
+                    if Idx::SHAPE.is_debug() {
+                        builder = builder.debug(|this, f| {
+                            (<VTableView<Idx>>::of().debug().unwrap())(&this.start, f)?;
+                            write!(f, "..")?;
+                            (<VTableView<Idx>>::of().debug().unwrap())(&this.end, f)?;
+                            Ok(())
+                        });
+                    }
```
not-my-profile, line 36: I'm curious what's actually the purpose of reimplementing Debug when facet-pretty isn't even using these impls? Also how would one call this debug impl?

fasterthanlime:
Let's hold off on merging this, it looks good but I would like to finish the type PR first and I don't want to rebase it yet again.

not-my-profile:
Thanks for merging it! btw I had some questions in the PR description and a review comment:

> My parser is also using core::iter::Peekable so I'd also like to implement Facet for it but I'm not sure what do about its private fields? Omit them?

> I'm curious what's actually the purpose of reimplementing Debug when facet-pretty isn't even using these impls? Also how would one call this debug impl?

fasterthanlime:
> I'm curious what's actually the purpose of reimplementing Debug when facet-pretty isn't even using these impls? Also how would one call this debug impl?

I agree with that, maybe we could lose the Debug impl tbh, but it felt natural to have it next to Display etc.

You could call it with unsafe code by directly invoking the function pointer from the vtable, I'm not sure it's exposed safely via facet-reflect right now.

> My parser is also using core::iter::Peekable so I'd also like to implement Facet for it but I'm not sure what do about its private fields? Omit them?

I think that field should probably just be skipped — what are you using facet for mostly? the facet-pretty impl?

fasterthanlime:
> I want to try building some graphical debugger for my parser (or rather any other simple parser).

right right sorry I read that too. would you actually need to see the field that has the `Peekable`?

not-my-profile:
No not at all. I was just wondering. What the common practice is regarding private fields.

PR #531, opened 2025-05-09
refactor(toml)!: deserialize port WIP

tversteeg:
Very rough at the moment, doesn't even properly parse the basic example. But I need to sleep now, so this is to let you know where I'm standing.

### DIFF
--- a/facet-toml/src/deserialize/mod.rs
+++ b/facet-toml/src/deserialize/mod.rs
@@ -55,20 +55,6 @@ impl Toml {
         })
     }
 
-    /// Pop the last item and remove it from the document.
-    fn pop_and_remove(&mut self) {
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Maybe it's just too early for us to migrate TOML to facet-deserialize? We want the base to be a TOML event based parser, not something like the current document based crate we have, as far as I can tell.

I think that's the mismatch here. Since the toml crate already deserializes to a Document object, you are then struggling to emit corresponding parsing events. What we should be doing instead is making or looking for, a TOML parser, which I believe @epage is looking into.

I'm okay with imperfect solutions at this stage, and we can decide it's just too early to migrate facet-toml to facet-deserialize.

PR #546, opened 2025-05-10
Fix #338: Add canada.json benchmark for facet-json

### LINKED ISSUE #338: json: Parse canada.json
It's a classic!

https://github.com/mloskot/json_benchmark/blob/master/data/canada.json

fasterthanlime:
- Add canada.json as XZ-compressed file for smaller repo size
- Set up benchmark comparing facet-json and serde_json performance
- Fix String serialization in facet-json by enabling std feature for facet-reflect
- Add test to validate String serialization works correctly


fasterthanlime:
On my machine:

CleanShot 2025-05-10 at 11 45 24@2x

fivemservereyestore:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #560, opened 2025-05-10
feat(facet-args): default field value/function

### LINKED ISSUE #443: facet-args: using named argument short-code causes panic

### LINKED ISSUE #520: `facet-pretty` respects `NO_COLOR=1`
`NO_COLOR` is a universal way to suppress terminal color codes.

Currently `facet-pretty` still prints color codes.

references:
- https://www.gnu.org/software/gettext/libtextstyle/manual/html_node/The-NO_005fCOLOR-variable.html
- https://no-color.org/

### LINKED ISSUE #521: Question: how to implment a transparent foreign type (E.g. Regex)?
What I want is a ScalarLike type, which is not allowed to access its internal, but it can be deserialized with `FromStr::from_str` and can be serialized with `Display::fmt`

```rust
#[derive(Facet)]
struct Foo {
  re: Regex,
}
```

I found that there is a `Opaque` struct in the crate, but it's inner field `.0` is not accessable.

lmmx:
- feat(facet-args): default and default value, tests


fasterthanlime:
I still believe this should live in `Wip`, not in `facet-deserialize` or `facet-args`.

And if it truly should live in facet-deserialize.. couldn't `facet-args` be implemented on top of `facet-deserialize`?

fasterthanlime:
(However in the meantime, let's merge this I guess, to get the functionality in, with tests, then when we change the approach we can look out for regressions).

lmmx:
Indeed, yeah good call, I guess the tests will be able to pin the correct behaviour down

I didn't think there's any common ground with the positional/named params so perhaps a smaller implementation just for this might be fine, but no you're right less maintenance surface area and duplication of effort is good.

I'd think the behaviour where a bool should be default false if absent would not be something that's handled differently to a regular struct

lmmx:
I'm just going to roll back that conflict resolution merge and force re-push

fasterthanlime:
> I'm just going to roll back that conflict resolution merge and force re-push

No worries, I've got a clean rebase I'm merging now :)

fasterthanlime:
Merged manually.

lmmx:
Cheers Amos ! :saluting_face:

lmmx:
FYI the commit took out the recent gitignore addition

```
**/.claude/settings.local.json
**/CLAUDE.local.md
```

fasterthanlime:
> FYI the trunk commit has removed the recent gitignore addition

Nope, just blank lines!

CleanShot 2025-05-10 at 21 17 16@2x

PR #572, opened 2025-05-11
Add `Facet` impl for `Box`

kylewlacy:
Pretty self-explanatory: this PR adds a new impl of `Facet` for `Box<T>`

There was already a [`KnownSmartPointer::Box`](https://docs.rs/facet/0.25.1/facet/enum.KnownSmartPointer.html#variant.Box) variant, so this change ended up being pretty minimal. Full disclosure: I'm still not fully up to speed on all the inner workings of facet and I started from the `Arc<_>` impl as a basis, so I'm _hoping_ everything here holds up to scrutiny. Miri isn't complaining so I take that as a good sign

---

facet's been on my radar for a while, and I finally have a use-case where I _think_ it'll help solve a problem that I can't solve (easily) with Serde today. While working through the changes, I was surprised to see that `Box<_>` didn't have a `Facet` impl... I saw this as a good chance to get my hands a bit dirty and get a feel for how facet ticks!

### DIFF
--- a/facet-core/src/impls_alloc/boxed.rs
+++ b/facet-core/src/impls_alloc/boxed.rs
@@ -1,0 +1,166 @@
+use crate::{
+    Def, Facet, KnownSmartPointer, PtrConst, PtrMut, PtrUninit, Shape, SmartPointerDef,
+    SmartPointerFlags, SmartPointerVTable, TryBorrowInnerError, TryFromError, TryIntoInnerError,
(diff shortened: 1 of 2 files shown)


fasterthanlime (approved):

fasterthanlime:
Thanks for this! I couldn't see anything suspicious at first glance, if miri couldn't either, that is indeed a good sign.

PR #573, opened 2025-05-11
Add support for the time crate

fasterthanlime:
While doing this it reminded me that I think `ScalarType` is kinda duplicated with ScalarAffinity.

And also that I think DateTime affinities should have vtables.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -259,6 +259,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "deranged"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9c9e6a11ca8224451684bc0d7d5a7adbf8f2fd6887261a1cfc3c0432f9d4068e"
+dependencies = [
+ "powerfmt",
+]
+
+[[package]]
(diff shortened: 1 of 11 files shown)


fasterthanlime:
> What do you think about the other common time libraries, `chrono` and `jiff`? Do you want me to implement those similarly to this one?

That'd be nice! Eventually I think we should have vtables for the different datetime types but.. for now at least we get JSON support (and maybe TOML). It's not very flexible right now because it's only RFC3339, but there are worse defaults right.

BurntSushi:
> It's not very flexible right now because it's only RFC3339

Have you considered support RFC 9557? That would be ideal since it lets you encode the time zone identifier in band and avoids DST arithmetic footguns if you're using a sufficiently sophisticated datetime library. :-)

fasterthanlime:
> Have you considered support RFC 9557? That would be ideal since it lets you encode the time zone identifier in band and avoids DST arithmetic footguns if you're using a sufficiently sophisticated datetime library. :-)

This might be an even better default, ideally facet itself should not pick just the one option though, so, soon enough we'll need some way to specify which standard to use.

PR #565, opened 2025-05-10
refactor(facet-args): rely on facet-deserialize

lmmx:
- Refactor of facet-args to rely on facet-deserialize, with structured default field parsing (i.e. done on a StackRunner)
  - which was then extracted into a dedicated module `defaults.rs` (with tests)
  - default handling done in dedicated functions moved over to a proper module `parse.rs` (with tests)

- I am still not using a Format trait (I drafted it but it didn't take and I walked it back), but it _is_ using facet-deserialize where it wasn't before so I'd suggest going with this?

- Note: 2 small `pub` changes in facet-deserialize (`original_input` field and `pop` fn)

- :shipit: Squished into one commit, ship/review at will

- **stretch goal** - while this PR is ready, I wanted to try developing a format trait and sort of got it, but it's a bit involved and still unsure if it will take so split that off to a draft PR #575 (that one is WIP)

PR #575, opened 2025-05-12
feat: facet-args `CliFormat` trait impl; deserialize `&str` field

### LINKED ISSUE #561: [facet-args] Use `facet-deseralize` to create the `Wip` from CLI args
As mentioned in #550 ([via](https://github.com/facet-rs/facet/issues/550#issuecomment-2869046040)), RE: facet-args `from_slice` (which recapitulates what has recently been put into a nice structured approach in the `StackRunner` in facet-deserialize), the creation of `Wip` from the CLI args slice should avoid repeating the code we have there.

- This will also mean that facet-args gets access to all the perf improvements coming to facet-deserialize

### LINKED ISSUE #593: [facet-deserialize] `&str` support
If I'm reading this rightly, if the scalar string found by the facet-deserialize `handle_scalar` function gets a type that is not an enum, it'll always call `to_string()` on it

I found this the hard way after trying many ways to get `&str` out of facet-deserialize, and they all came back as `String`!

https://github.com/facet-rs/facet/blob/a0322be1c0ce65275004d4f9b09073dc871cc7bf/facet-deserialize/src/lib.rs#L635

Is this something we can change (or will it disrupt facet-json)?

**edit** ignore the following, it's not this

~The type does distinguish between a `Type::Primitive(PrimitiveType::Textual(TextualType::Str))` which is a borrowed string vs any other type including `String` (I think?)~

https://github.com/facet-rs/facet/blob/a0322be1c0ce65275004d4f9b09073dc871cc7bf/facet-core/src/types/ty/primitive.rs#L10

lmmx:
This PR was split off from #565 to try to develop a `CliFormat` trait, and did :tada:

- Ready to merge
- Closes #561
- Closes #593

## Status

- [x] All tests passing
- Possible follow-up: make an enum for the `wanted` values not static str
- Removed the dead code for parsing fields (deleted defaults and parse modules entirely)

## Open questions:

- Was it right to use a `Box::leak` to leak borrowed CLI args, or should I change it to use thread-local storage with an `unsafe` block? (Currently this has `#![forbid(unsafe_code)]`

<Details><summary>Click to show more about this design choice</summary>

## Unsafe vs. leaky design choice

(Opinions welcome)

**edit** I edited it  7406d6f  to leak rather than using unsafe-in-a-safe-way thread-local, might but it pack (and downgrade the forbid to deny and allow this part)

If I've got this right, we can either:

- use an unsafe block to borrow any `&str` strings (would require removing the forbid(unsafe) directive). I put this in 671bd43
- leak memory to borrow any `&str` strings (I think this is worse). The diff to add this would be the following:

<details><summary>Click to show diff</summary>

```diff
diff --git a/facet-args/src/format.rs b/facet-args/src/format.rs
index 22d15ab9..672a5295 100644
--- a/facet-args/src/format.rs
+++ b/facet-args/src/format.rs
@@ -383,55 +383,24 @@ impl Format for CliFormat<'_> {
                         }),
                     );
                 }
-
-                // The key fix: create appropriate string representation based on target type
-                // Use static strings for &str fields, which persist for the duration of the program
-                // if shape.is_type::<&str>() {
-                // &str field: hand out a *borrowed* slice of the CLI token
-                if matches!(
-                    shape.ty,
-                    Type::Primitive(PrimitiveType::Textual(TextualType::Str))
-                ) {
-                    return (
-                        nd,
-                        Ok(Spanned {
-                            node: Outcome::Scalar(Scalar::String(Cow::Borrowed(arg))),
-                            span,
-                        }),
-                    );
-                    // // Store the string in the STATIC_STRINGS thread-local set for &str values
-                    // let static_str_ref = STATIC_STRINGS.with(|strings| {
-                    //     let mut strings = strings.borrow_mut();
-                    //     let index = strings.len();
-                    //     strings.push(arg.to_string());
-
-                    //     // Get a static reference to the string in the vector
-                    //     // This is safe because the thread-local storage lasts for the entire program
-                    //     #[allow(unsafe_code)]
-                    //     {
-                    //         unsafe {
-                    //             let ptr = strings[index].as_str() as *const str;
-                    //             &*ptr
-                    //         }
-                    //     }
-                    // });
-                    // (
-                    //     nd,
-                    //     Ok(Spanned {
-                    //         node: Outcome::Scalar(Scalar::String(Cow::Borrowed(static_str_ref))),
-                    //         span,
-                    //     }),
-                    // )
+                // String types
+                // `String`: own the token; otherwise: hand out a borrow.
+                let scalar = if shape.is_type::<String>() {
+                    Scalar::String(Cow::Owned(arg.to_string()))
                 } else {
-                    // For any other type (including String), use Cow::Owned
-                    (
-                        nd,
-                        Ok(Spanned {
-                            node: Outcome::Scalar(Scalar::String(Cow::Owned(arg.to_string()))),
-                            span,
-                        }),
-                    )
-                }
+                    // `&str` (or any other borrowed-string compatible shape)
+                    let leaked: &'static str = Box::leak(arg.to_string().into_boxed_str());
+                    eprintln!("Leak me");
+                    Scalar::String(Cow::Borrowed(leaked))
+                };
+
+                (
+                    nd,
+                    Ok(Spanned {
+                        node: Outcome::Scalar(scalar),
+                        span,
+                    }),
+                )
             }
 
             // List items
diff --git a/facet-args/src/lib.rs b/facet-args/src/lib.rs
index cd3fad38..10211aab 100644
--- a/facet-args/src/lib.rs
+++ b/facet-args/src/lib.rs
@@ -1,7 +1,7 @@
 #![warn(missing_docs)]
 #![warn(clippy::std_instead_of_core)]
 #![warn(clippy::std_instead_of_alloc)]
-// #![forbid(unsafe_code)]
+#![forbid(unsafe_code)]
 #![doc = include_str!("../README.md")]
 
 extern crate alloc;
```

</details>

</details>


lmmx:
@fasterthanlime any thoughts on the `Box::leak` approach? I am just re-reviewing it and still wondering if that was truly the best way.

https://github.com/facet-rs/facet/blob/a0a6a01850cf313899de419b6e3d34173c514e14/facet-args/src/format.rs#L391

Is that feasible in this project / just in facet-args, or neither (and should be replaced with an `unsafe` block using the thread-local RefCell approach)?

https://github.com/facet-rs/facet/blob/a0a6a01850cf313899de419b6e3d34173c514e14/facet-args/src/format.rs#L125-L129

fasterthanlime:
> I am just re-reviewing it and still wondering if that was truly the best way.

Oh my opinion is that we shouldn't leak memory always. Aren't the arguments people are deserializing from `&'static` to start with? Also, if we're deserializing to a Cow, why not allocate and return a`Cow::Owned`

lmmx:
> > I am just re-reviewing it and still wondering if that was truly the best way.
> 
> Oh my opinion is that we shouldn't leak memory always. Aren't the arguments people are deserializing from `&'static` to start with? Also, if we're deserializing to a Cow, why not allocate and return a`Cow::Owned`

Ah, well mainly because this is perhaps my first proper entanglement with `Cow` and I don't know better :grimacing: Thank you for the nudge I'll see about it!

lmmx:
Yeah it looks like I got the lifetimes wrong, it should be more like

```diff
-pub struct CliFormat<'a> {
-    args: &'a [&'a str],
-}
+pub struct CliFormat;
 
-impl fmt::Display for CliFormat<'_> {
+impl fmt::Display for CliFormat {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "CliFormat")
     }
 }
-impl<'a> CliFormat<'a> {
-    /// Create a new CLI format instance
-    pub fn new(args: &'a [&'a str]) -> Self {
-        Self { args }
-    }
-
+impl CliFormat {
     /// Helper function to convert kebab-case to snake_case
     pub(crate) fn kebab_to_snake(input: &str) -> Cow<str> {
         if !input.contains('-') {
@@ -36,104 +29,24 @@ impl<'a> CliFormat<'a> {
 }
 
 /// Parses command line arguments using the Facet deserialization framework
-pub fn from_slice_with_format<'input, 'facet, T>(s: &'facet [&'input str]) -> Result<T, ArgsError>
-where
-    T: Facet<'facet>,
-    'input: 'facet,
+pub fn from_slice<'input: 'facet, 'facet, T: Facet<'facet>>(
+    input: &'input [u8]
+) -> Result<T, DeserError<'input>>
 {
```

I guess this is more of a half way point towards the proper Format trait implementation then.

Not sure whether to pursue this or start over :disappointed:

Issue #471, opened 2025-04-28
Support json-structure


fasterthanlime:
Looks cool? Might be nice to get on the ground floor of this:

https://json-structure.org/

lmmx:
Really cool 😮 Great find, I found this blog helpful to get the idea (a better data definition language than JSON Schema, as opposed to document validation) https://vasters.com/clemens/2025/02/19/a-thought-experiment-new-json-schema

PR #577, opened 2025-05-12
Add new `SetDef` and impls for `HashSet` and `BTreeSet`

### LINKED ISSUE #247: Support `HashSet<T>`
Like the std docs say, it's pretty much `HashMap<T, ()>` — idk if it's worth a separate vtable?

kylewlacy:
Closes #247

This PR adds initial support for sets with a new `Def::Set(SetDef)` variant-- plus a new `SetVTable` to go along with it. It also adds impls of `Facet` for both `HashSet<T, S>` and `BTreeSet<T>`

`SetVTable` has the following methods:

- `init_in_place_with_capacity(PtrUninit, usize) -> PtrMut`
- `insert(PtrMut, T) -> bool`
- `len(PtrConst) -> usize`
- `contains(PtrConst, T) -> bool`
- `iter(PtrConst) -> PtrMut`

From a high level, I was split between adding set as its own def, or if I should try to build on map (or possibly list). Personally, I feel it's right in between map and a list, but different enough from either to warrant its own category. I learned about JSON Structure after spotting #471, and was ultimately swayed because [a set is one of the core types in JSON Structure](https://json-structure.org/json-structure-primer.html#21-differences-to-json-schema-drafts). But, I'm not married to this design and would be happy to iterate on this PR!

### DIFF
--- a/facet-core/src/impls_alloc/btreeset.rs
+++ b/facet-core/src/impls_alloc/btreeset.rs
@@ -1,5 +1,6 @@
 use core::hash::Hash;
 
+use alloc::boxed::Box;
 use alloc::collections::BTreeSet;
 
 use crate::ptr::{PtrConst, PtrMut};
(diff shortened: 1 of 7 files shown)


fasterthanlime (approved):
Looking good, thanks for tackling this!

PR #583, opened 2025-05-12
Fixes around Camino/UUID/ULID

kylewlacy:
I started work on adding support for another crate (`bstr`), and had to do a gut check as I was referencing the impl from `camino`. I noticed that the `try_borrow_inner` method seemed to be returning a value of type `&str`, instead of `String` like it should've.

...anyway, I fell down a bit of a rabbit hole, and ended up with a few changes impacting Camino, UUID, and ULID integrations:

- Remove invalid `try_borrow_inner` impls across Camino/UUID/ULID (I'm not even really sure any of the 3 should have inner shapes at all...? but that's a bigger question)
- Fix `value_vtable!()` invocations for UUID/ULID. I think the prior version was valid, but this change gives us some more traits (namely `Display`, which was the important one)
- Remove `ScalarType` variants that were specific to Camino/UUID/ULID. It feels weird having the scalar types be feature-gated in general, and these particular ones were causing pain around serialization. It feels like `ScalarAffinity` is a cleaner fit
- Remove `camino` / `ulid` / `uuid` features/optional deps from `facet-reflect`. Without the `ScalarType` variants, these features had no effect
- Support serializing Camino/UUID/ULID types as strings. This defers to the `Display` impl based on the shape's affinity, basically piggy-backing off of #573
- Add some JSON (de)serialization tests for Camino/UUID/ULID

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -455,6 +455,7 @@ dependencies = [
 name = "facet-json"
 version = "0.23.4"
 dependencies = [
+ "camino",
  "eyre",
  "facet",
  "facet-core",
@@ -465,6 +466,8 @@ dependencies = [
  "insta",
  "log",
  "time",
+ "ulid",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


fasterthanlime:
Fixed memory leaks & merged by hand.

PR #586, opened 2025-05-12
feat(core): add `jiff` implementations

tversteeg:
There are also a lot more types in `jiff`, I picked the ones that represent date-time, should we also add implementations for the other types?

### DIFF
--- a/facet-core/src/lib.rs
+++ b/facet-core/src/lib.rs
@@ -50,6 +50,9 @@ mod impls_time;
 #[cfg(feature = "url")]
 mod impls_url;
 
+#[cfg(feature = "jiff")]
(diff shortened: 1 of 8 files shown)


facet-core/src/lib.rs
```diff
@@ -47,6 +47,9 @@ mod impls_ulid;
 #[cfg(feature = "time")]
 mod impls_time;
 
+#[cfg(feature = "jiff")]
```
BurntSushi, line 50: I'd strongly urge you to name this `jiff02`. `jiff 1.0` is going to be released this summer.

(Unless y'all are doing frequent breaking change releases and don't mind doing one to bump the Jiff version.)

PR #587, opened 2025-05-12
fix: wrong offset for end field of core::ops::Range

not-my-profile:


facet-core/src/impls_core/ops.rs
```diff
@@ -25,7 +25,7 @@ unsafe impl<'a, Idx: Facet<'a>> Facet<'a> for core::ops::Range<Idx> {
                                 Field::builder()
                                     .name("end")
                                     .shape(Idx::SHAPE)
-                                    .offset(mem::offset_of!(core::ops::Range<Idx>, start))
+                                    .offset(mem::offset_of!(core::ops::Range<Idx>, end))
```
fasterthanlime, line 28: thanks!

fasterthanlime (approved):

PR #588, opened 2025-05-12
chore(args)!: Remove arg field parsing to non-public

sinon:
### Why?

Noticed that there was some new methods on the public API that seemed more internal in nature

@lmmx - please object if you did mean for these to be part of the libraries public API

### What?

Switch to `pub(crate)` move unit tests out of integration and instead into the relevant modules

### Testing

`cargo doc --open`


lmmx:
Oh I did not know `pub(crate)` was a thing! 😅 Yup that sounds exactly what I was looking for, internal use by other modules in the same crate.

lmmx:
I took a look around the other crates at where their tests were before I put these separately from the module, but if I've got the house style wrong then apologies on that.

sinon:
> Oh I did not know pub(crate) was a thing! 😅 Yup that sounds exactly what I was looking for, internal use by other modules in the same crate.

Good article on the topic: https://kobzol.github.io/rust/2025/04/23/two-ways-of-interpreting-visibility-in-rust.html (though it's really about 2 ways to achieve the same thing, not sure if there is a preference in these libs for which is the preferred style 🤔 )

> I took a look around the other crates at where their tests were before I put these separately from the module, but if I've got the house style wrong then apologies on that.

Easy to do as some of the most tested libs like `facet-reflect` / `facet-core` are all libraries for libraries with a pretty big public interface so integration tests as the norm make sense.

Issue #582, opened 2025-05-12
facet-serialize should expose the enum discriminant using (e.g.) start_enum_variant


RoastVeg:
This one cropped up working on an XDR (RFC 4506) serializer, which always (?) needs to include the discriminant for enums. I currently have this modification to `serialize_iterative`:

```Rust
match (peek.shape().def, peek.shape().ty) {
    // snip
    (_, Type::User(ut)) => match ut {
        // snip
        UserType::Enum(_) => {
            let peek = peek.into_enum().unwrap();
            let variant = peek.active_variant().unwrap();
            let variant_index = peek.variant_index().unwrap();
            let discriminant = variant
                .discriminant
                .map(|d| d as u64)
                .unwrap_or(variant_index as u64);
            self.start_enum_variant(discriminant)?;
            for (field, field_peek) in peek.fields_for_serialize() {
                stack.push(SerializeTask::Value(field_peek, Some(field)));
            }
        }
    }
}
```

If this is the right approach, then I'll open a PR

fasterthanlime:
I have no objection to this!

PR #592, opened 2025-05-12
Add facet-xdr

RoastVeg:
I did some proptesting [out of tree](https://github.com/RoastVeg/facet-xdr/blob/79246b311e13923de2294ffbb29a2e7b7f882a21/src/lib.rs#L802) while I was making this, but since proptest is full of syn I figured it wasn't worth porting over.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -621,6 +621,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "facet-xdr"
+version = "0.1.0"
+dependencies = [
+ "facet",
+ "facet-core",
+ "facet-reflect",
+ "facet-testhelpers 0.17.2",
+]
+
+[[package]]
 name = "facet-yaml"
 version = "0.24.5"
 dependencies = [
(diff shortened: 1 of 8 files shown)


fasterthanlime:
> I did some proptesting [out of tree](https://github.com/RoastVeg/facet-xdr/blob/79246b311e13923de2294ffbb29a2e7b7f882a21/src/lib.rs#L802) while I was making this, but since proptest is full of syn I figured it wasn't worth porting over.

Might be worth in the long run!

fasterthanlime (approved):
Let's get this released, we can iterate on it later :)

RoastVeg:
> Is there a specific reason you chose to implement it directly with `facet-reflect` and not use `facet-serialize` and/or `facet-deserialize`?

Wrt `facet-serialize`, I would need to add a fix for #582 which should be trivial.

I have found `facet-deserialize` pretty unsuitable for non-self-describing formats like this. It's probably prudent to wait until there is more maturity there.

Issue #578, opened 2025-05-12
Make lists more flexible


kylewlacy:
Currently (as of facet v0.25.1) [`ListVTable`](https://docs.rs/facet/0.25.1/facet/struct.ListVTable.html) has the following methods:

- `init_in_place_with_capacity` (optional)
- `push`
- `len`
- `as_ptr`
- `as_mut_ptr`

Unlike [`MapVTable`](https://docs.rs/facet/0.25.1/facet/struct.MapVTable.html), there's no `iter` method or similar. Instead, users call `as_ptr` to get the pointer to the first item, `len` to get the number of items, and iterate through with pointer math (e.g. see [`PeekListLike::get`](https://github.com/facet-rs/facet/blob/8e566762ff39df16afb442e11fa70b798055c3d3/facet-reflect/src/peek/list_like.rs#L118)).

This is pretty restrictive! There aren't many collections besides `Vec` that can always be iterated like this. A few examples that couldn't currently be represented with `ListVTable`:

- [`VecDeque`](https://doc.rust-lang.org/stable/std/collections/struct.VecDeque.html) - made up of 1 or 2 contiguous slices (but can be converted to one using `.make_contiguous()`)
- [`LinkedList`](https://doc.rust-lang.org/stable/std/collections/struct.LinkedList.html) - not contiguous at all (but... nearly useless?)
- [`bytes::Byes`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html) - multiple slices

fasterthanlime:
I agree with this! List should not require contiguity (hopefully that's the correct way to say this).

It would be nice for contiguous data structures to have some sort of way to say "yeah it is contiguous you can do a fast iterate without going through the vtable" but 🤷

Ichmed:
Maybe use a Marker trait like `Copy`? I don't know well that fares with how the specialization works in Facet.

`impl<T> Contiguous for Vec<T> {}`

PR #617, opened 2025-05-15
Refactor `MapVTable` and `SetVTable` to share a common `IterVTable`

kylewlacy:
Relates to #578

This PR adds a new `IterVTable`, which covers the functionality of both the prior `MapVTable` and `SetVTable`. The new vtable also has an `init_with_value` function, which also replaces both previous `iter` functions (this was so the iter vtable could have both the `init` and `dealloc` functions for symmetry)

To account for maps returning key/value pairs and sets returning just values, `IterVTable` includes both `next`/`next_back` functions (required, yields a single value) plus `next_pair`/`next_pair_back` functions (optional, yields two values). This feels a little gross, but my thought was that you can view a map like an iterator over its values, so it kinda makes sense?[^iter-keys] (I also have an alternative idea for implementing this but it's also a little gross, so I'll open it as a follow-up PR for discussion)

This change is on the road to #578, where I see the final step as giving `ListVTable` its own iterator. It seemed silly to have another separate iter vtable just for lists, so I wanted to generalize first.

[^iter-keys]: I think it would also be valid to see a map as an iterator over its _keys_ instead of its _values_, I don't have a strong opinion on which is better honestly...

### DIFF
--- a/facet-core/src/impls_alloc/btreemap.rs
+++ b/facet-core/src/impls_alloc/btreemap.rs
@@ -6,7 +6,7 @@ use alloc::{
 };
 
 use crate::{
-    Def, Facet, MapDef, MapIterVTable, MapVTable, MarkerTraits, PtrConst, PtrMut, Shape, Type,
+    Def, Facet, IterVTable, MapDef, MapVTable, MarkerTraits, PtrConst, PtrMut, Shape, Type,
     UserType, VTableView, ValueVTable,
 };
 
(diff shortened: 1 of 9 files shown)


fasterthanlime (approved):
Makes sense to me

fasterthanlime:
> 1. I think it would also be valid to see a map as an iterator over its _keys_ instead of its _values_, I don't have a strong opinion on which is better honestly... 

(I'm honestly not sure either!)

PR #618, opened 2025-05-15
Make `IterVTable` generic over the iteratee

kylewlacy:
Follow-up to #617

This PR builds on #617, with a different way to handle the "map vs. set" iteration problem. The original way was to basically offer two different ways to iterate: `next` which returnes a single `PtrConst`, and `next_pair` which returns a pair of `PtrConst` values (i.e. the key and value of a map)

This PR tweaks the new `IterVTable` type to be generic over what `next` should return-- the iteratee. Sounds easy enough... except `NextFn` (like all vtable methods) uses [Higher-Ranked Trait Bounds](https://doc.rust-lang.org/nomicon/hrtb.html) so it can be generic over lifetimes. To properly type this, we need to "apply" this higher-ranked lifetime to the generic type.

The tool we need for that is a Higher-Kinded Type. Luckily, [GATs](https://blog.rust-lang.org/2022/10/28/gats-stabilization/) are isomorphic to Higher-Kinded Types, but it's a little clunky. That's what the new `ItemIter` trait is-- it lets us fill in a lifetime for either `PtrConst` or `(PtrConst, PtrConst)`, letting us treat it like a Higher-Kinded Type.

### DIFF
--- a/facet-core/src/types/def/iter.rs
+++ b/facet-core/src/types/def/iter.rs
@@ -12,16 +12,8 @@ pub type IterInitWithValueFn = for<'value> unsafe fn(value: PtrConst<'value>) ->
 /// # Safety
 ///
 /// The `iter` parameter must point to aligned, initialized memory of the correct type.
-pub type IterNextFn = for<'iter> unsafe fn(iter: PtrMut<'iter>) -> Option<PtrConst<'iter>>;
-
-/// Advance the iterator, returning the next value pair from the iterator.
-/// For example, this would return the next key/value pair from a map.
-///
-/// # Safety
-///
-/// The `iter` parameter must point to aligned, initialized memory of the correct type.
-pub type IterNextPairFn =
-    for<'iter> unsafe fn(iter: PtrMut<'iter>) -> Option<(PtrConst<'iter>, PtrConst<'iter>)>;
+pub type IterNextFn<T> =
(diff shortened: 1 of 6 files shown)


facet-core/src/types/def/iter.rs
```diff
@@ -0,0 +1,152 @@
...
+///
+/// This trait is needed as a utility, so the functions within [`IterVTable`]
+/// can apply the appropriate lifetime to their result types. In other words,
+/// this trait acts like a higher-kinded type that takes a lifetime.
+pub trait IterItem {
+    /// The output type of the iterator, bound by the lifetime `'a`
+    type Item<'a>;
+}
+
+impl IterItem for PtrConst<'_> {
```
kylewlacy, line 142: Also side note: there's actually no reason this needs to be implemented on `PtrConst<'_>` directly. The convention I've seen before would look like this:

```rust
// Empty enum (uninhabitable, only exists at compile-time)
pub enum PtrConst_ { }

impl IterItem for PtrConst_ {
  type Item<'a> = PtrConst<'a>;
}
```

It feels kinda cursed to impl it on the generic type directly, but it looks more intuitive

fasterthanlime:
The prerequisite has been merged, so this can be taken out of draft, I think?

kylewlacy:
> The prerequisite has been merged, so this can be taken out of draft, I think?

Yep, should be good to review now (still needs to be rebased, which I'll do once I get off work today)

fasterthanlime (approved):
Looks good to me, merging manually!

PR #619, opened 2025-05-15
feat(deserialize): multi-input type support via generic `Format`

lmmx:
- **feat(deserialize): support multiple input types via generic `Format`**

**Summary**: refactor with no change to behaviour, allows us to support `&str` arguments in a Format needed for facet-args
**Motivation**: JSON is a sequence of bytes, but the OS splits up the CLI args for us when making `std::env::args` so we need a way to iterate over those strings, this allows us to do that

This PR keeps the currently hardcoded `Format` input type as `u8` as the default while adding the ability to take `&str`

It demonstrates usage of this in the 2 tests included, both deserializing the same Facet struct with one field `nom: String` (note the change in string type in this PR refers to the deserialisation data _input_ type not the Facet field type)

- `u8` test (which we already know works thanks to facet-json), here implements a trivial mock bytes format that reads 2 bytes `b"xy"` and always produces the same output
  - imaginary scenario where `b"x"` is always interpreted as the field name "nom" and `b"y"` is always code for the field value "test"
- `&str` test (which gives a simple proof of concept for a facet-args rewrite), here implements a trivial mock CLI arguments format
   - more realistic scenario where `"--nom"` is always stripped of "--" prefix and interpreted as the field name ("nom") and the other value "test" is not interfered with

image

### DIFF
--- a/facet-args/src/defaults.rs
+++ b/facet-args/src/defaults.rs
@@ -24,8 +24,8 @@ pub(crate) fn apply_field_defaults(wip: Wip<'_>) -> Result<Wip<'_>, ArgsError> {
 
     // Set up StackRunner for default handling
     let mut runner = StackRunner {
-        original_input: &[],
-        input: &[],
+        original_input: &[] as &[u8],
+        input: &[] as &[u8],
         stack: vec![],
         last_span: Span::new(0, 0),
     };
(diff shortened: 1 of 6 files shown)


lmmx:
Test coverage fell even though I added tests again haha, review/merge at your leisure @fasterthanlime 🫡

PR #620, opened 2025-05-15
feat(core): add `as_string` attribute

### LINKED ISSUE #585: Derive attribute for items implement `FromStr` and `ToString`
While implementing the `jiff` types, I noticed that a lot of `Facet` implementations can be easily created if they implement `core::str::FromStr` and `core::fmt::Display`.

So I was thinking maybe add an attribute to the derive macro to simplify this?

Something like:

```rust
#[derive(Facet)]
#[facet(as_string)]
struct MyType {
  ..
}

impl Display for MyType {
  ..
}

impl FromStr for MyType {
 ..
}
```

Not entirely sure about the `as_string` naming.

tversteeg:
Only got the attribute so far, still need to actually implement it.

Fixes #585.

### DIFF
--- a/facet-core/src/types/mod.rs
+++ b/facet-core/src/types/mod.rs
@@ -119,6 +119,9 @@ pub enum ShapeAttribute {
     /// missing in the input should be filled with corresponding field values from
     /// a `T::default()` (where T is this shape)
     Default,
+    /// Indicates that, this shape should be serialized with its `ToString`
+    /// implementation, and deserialized with its `FromStr` implementation
+    AsString,
     /// Indicates that this is a transparent wrapper type, like `NewType(T)`
     /// it should not be treated like a struct, but like something that can be built
     /// from `T` and converted back to `T`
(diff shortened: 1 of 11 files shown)


facet-core/src/types/mod.rs
```diff
@@ -119,6 +119,9 @@ pub enum ShapeAttribute {
     /// missing in the input should be filled with corresponding field values from
     /// a `T::default()` (where T is this shape)
     Default,
+    /// Indicates that, this shape should be serialized with its `ToString`
+    /// implementation, and deserialized with its `FromStr` implementation
+    AsString,
```
fasterthanlime, line 124: I don't love that this is another `ShapeAttribute` instead of a `ShapeFlag` but I guess that refactor hasn't happened yet.

facet-core/src/types/value.rs
```diff
@@ -723,6 +767,7 @@ pub struct ValueVTableBuilder<T> {
     drop_in_place: Option<DropInPlaceFn>,
     invariants: Option<InvariantsFnTyped<T>>,
     parse: Option<ParseFnTyped<T>>,
+    to_string: Option<ToStringFnTyped<T>>,
```
fasterthanlime, line 770: The standard ToString is just an extension trait implemented on anything that implements `Display`, right? I'm not clear on how this is actually different from Display.

If this is actually `serialize_to_string` (and `deserialize_from_string`), as a version of `ToString` / `FromStr` that can be used for text formats like JSON with no "date types" etc., then I think we should consider naming it differently.

fasterthanlime:
@tversteeg I see this has been closed, are you planning on doing it as another pull request? or are you waiting on ShapeFlags?

PR #621, opened 2025-05-15
feat(args): `Cli` Format trait impl; deserialize `&str` field

### LINKED ISSUE #622: Source ID for `DeserError` is hardcoded as "json"
It looks like "json" is hardcoded in the `source_id`:

https://github.com/facet-rs/facet/blob/f3edc1243582781332684d2aaea9d4ff84bb42dd/facet-deserialize/src/error.rs#L234

Would this be better to live in the Format?

```rust
    fn format_name() -> &'static str;
```

lmmx:
(Take 2! Previous version: #575)

- **feat: facet-args `Cli` trait impl; deserialize `&str` field**

Building on #619 (literally a continuation of that branch, I'll rebase if needed or I presume this PR can be merged instead of it) which allowed me to **actually use** the data passing mechanism of `NextData` rather than using dummy bytes and global variables (which in hindsight was obviously a bad design)

## Deletions

- [x] Field parsing is done in the trait, so `parse.rs` is no longer needed
- [x] Defaults (including absent bool is default falsey) are now handled in the trait so `defaults.rs` is no longer needed
- [x] Argument diagnostics are now handled in DeserError so custom ArgsError is no longer needed

## Changes

- [x] `facet-args` can now read `&str` fields
- [x] `facet-args` now uses the `Format` trait from `facet-deserialize`
- [x] Allowed source ID to be configured in error output
  - Now says "args" and the position/total number of CLI arg chars rather than "json")
  - Resolves #622
- [x] Use arg lengths to calculate ranges for diagnostics, including highlighting the space after an arg to hint when an argument should have _followed_ another flag
- [x] Single char field names (e.g. "x") no longer need to specify the value `short = "x"`, just `short` will suffice for `-x`

## Diagnostics

New arg-spanning diagnostics (without incrementing our position per-byte, only converted for diagnostics)

- The diagnostics have been improved to underline the specific arg of relevance (if there is one), or to underline the gap between the args / point to the end of the args if we were expecting an arg _after_ another arg

Screenshot from 2025-05-15 23-31-55


lmmx:
Args can now show (helpfully) where a value was expected in the intermediate position i.e. after a particular flag
Screenshot from 2025-05-15 22-53-16

fasterthanlime:
Merged manually as 0573a9e56076dcae9af16db4c5068be44cc643b8

PR #625, opened 2025-05-16
Implement recursive descent parsing for facet-json

ThouCheese:


fasterthanlime:
Have you considered "just" using a PEG?

ThouCheese:
@fasterthanlime Do you mean creating a PEG and then feeding that into a parser generator?

ThouCheese:
Really cool idea, really hard to maintain, maybe revisit in the future when stuff has settled down a bit more, but even then think about it before starting to type for once

PR #629, opened 2025-05-16
Support &str slices in facet-pretty

### LINKED ISSUE #626: facet-pretty: unsupported peek variant for &str
When calling `FacetPretty::pretty` for a `&str` (or a type containing one), facet-pretty gives the message `unsupported peek variant: "<&str contents>"`.

I'm not familiar with the codebase but I believe this is because `&str`'s def on its shape is `Undefined`.

fasterthanlime:
Closes #626

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -535,6 +535,7 @@ dependencies = [
  "facet-core",
  "facet-reflect",
  "facet-testhelpers 0.17.2",
+ "insta",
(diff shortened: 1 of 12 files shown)


fasterthanlime:
So this patches #626, the most egregious case, but we still can't pretty-print something like `&[u8]` for example — so it's _still_ a good first issue I suppose.

PR #635, opened 2025-05-17
Update list vtables to allow for more types of lists

### LINKED ISSUE #578: Make lists more flexible

kylewlacy:
Closes #578, should unblock #597 (I think)

This PR updates `ListVTable` with the following changes:

- Make `as_ptr` / `as_mut_ptr` fields optional
- Add new `get` / `get_mut` fields to get an element by index in the list
- Add new `iter_vtable` field: a vtable for creating/using/destroying an iterator for the list
  - See these PRs for more context: #617, #618
  - The iterators aren't all that useful when the `get` method is O(1), but that also means that more unusual collections can fit into `ListDef` more naturally (e.g. `LinkedList`, which can't efficiently implement `get` but can still iterate efficiently... relatively speaking)

I also updated `PeekList` and `PeekListLike` to account for lists that have / don't have `as_ptr`. Basically, it tries to use `as_ptr` as a fast path if available, but falls back to using `iter_vtable` to create an iterator if needed.

Also note that, the code paths where `as_ptr` is _not_ set are never hit, since `Vec` provides `as_ptr`. For validation, I commented out the `as_ptr` and `as_mut_ptr` methods on `Vec`, and tests still passed both normally and under Miri.

### DIFF
--- a/facet-core/src/impls_alloc/vec.rs
+++ b/facet-core/src/impls_alloc/vec.rs
@@ -1,6 +1,7 @@
 use crate::*;
 use core::hash::Hash as _;
 
+use alloc::boxed::Box;
 use alloc::vec::Vec;
 
 type VecIterator<'mem, T> = core::slice::Iter<'mem, T>;
--- a/facet-core/src/impls_alloc/vec.rs
+++ b/facet-core/src/impls_alloc/vec.rs
@@ -3,6 +3,8 @@ use core::hash::Hash as _;
 
 use alloc::vec::Vec;
 
+type VecIterator<'mem, T> = core::slice::Iter<'mem, T>;
+
 unsafe impl<'a, T> Facet<'a> for Vec<T>
 where
     T: Facet<'a>,
@@ -123,6 +125,30 @@ where
                                     let vec = ptr.as_mut::<Self>();
                                     PtrMut::new(vec.as_mut_ptr())
                                 })
+                                .iter_vtable(
+                                    IterVTable::builder()
+                                        .init_with_value(|ptr| unsafe {
+                                            let vec = ptr.get::<Self>();
+                                            let iter: VecIterator<T> = vec.iter();
+                                            let iter_state = Box::new(iter);
+                                            PtrMut::new(Box::into_raw(iter_state) as *mut u8)
+                                        })
+                                        .next(|iter_ptr| unsafe {
+                                            let state = iter_ptr.as_mut::<VecIterator<'_, T>>();
+                                            state.next().map(|value| PtrConst::new(value))
+                                        })
+                                        .next_back(|iter_ptr| unsafe {
+                                            let state = iter_ptr.as_mut::<VecIterator<'_, T>>();
+                                            state.next_back().map(|value| PtrConst::new(value))
+                                        })
+                                        .dealloc(|iter_ptr| unsafe {
+                                            drop(Box::from_raw(
+                                                iter_ptr.as_ptr::<VecIterator<'_, T>>()
+                                                    as *mut VecIterator<'_, T>,
+                                            ));
+                                        })
+                                        .build(),
+                                )
                                 .build()
                         },
                     )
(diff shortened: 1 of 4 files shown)


facet-core/src/impls_alloc/vec.rs
```diff
@@ -3,6 +3,8 @@ use core::hash::Hash as _;
 
 use alloc::vec::Vec;
 
+type VecIterator<'mem, T> = core::slice::Iter<'mem, T>;
+
```
kylewlacy, line 7: I added this type alias to ensure that all the methods in `iter_vtable` align when casting from/to `PtrMut`

facet-reflect/src/peek/list.rs
```diff
@@ -1,24 +1,38 @@
...
+use facet_core::{ListDef, PtrConst, PtrMut};
 
 /// Iterator over a `PeekList`
 pub struct PeekListIter<'mem, 'facet_lifetime> {
-    list: PeekList<'mem, 'facet_lifetime>,
+    state: PeekListIterState<'mem>,
     index: usize,
     len: usize,
+    def: ListDef,
+    _list: PhantomData<Peek<'mem, 'facet_lifetime>>,
```
kylewlacy, line 11: I'm fuzzy on what `'facet_lifetime` is for, but I _think_ this `PhantomData` gives `PeekListIter` the same variance that had before this PR?

facet-reflect/src/peek/list_like.rs
```diff
@@ -35,21 +35,33 @@ impl ListLikeDef {
 
 /// Iterator over a `PeekListLike`
 pub struct PeekListLikeIter<'mem, 'facet_lifetime> {
-    list: PeekListLike<'mem, 'facet_lifetime>,
+    state: PeekListLikeIterState<'mem>,
     index: usize,
     len: usize,
+    def: ListLikeDef,
+    _list: PhantomData<Peek<'mem, 'facet_lifetime>>,
```
kylewlacy, line 42: https://github.com/facet-rs/facet/pull/635/files#r2094086373

facet-reflect/src/peek/list.rs
fasterthanlime: We allow exactly one lifetime for types that derive `Facet`, and it is invariant. That's what `facet_lifetime` is. I think we should call it `borrow` or something, but it's hard to name correctly :(

fasterthanlime: I opened https://github.com/facet-rs/facet/issues/636 to track the naming issue btw

fasterthanlime (approved):

PR #640, opened 2025-05-18
Remove nonsensical trait bound

### LINKED ISSUE #638: I don't think this trait bound makes sense
In `facet-json` (because of `facet-serialize`):

```rust
/// Serializes a value to JSON
#[cfg(feature = "std")]
pub fn to_string<'input, 'facet, T: Facet<'facet>>(value: &'input T) -> String
where
    'input: 'facet,
{
    recursive::to_string(value, 0)
}
```

I don't think `'input: 'facet` makes sense, but if removed, it breaks things.

We need to have a little lifetime review in the codebase, me thinks.

fasterthanlime:
Closes #638

### DIFF
--- a/facet-reflect/src/peek/value.rs
+++ b/facet-reflect/src/peek/value.rs
@@ -44,7 +44,7 @@ pub struct Peek<'mem, 'facet, 'shape> {
     /// Shape of the value
     pub(crate) shape: &'shape Shape<'shape>,
 
-    invariant: PhantomData<fn(&'facet ()) -> &'facet ()>,
+    phantom: PhantomData<&'facet ()>,
 }
 
 impl<'mem, 'facet, 'shape> Peek<'mem, 'facet, 'shape> {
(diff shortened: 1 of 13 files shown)


facet-reflect/src/peek/value.rs
```diff
@@ -44,7 +44,7 @@ pub struct Peek<'mem, 'facet, 'shape> {
     /// Shape of the value
     pub(crate) shape: &'shape Shape<'shape>,
 
-    invariant: PhantomData<fn(&'facet ()) -> &'facet ()>,
+    phantom: PhantomData<&'facet ()>,
```
Veykril, line 47: I don't think this is correct. If we make this lifetime covariant, then that allows `Peek::get` to return a bigger contravariant lifetime than what the facetted type contains. That is, the compiletest here will start compiling https://github.com/facet-rs/facet/pull/642/files#diff-b322a88de4968af8e4616fb2dc528ef3f69e3143a6a5190a56a7641155653d0b

fasterthanlime: Alright, then that means we cannot use Peek in StackRunner for facet-serialize, which.. is okay I guess, we can store its components separately.

PR #641, opened 2025-05-18
Refactor serialization to avoid reversing or collecting iterators

kylewlacy:
This PR makes it so `facet-serialize` no longer requires [`DoubleEndedIterator`](https://doc.rust-lang.org/stable/std/iter/trait.DoubleEndedIterator.html) when iterating over values. In turn, this also means that serialization can use plain iterator values in more places, without resorting to collecting e.g. into a `Vec`.

The "trick" is in `SerialzeTask`, which now holds iterators wherever possible. When popped from the stack, it gets the next value from the iterator then pushes it back onto the stack (until the iterator is empty).

I was hoping this would lead to perf improvements, but locally, I wasn't seeing a noticeable difference... I still felt like this change had other minor benefits (e.g. less memory use during iteration, `iter_vtable` for `HashMap` no longer needs to collect its keys into a vec, etc.)

### DIFF
--- a/facet-reflect/src/peek/fields.rs
+++ b/facet-reflect/src/peek/fields.rs
@@ -5,7 +5,7 @@ use facet_core::{Field, FieldFlags};
 use crate::Peek;
 use alloc::{vec, vec::Vec};
 
-use super::{PeekEnum, PeekStruct};
+use super::{PeekEnum, PeekStruct, PeekTuple};
 
 /// Trait for types that have field methods
 ///
@@ -16,9 +16,7 @@ pub trait HasFields<'mem, 'facet, 'shape> {
     fn fields(&self) -> FieldIter<'mem, 'facet, 'shape>;
 
     /// Iterates over fields in this type that should be included when it is serialized
-    fn fields_for_serialize(
-        &self,
-    ) -> impl Iterator<Item = (Field<'shape>, Peek<'mem, 'facet, 'shape>)> {
+    fn fields_for_serialize(&self) -> FieldsForSerializeIter<'mem, 'facet, 'shape> {
         FieldsForSerializeIter {
             stack: vec![self.fields()],
         }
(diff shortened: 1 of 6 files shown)


fasterthanlime:
I have been wanting to do that for a while, so thank you for that

facet-reflect/src/peek/fields.rs
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+}
+
+impl ExactSizeIterator for FieldIter<'_, '_, '_> {}
+
+/// An iterator over the fields of a struct or enum that should be serialized. See [`HasFields::fields_for_serialize`]
+pub struct FieldsForSerializeIter<'mem, 'facet, 'shape> {
+    stack: Vec<FieldIter<'mem, 'facet, 'shape>>,
+}
```
kylewlacy, line 157: This type is a hand-written iterator to re-create the old `fields_for_serialize()` impl Trait method. This was necessary so that `facet-serialize` could store a nameable iterator type (I also tried `Box<dyn Iterator<...>>` but hit some weird lifetime bounds issues... plus I didn't feel great about needing an extra box for this

(One minor plus to this implementation is that it's no longer recursive!)

facet-serialize/src/lib.rs
```diff
@@ -473,9 +494,11 @@ where
                                 let fields = peek_struct.fields_for_serialize().count();
```
kylewlacy, line 494: Oh, one really subtle thing that I hate!! this is iterating over all the flattened fields twice: once here to count the number of flattened fields, and once while doing the actual serialization

This is needed so we can call `serializer.start_object(Some(_))`, but I couldn't figure out an efficient way to count the number of flattened fields without just iterating (including allocating a vec)...

If there's a way to impl `ExactSizedIterator` for `FieldsForSerializeIter` then we could avoid this

fasterthanlime: > If there's a way to impl `ExactSizedIterator` for `FieldsForSerializeIter` then we could avoid this

Not sure about this, but maybe we can remove the hint...

fasterthanlime (approved):

PR #643, opened 2025-05-18
Allow nav to scroll on facet.rs website

gennyble:
I noticed that the `<nav>`, on the right of the screen, was overflowed and the text was running vertically past the end of it's container. Letting the y axis scroll fixed this :)

wasn't sure how to organize the CSS, but after the height seemed like a smart place.


fasterthanlime:
Thanks!!

fasterthanlime:
Deployed!

PR #647, opened 2025-05-20
Use std iterators for some collections

kylewlacy:
This PR replaces the custom iterators for various collection types, to instead use the built-in iterators from `std` (or `alloc`). This applies to `HashMap`, `HashSet`, `BTreeMap`, and `BTreeSet`. In addition to simplifying the implementation, this also means we no longer need to `.collect()` for `HashMap` or `HashSet` when creating the iterator!

This was unblocked by #641 (previously, `HashMap` needed a custom iterator that buffered the keys in order to implement `DoubleEndedIterator`)

As part of this change, the iter vtable for `HashMap` no longer provides an `next_back`, while the iter vtable for `BTreeSet` now _does_ provide it

### DIFF
--- a/facet-core/src/impls_alloc/btreemap.rs
+++ b/facet-core/src/impls_alloc/btreemap.rs
@@ -1,19 +1,13 @@
 use core::write;
 
-use alloc::{
-    boxed::Box,
-    collections::{BTreeMap, VecDeque},
-};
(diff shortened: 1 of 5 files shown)


fasterthanlime:
Awesome work there, thanks so much!

PR #656, opened 2025-05-20
Fix marker traits for references and also test marker traits in facts

tyilo:
I don't know why the last test in `test_mut_ref` doesn't work when checking its `PartialEq` and `Ord` implementations. Seems like a bug in `Facet` somewhere...


tyilo:
I think I found the problem with the `test_mut_ref`: See https://github.com/facet-rs/facet/pull/657 and https://github.com/facet-rs/facet/issues/658

tyilo:
Rebasing on top of https://github.com/facet-rs/facet/pull/657 fixes the `PartialEq` and `Ord` for `test_mut_ref`.

PR #657, opened 2025-05-20
Fix UB for unsized types

tyilo:


tyilo:
I fixed the main issue, but there is still a problem with `&&str`, probably due to the `thin ptr -> wide ptr`.

tyilo:
For this to be fixed I think `PtrConst` needs to track the metadata for wide pointers, but to do this I could no longer be `#[repr(transparent)]` and thus the `transmute`'s from `T`-typed to `PtrConst`-typed vtable functions would no longer work...

fasterthanlime:
> For this to be fixed I think `PtrConst` needs to track the metadata for wide pointers, but to do this it could no longer be `#[repr(transparent)]` and thus the `transmute`'s from `T`-typed to `PtrConst`-typed vtable functions would no longer work...

I have been wondering how to make `PtrConst` work with trait objects.. and have no answer for now.

tyilo:
I think we can merge this now unless we want more tests.

fasterthanlime:
Well, I had uses for `Arc<dyn Trait>` but I guess I'll find alternative ways of dealing with them :(

tyilo:
I think it should be possible to add back support for that

PR #663, opened 2025-05-21
Add support for the `bytes` crate

### LINKED ISSUE #597: Support the `bytes` crate

kylewlacy:
Closes #597

This PR adds `bytes` as an optional dependency / feature-- when enabled, it impls `Facet` for the [`Bytes`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html) and [`BytesMut`](https://docs.rs/bytes/latest/bytes/struct.BytesMut.html) types.

Turns out [I was wrong before](https://github.com/facet-rs/facet/issues/578#issue-3055805317): both `Bytes` and `BytesMut` are backed by a single contiguous slice. That means they both provide an `as_ptr` function for faster iteration

Dealing with `Bytes` still ended up being a little tricky because it's immutable. I felt the most direct option for now was to make mutating functions from `ListVTable` optional, making it effectively immutable. That means you can't e.g. deserialize `Bytes` (unless it's empty), since `Wip` can't insert any elements.

(I wonder if there would be a way to make it so `Wip<Bytes>` could be backed by a `BytesMut` behind the scenes, converting to `Bytes` at zero-cost via [`BytesMut::freeze`](https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.freeze) only once the type is finished...?)

### DIFF
--- a/facet-json/tests/option.rs
+++ b/facet-json/tests/option.rs
@@ -17,15 +17,48 @@ fn test_from_json_with_option() {
     }
 
     let json = r#"{
-    "name": "Alice",
-    "age": null,
-    "inner": {
-        "foo": 42
-    }
-}"#;
+        "name": "Alice",
+        "age": null,
+        "inner": {
+            "foo": 42
+        }
+    }"#;
 
     let test_struct: Options = from_str(json)?;
     assert_eq!(test_struct.name.as_deref(), Some("Alice"));
     assert_eq!(test_struct.age, None);
     assert_eq!(test_struct.inner.as_ref().map(|i| i.foo), Some(42));
+}
+
+#[test]
+fn test_from_json_with_nested_options() {
+    #[derive(Facet)]
+    struct Options {
+        name: Option<Option<String>>,
+        age: Option<Box<u32>>,
+        inner: Option<Box<Option<Inner>>>,
+    }
+
+    #[derive(Facet)]
+    struct Inner {
+        foo: i32,
+    }
+
+    let json = r#"{
+        "name": "Alice",
+        "age": 5,
+        "inner": {
+            "foo": 42
+        }
+    }"#;
+
+    let test_struct: Options = from_str(json)?;
+    assert_eq!(test_struct.name.flatten().as_deref(), Some("Alice"));
+    assert_eq!(test_struct.age, Some(Box::new(5)));
+    assert_eq!(
+        test_struct
+            .inner
+            .and_then(|inner| inner.map(|inner| inner.foo)),
+        Some(42)
+    );
 }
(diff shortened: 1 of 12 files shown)


fasterthanlime:
> (I wonder if there would be a way to make it so `Wip<Bytes>` could be backed by a `BytesMut` behind the scenes, converting to `Bytes` at zero-cost via [`BytesMut::freeze`](https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.freeze) only once the type is finished...?)

I'm facing similar issues with `Arc<T>` right now x) it's fun!

fasterthanlime:
This is neat, but I think we should at least be able to deserialize before merging this. I don't know. Feels incomplete without that.

kylewlacy:
> This is neat, but I think we should at least be able to deserialize before merging this. I don't know. Feels incomplete without that.

Yeah I think that's fair, it'd especially be surprising because serializing works be deserializing doesn't... I'll mark this PR as a draft for now

fasterthanlime:
I can take it over if needed? I think deserializing to a Vec of u8 in an intermediate frame would work

kylewlacy:
> I can take it over if needed? I think deserializing to a Vec of u8 in an intermediate frame would work

I'm pretty indifferent to be honest-- feel free to pick it up if you want, otherwise I should have time to revisit in the next few days :slightly_smiling_face:

fasterthanlime:
> otherwise I should have time to revisit in the next few days 🙂

Cool, I'll let you get to it at your own pace then :)

kylewlacy:
Alright ready for review! After looking over the different options, I felt like just setting the inner shape for `Bytes` to `BytesMut` would be the most straightforward path. I also did a bit of work on the deserializer side to handle inner shapes too (basically handled just like smart pointers, but still a separate code path)

facet-json/tests/option.rs
```diff
@@ -17,15 +17,48 @@ fn test_from_json_with_option() {
...
+    }"#;
 
     let test_struct: Options = from_str(json)?;
     assert_eq!(test_struct.name.as_deref(), Some("Alice"));
     assert_eq!(test_struct.age, None);
     assert_eq!(test_struct.inner.as_ref().map(|i| i.foo), Some(42));
 }
+
+#[test]
+fn test_from_json_with_nested_options() {
```
kylewlacy, line 34: Basically unrelated to this PR, but the changes in the deserializer incidentally made this work... although for some reason, setting `"inner": null` still fails to deserialize

fasterthanlime:
> I felt like just setting the inner shape for `Bytes` to `BytesMut` would be the most straightforward path.

That's true. At this point, it's really hard to grasp the semantics of inner shapes, but it makes sense from a deserialization standpoint 🙃

fasterthanlime (approved):

PR #664, opened 2025-05-21
feat(deserialize,args): pop Subspan stack; `--key=val` CLI flags

lmmx:
**Summary**: populate the Substack in the state machine lender iterator data-passing style via NextData

- Adds a new `Outcome::Resegment` enum variant intended for use with `Format::SpanType = Raw` (e.g. facet-args)
  - In these formats, a `Span` is a meaningful unit, but in some cases may not actually be a good unit on first pass
  - For example, it is equivalent in a CLI to pass `["--key", "value"]` or `["--key=value"]` and we would want to be able to 'resegment' the latter to appear like the former at deserialisation

- The Format's `next()` method is allowed (opt-in) to find more meaningful units, put them on a 'substack' that lives on the runner alongside the regular stack, where they live for the duration of the individual arg's processing (hence "subspan")

- The format's `next()` method must pass spans of length 0, aliased as `stay_put`, until the last Subspan on the `nd.runner.substack` (made public via `substack()`) is processed and then move on by passing a Span of length 1, aliased as `step_forth`)
  - The `Substack` checking in the `next!` macro in `deserialize_wip` is compiled out via a const generic bool set via trait bound on the `Format::SpanType` (Raw/Cooked), i.e. will never be checked and will never be set
    - NB: it can only be set in this part of the code, the runner is non-mutable from the format's `next()` method itself (courtesy of a lending iterator ownership transfer)

- Adds a new `Instruction::SubstackClose` enum variant to trigger stack-based clearing, so the runner stack controls how long the substack persists (it should only persist for the duration of a Span, but we don't need to introduce any counter/index state to monitor when we're about to leave this, the existing stack suffices)

## Demo

```rust
#[test]
fn test_eq() {
    #[derive(Facet, Debug)]
    struct Args {
        #[facet(named)]
        hello: String,
    }
    let args: Args = facet_args::from_slice(&["--hello=world"])?;
    assert_eq!(args.hello, "world".to_string());
}
```

## Key parts

- `Substack` is created in `deserialize_wip`

https://github.com/facet-rs/facet/blob/858c8ceb2487b1e8ef955561b624011947c2cb34/facet-deserialize/src/lib.rs#L445

- It lives on the `StackRunner`

https://github.com/facet-rs/facet/blob/858c8ceb2487b1e8ef955561b624011947c2cb34/facet-deserialize/src/lib.rs#L619-L620

- A `Substack` is a struct with `Option<Vec<Subspan>>` in its `spans`

https://github.com/facet-rs/facet/blob/858c8ceb2487b1e8ef955561b624011947c2cb34/facet-deserialize/src/span.rs#L134-L137

https://github.com/facet-rs/facet/blob/858c8ceb2487b1e8ef955561b624011947c2cb34/facet-deserialize/src/span.rs#L108-L131

- For formats whose `SpanType` (generic type param `C`) is Cooked (the default), it is impossible to add to the substack (i.e. the substack is opt-in by the format).

https://github.com/facet-rs/facet/blob/858c8ceb2487b1e8ef955561b624011947c2cb34/facet-deserialize/src/span.rs#L188-L214

- ...so formats with SpanType = Cooked like facet-json etc. will only ever be able to have a `None` in the `substack` field, never a `Vec`

https://github.com/facet-rs/facet/blob/858c8ceb2487b1e8ef955561b624011947c2cb34/facet-deserialize/src/span.rs#L216-L228

- Specifically the access gets compiled out of the `next!` macro for non-`Formats::SpanType = Raw` formats like facet-json which don't use `Subspan`s :checkered_flag: here:

https://github.com/facet-rs/facet/blob/858c8ceb2487b1e8ef955561b624011947c2cb34/facet-deserialize/src/lib.rs#L471-L477

- and the clone gets compiled out here:

https://github.com/facet-rs/facet/blob/858c8ceb2487b1e8ef955561b624011947c2cb34/facet-deserialize/src/lib.rs#L479-L483

### DIFF
--- a/facet-args/src/arg.rs
+++ b/facet-args/src/arg.rs
@@ -1,5 +1,7 @@
 use alloc::borrow::Cow;
+use facet_deserialize::{Subspan, SubspanMeta};
 
+#[derive(Debug)]
 pub(crate) enum ArgType<'a> {
     LongFlag(Cow<'a, str>),
     ShortFlag(&'a str),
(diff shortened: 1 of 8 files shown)


lmmx:
Diagnosed the source of the bug: the substack is not in fact cleared when advancing to the next span (i.e. argument)

Screenshot from 2025-05-21 23-30-17

lmmx:
**Fixed** the fix was

```rust
            if outcome.span.start() != $runner.last_span.start() {
                // Substack expires after a span, clear substack
                trace!("Substack: {} {:?}", "set".cyan(), $runner.substack.get());
                $runner.substack.clear();
            } else if !$runner.substack.get().is_empty() {
                trace!("Substack: {}", "carried".cyan());
            } else {
                trace!("Substack: {}", "-".red());
            }
```

fasterthanlime:
> * This introduces an `Option<Vec<Subspan>>` onto `StackRunner` but any usage is eliminated via const generic (`const USES_SUBSTACK = false`) so it'll just stay as `None` if not used 😸

Very nice!!

Issue #631, opened 2025-05-16
Facet might think a struct is Send (b/c all its fields are) when the struct actually isn't


fasterthanlime:
An example is `std::sync::MutexGuard` — it has an `impl !Send`. Facet would _believe_ it's Send, when it's really not.

Also, any struct that contains an `UnsafeCell` will have Facet believe it's _not_ Send, when it could be (through `unsafe impl Send for X`).

The fix for that is: documenting it.

thx to @lcnr for raising this!

PR #666, opened 2025-05-22
Add indirection to vtable fns to fix cyclic types

### LINKED ISSUE #644: Make `clone_into` and others fallible-at-runtime, since we can't check them at const time.
As of https://github.com/facet-rs/facet/commit/4530132326a7d21a436b52bf60c94c8d76d5300e, this code:

```rust
#[derive(Debug, Clone, facet::Facet)]
struct Foo {
    inner: Vec<Foo>,
}
```

...results in this compile error:

```
error[E0391]: cycle detected when simplifying constant for the type system `<impl at src/main.rs:1:24: 1:36>::SHAPE`
--> src/main.rs:1:24
    |
1   | #[derive(Debug, Clone, facet::Facet)]
    |                        ^^^^^^^^^^^^
    |
note: ...which requires const-evaluating + checking `<impl at src/main.rs:1:24: 1:36>::SHAPE`...
--> src/main.rs:1:24
    |
1   | #[derive(Debug, Clone, facet::Facet)]
    |                        ^^^^^^^^^^^^
(... lots more output ...)
```

After some digging, I think the error boils down to this line: https://github.com/facet-rs/facet/blob/4530132326a7d21a436b52bf60c94c8d76d5300e/facet-core/src/impls_alloc/vec.rs#L26

...which is trying to eagerly evaluate `Foo::SHAPE` within the definition of `Foo::SHAPE` itself. In this case, this is due to the implementation on `Vec<T>`, but this same failure would occur for any shape that eagerly references a generic type's shape (I did some quick spot checks-- it looks like `&[T]` also fails, but `Option<T>`, `Box<T>`, and `&T` are safe today?)

kylewlacy:
Closes #644

This PR updates all of the optional functions in `VTableValue` to use an extra layer of indirection (e.g. replacing `debug: Option<DebugFn>` with `debug: fn() -> Option<DebugFn>`). Also, the marker traits were given the same treatment: the vtable now stores `marker_traits: fn() -> MarkerTraits`

This extra layer of indirection is required to support cyclic types. There are a lot of types that opt-in to optional functions or marker traits depending on if its generic parameter(s) implement the function/trait

This does give up some const-ness. It's no longer possible to determine which traits a shape implements within a `const` context, even marker traits...

### DIFF
--- a/facet-core/src/impls_alloc/arc.rs
+++ b/facet-core/src/impls_alloc/arc.rs
@@ -148,9 +148,9 @@ unsafe impl<'a, T: Facet<'a>> Facet<'a> for alloc::sync::Arc<T> {
             Ok(())
         });
 
-        vtable.try_from = Some(try_from::<T>);
-        vtable.try_into_inner = Some(try_into_inner::<T>);
-        vtable.try_borrow_inner = Some(try_borrow_inner::<T>);
+        vtable.try_from = || Some(try_from::<T>);
+        vtable.try_into_inner = || Some(try_into_inner::<T>);
+        vtable.try_borrow_inner = || Some(try_borrow_inner::<T>);
         vtable
     };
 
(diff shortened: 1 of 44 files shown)


fasterthanlime:
> This does give up some const-ness. It's no longer possible to determine which traits a shape implements within a `const` context, even marker traits...

Well, apparently this was not very precise to begin with, see #631

fasterthanlime (approved):

PR #678, opened 2025-05-25
some binary size tests?

ThouCheese:


.github/workflows/code-size.yml
```diff
@@ -0,0 +1,166 @@
...
+              comment += '\n\n' + bloatData + '\n\n*(See workflow artifacts for full details)*';
+            }
+
+            // Post comment
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: comment
+            });
```
fasterthanlime, line 136: I would probably write all this as a Rust tool itself (could be in a separate repo) or as a TypeScript codebase at least — this feels way too fragile, but 🤷 could be a start.

fasterthanlime:
Are we okay closing this now that cargo-bloat & cargo-llvm-lines are ran in CI and reported as a comment?

PR #679, opened 2025-05-25
feat(args): Support Vec type fields as CLI args

lmmx:
Let's get some `Vec`s!


lmmx:
Hmm no `Vec`s yet

Screenshot from 2025-05-26 00-01-45

lmmx:
Now got scalar parsing and list starting, but still not list creation (need to copy some aspect of the pushback approach?)

Screenshot from 2025-05-26 00-46-17

fasterthanlime:
Closing since facet-args is now in https://github.com/facet-rs/facet-args

PR #700, opened 2025-05-30
Solve code comment inconsistencies and add missing validations

YichiZhang0613:
Fix problems mentioned in #701.

### DIFF
--- a/facet-core/src/types/def/iter.rs
+++ b/facet-core/src/types/def/iter.rs
@@ -119,6 +119,9 @@ impl<T: IterItem> IterVTableBuilder<T> {
     ///
     /// This method will panic if any of the required fields are `None`.
     pub const fn build(self) -> IterVTable<T> {
+        assert!(self.init_with_value.is_some());
+        assert!(self.next_back.is_some());
+        assert!(self.size_hint.is_some());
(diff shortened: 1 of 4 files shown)


facet-core/src/types/def/iter.rs
```diff
@@ -119,6 +119,9 @@ impl<T: IterItem> IterVTableBuilder<T> {
     ///
     /// This method will panic if any of the required fields are `None`.
     pub const fn build(self) -> IterVTable<T> {
+        assert!(self.init_with_value.is_some());
+        assert!(self.next_back.is_some());
```
fasterthanlime: I don't think that's right — it's possible for iterators to not support next_back or not have size hint and that's fine, right?

YichiZhang0613: > I don't think that's right — it's possible for iterators to not support next_back or not have size hint and that's fine, right?

OK, I have modified correspondingly.

fasterthanlime:
I'll review again once the tests pass :) The comment update is good though.

fasterthanlime (approved):

fasterthanlime:
Thanks for your contribution!

PR #708, opened 2025-05-31
Reduce indexing in `write_json_string`

Veykril:


ThouCheese:
Where is the perf gain Lebowski

PR #711, opened 2025-06-01
Remove unnecessary clone

ThouCheese:


ThouCheese:
Well even though it does not seem to affect performance I still like it for removing an apparently needless method call

PR #712, opened 2025-06-01
Use a Vec instead of a Writer for the json serialializer

ThouCheese:


facet-json/Cargo.toml
```diff
@@ -10,20 +10,15 @@ keywords = ["json", "serialization", "deserialization", "reflection", "facet"]
 categories = ["encoding", "parser-implementations"]
 
 [features]
-std = ["facet-core/std", "alloc"] # Uses libstd and alloc
```
fasterthanlime, line 13: Bit confused by this removal — in the case of std I guess we would have the Write adapter, and other std-only things?

fasterthanlime (approved):

PR #713, opened 2025-06-01
flake: formatter: set to alejandra

spikespaz:
This chooses Alejandra as the flake's formatter. For VS Code, this plays well with:
```json
"nix.serverSettings": {
  "nixd": {
    "formatting": {
      "command": [
        "nix",
        "fmt",
        "--",
        "--"
      ]
    }
  }
},
```

To format treewide on the command line, use `nix fmt -- .`.

I chose Alejandra for the formatter because despite all of the nixpkgs team's effort to force people to use `nixfmt-rfc-style`, many people still choose Alejandra because of the familiar style that closely resembles hand-written code. The original `nixfmt`, before the takeover (still available as `nixfmt-classic`) probably makes the most people happy, but doesn't get updates anymore. `nixfmt-classic` is the most diff-friendly one, but Alejandra is more common.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,36 +7,38 @@
   # shell.nix compatibility
   inputs.flake-compat.url = "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz";
 
-  outputs = { self, nixpkgs, ... }:
-    let
-      # System types to support.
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Thanks!

PR #718, opened 2025-06-02
Add bigapi-cli-facet crate, add cargo bloat workflow

fasterthanlime:


fasterthanlime:
~I guess I'm going to have to iterate on that one on main~

edit: Ah, no.

PR #726, opened 2025-06-02
facet-toml: ser: TomlSerializer: move traces to avoid debug-only args

spikespaz:
It feels strange to have functions take arguments that are only used in debug contexts, and have no potential for future use. This commit removes the `type_name` argument from `push_key` and `pop_key`. This argument had a misleading name to begin with.

Instead, the traces are moved to a position that has all of the required context; there is no further need to pass around static strings for debugging purposes, keeping business logic clean.


spikespaz:
Note that I'm working my way toward refactoring `TomlSerializer::write_value`, so that it makes sense to remove the `trace!` there as well. As yet, that's the best spot for it.

PR #728, opened 2025-06-03
flake: stylistic improvements

spikespaz:
This is my attempt to tidy up the flake a little more. These are opinions, but popular ones (within my circle at least).


fasterthanlime:
There's conflicts due to previous commits now, and because you're using an org to contribute, GitHub doesn't have "let maintainers edit that PR" functionality, so I can't fix that PR (I could check it out locally for cleanup and merge it myself though).

PR #747, opened 2025-06-05
Introduce Sized/Unsized vtable split

fasterthanlime:


fasterthanlime:
@tyilo I think it makes sense to merge a part where unsized vtables are not used a lot yet, and then merge a second PR where they are starting to get used. All the tests are passing with this current version, so... I think I'm going to merge it now.

PR #759, opened 2025-06-06
Feature unsynn quote

cehteh:
Yo Amos!

I see you like quote!, so I removed quote from facet and put quote in unsynn, now you can quote without depending or quote! (also the proc_macro2 dependency is gone)

Short story, long:
This is WIP/anticipatory of a not yet released unsynn 0.2 which will include its own quote! macro. for $reasons it will be slightly different (simpler) to the quote crate. there are no `#()` repetitions yet. But this can be easily handled see the patch.

Initially i was thinking i could just `impl quote::ToTokens for dyn unsynn::ToTokens ..` but unfortunally this didnt worked out, either because the `quote::quote!()` macro can't handle that or because the trait solver isn't smart enough. Anyway, i just got my "Hold my beer" moment yesterday: How complicated can a `quote!{}` in unsynn be? .. turns out its pretty trivial to do the basics. So unsynn has its own lean `quote!{}` macro now.

This is yet just a Draft PR (don't merge!) up for comment. For now I only touched `facet-macros-parse` I'll see to remove proc_macro2 deps (its reexported from unsynn) and quote from the rest of facet while going over it. Its a nice test case to me to see where unsynn lacks features.

### DIFF
--- a/facet-macros-parse/Cargo.toml
+++ b/facet-macros-parse/Cargo.toml
@@ -16,8 +16,7 @@ keywords = [
 categories = ["development-tools", "encoding"]
 
 [dependencies]
-unsynn = "0.1.0"
-quote = "1.0.40"
+unsynn = "0.2"
 
 [features]
 function = []
--- a/facet-macros-parse/src/function/fn_shape_input.rs
+++ b/facet-macros-parse/src/function/fn_shape_input.rs
@@ -41,7 +41,6 @@ pub fn parse_fn_shape_input(input: TokenStream) -> ParsedFnShapeInput {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use quote::quote;
 
     #[test]
     fn test_simple_function_name() {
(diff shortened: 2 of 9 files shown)


fasterthanlime:
We got https://github.com/facet-rs/facet/pull/905 — anything specific to know about 0.2.0 ?

cehteh:
0.2 Was mostly addition of new features, the only user-facing change was that DelimitedVec and Repeats got merged into one-size-fits all parser (more generics, configurable delimiter policy). The API still *should* be backward compatible.

0.2.1 fixes some corner cases. It still has a open bug where positions reported in errors may be incorrect (This doesn't affect parsing correctness)

0.3 is around the corner which will become a massive update, proc_macro2 will become optional then, plus many more new/improved things (everything that's needed to write a full rust parser in unsynn-rust).

fasterthanlime:
I just upgraded to 0.3.0 in another PR, so I'm gonna close this! #927

PR #763, opened 2025-06-07
facet-core: types: Shape: Debug: add alternate form

spikespaz:
I'm still working on `Display` and `Debug` for all the things, but I find myself constantly rebasing branches on this commit. I'd like to get it in, even if it's not the final form.


fasterthanlime:
Merged in #785

PR #765, opened 2025-06-08
facet-macros-emit: show generics in generated type_name fns

spikespaz:
This PR addresses the missing generics when `Display`-ing `Facet`-derived `struct`s and `enum`s. `const` generics and type parameters are shown in the order they appear; lifetimes are omitted, this is in-line with `std::any::type_name`'s behavior.

~~I'm still fiddling with the commits.~~ I'm not sure where `generate_type_name_fn` belongs, but I've put it in `facet-macros-emit/src/derive.rs`. If there's a better spot, I can rebase.

`Debug` is used to show `const` generics, for now, this is sufficient because all types that can be used in `const`-generic position implement `Debug`. For `const`-expressions, this will need to be changed, or removed.

Every time a change affects the codegen snapshots, it is followed by a commit (`run crago insta review`). If you would like these commits to be squashed, instead of separate, let me know.


spikespaz:
@fasterthanlime I have rebased on your changes to `ValueVTable`, and this is ready for review. I can squash as you please.

lmmx:
Some conflicts on this one @spikespaz , you might be able to resolve them by just `rebase -i origin/main`’ing unless there’s something tricky

facet-macros-emit/src/derive.rs
```diff
@@ -116,3 +116,54 @@ pub(crate) fn build_type_params(generics: Option<&GenericParams>) -> TokenStream
...
+            // Lifetimes not shown by `std::any::type_name`, this is parity.
+            GenericParam::Lifetime { .. } => None,
+            GenericParam::Const { name, .. } => Some(quote! {
+                write!(f, "{:?}", #name)?;
+            }),
+            GenericParam::Type { name, .. } => Some(quote! {
+                <#name as ::facet::Facet>::SHAPE.vtable.type_name()(f, opts)?;
+            }),
+        });
+        // TODO: is there a way to construct a DelimitedVec from an iterator?
```
lmmx, line 140: Is this what you’re after?

https://github.com/cehteh/unsynn/blob/236a9f29d0f52c680093645317134c6cc00f6aca/src/container.rs#L365-L393

spikespaz: https://github.com/cehteh/unsynn/commit/3d4aa5af009cf7af5722192cd35645244ee4c49d hasn't appeared in a release yet, let's leave the TODO there.

fasterthanlime:
Merged in #786

PR #766, opened 2025-06-08
facet-dev: prepush: nextest: do not fail fast

spikespaz:
I was being gaslit by `facet-dev`'s invocation of `nextest`, it was telling me that snapshots were changing at points in the history that made no sense. After chasing my tail for a while, I noticed it was because it skips any more jobs after a failure, but allows the current jobs to finish. I took its suggestion of using the `--no-fail-fast` flag. I still have no idea how `facet-dev` works, or why `nextest` changes the snapshots insead of `codegen`, why this is separate from `cargo insta test`, and so many other questions.

### DIFF
--- a/facet-dev/src/main.rs
+++ b/facet-dev/src/main.rs
@@ -884,7 +884,12 @@ fn pre_push() {
         },
         CommandInfo {
             name: "test".to_string(),
-            command: vec!["cargo".into(), "nextest".into(), "run".into()],
+            command: vec![
+                "cargo".into(),
+                "nextest".into(),
+                "run".into(),
+                "--no-fail-fast".into(),
+            ],
         },
     ];
     let total_commands = commands.len();


fasterthanlime:
> I still have no idea how `facet-dev` works, or why `nextest` changes the snapshots insead of `codegen`, why this is separate from `cargo insta test`, and so many other questions.

Easy! Afaict `cargo insta test` is a thin wrapper that adds `--no-fail-fast`. You can do it yourself though. `cargo nextest` is required to run our test suite because it gives us that nice "process per test" isolation (required for installing a log handler, etc.)

"gen" mostly does README templating (It used to generate tuple code too but that's no longer the case). "codegen" is the old name of all that and is no longer present in there — I'm going to fix the docs in another PR.

fasterthanlime:
More helpfully maybe: if you changed the proc macro and some snapshot (insta) tests fail, you can do:

```
cargo nextest run --no-fail-fast
```

Then:

```
cargo insta review
```

`review` is actually a separate command, not just a wrapper above cargo-test.

spikespaz:
You said `cargo insta test` is a wrapper that adds `--no-fail-fast` for you, I assume you mean to something other than `nextest`? Because running `cargo insta test` shows many failures, `cargo nextest run --no-fail-fast` succeeds.

spikespaz:
So I've switched up my habits to run `cargo nextest run --no-fail-fast` manually, and to some degree, this PR lost value.

I'd like to know if you think it's worth adding `--no-fail-fast` by default (to save a step or prevent running tests twice in some scenarios) before I close this.

On one hand, a developer may want fast feedback to fix problems in small batches. On the other, the subcommand is called `prepush`, and in that scenario I can see it still being useful to run all tests, and generate new snapshots completely.

fasterthanlime:
> Because running `cargo insta test` shows many failures, `cargo nextest run --no-fail-fast` succeeds.

Mhh that seems weird, are they running different tests?

spikespaz:
It looks like they're running the same tests. On closer inspection, it looks like `color-eyre` is failing to hook, but I don't know what that means.

I assume for you they both do the same thing?

```sh
cargo nextest run --no-fail-fast > cargo_nextest_run.txt 2>&1
```
Output: [cargo_nextest_run.txt](https://github.com/user-attachments/files/20697386/cargo_nextest_run.txt)

```sh
cargo insta test > cargo_insta_text.txt 2>&1
```
Output: [cargo_insta_test.txt](https://github.com/user-attachments/files/20697397/cargo_insta_test.txt)

PR #774, opened 2025-06-09
split up Partial::end into smaller bits

ThouCheese:


fasterthanlime:
This has rotted a bit:

  * There's one less lifetime in facet now
  * The SUBSTACK stuff has moved into facet-args's vendored copy of facet-(de)serialize
  * facet-(de)serialize have moved into their own repos:
    * https://github.com/facet-rs/facet-serialize
    * https://github.com/facet-rs/facet-deserialize

Do you think it's worth bringing this up-to-date @ThouCheese ?

ThouCheese:
Nope this was just an experiment to make that function a bit smaller, I think it'd be less work to give it another try

PR #780, opened 2025-06-11
Add facet-asn1 with initial DER support

RoastVeg:
Draft PR as there are a handful more types I'd like to support out-of-the-box as well as some test coverage improvements. PR reviews are welcome nonetheless.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -445,6 +445,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "facet-asn1"
+version = "0.1.0"
+dependencies = [
+ "facet",
+ "facet-core",
+ "facet-reflect",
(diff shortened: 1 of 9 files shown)


RoastVeg:
Ready at last.

I had to drop support for `const-oid` because I couldn't find a good `impl Facet` for either of `ObjectIdentifier` or `ObjectOdentifierRef`.

fasterthanlime:
Am I right in thinking we could use this to deserialize public certificates, for example?

fasterthanlime (approved):
Looks good from what I can see!

RoastVeg:
> Am I right in thinking we could use this to deserialize public certificates, for example?

Yes, among lots of other things we should be able to write structs for X509 certificates.

PR #781, opened 2025-06-12
facet-core: types: impl Display for Type

spikespaz:
These patches implement `Display` for `Type` for my upcoming work on `Debug` for `Shape`.

This is still considered "user-facing display formatting", despite that in many situations the output will be similar to `Debug`.

Complicated structures are formatted with custom syntax (sometimes valid Rust, such as function pointers, slices, arrays, etc.) surrounded by guillemet characters (`«{}»`) to indicate that accurate representation of structure has stopped at this depth.

There are no tests. Please let me know if you want me to set up snapshot tests for this code.

I am not sure that the occurrences of `unreachable!` and `assert!` are appropriate in the body of a `Display` implementation, but I figured it was okay since each instance should have been prevented by a compiler error beforehand. It may be worth adding another commit to replace these each with `debug_assert!` instead. I have avoided formatting invalid states, but that is also an option.

Some of this has raised my hackles, especially with `Repr` and `EnumRepr`. I have left comments in the code and extended descriptions on the relevant commits.

### DIFF
--- a/facet-core/src/types/ty/mod.rs
+++ b/facet-core/src/types/ty/mod.rs
@@ -41,6 +41,10 @@ pub enum Type<'shape> {
     Pointer(PointerType<'shape>),
 }
 
+// This implementation of `Display` is user-facing output, where the users are developers.
+// It is intended to show structure up to a certain depth, but for readability and brevity,
+// some complicated types have custom formatting surrounded by guillemet characters
+// (`«` and `»`) to indicate divergence from AST.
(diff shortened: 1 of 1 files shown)


spikespaz:
This failed because of coverage, is it important to add tests for `Display` stuff?

fasterthanlime:
> This failed because of coverage, is it important to add tests for `Display` stuff?

Nope, I merged it :)

PR #789, opened 2025-06-15
Use exact versions of dependencies in `facet`, closes #748

### LINKED ISSUE #748: Versioning of re-exported crates
I ran into an interesting issue related to how Facet exports its internal crates (https://github.com/mkeeter/fidget/issues/313).

Consider the following example:
```toml
# cargo.toml
[package]
name = "facet-test-2"
version = "0.1.0"
edition = "2024"

[dependencies]
facet = "=0.27.1"
```

```rust
#[derive(facet::Facet)]
struct Point {
    x: f64,
    y: f64,
}
fn main() {
    let builder = facet::Wip::alloc::<Point>();
}
```

This worked if you ran it in the past, when `facet 0.27.1` was the most recent release.  However, it does not work in the present day!

```
error[E0433]: failed to resolve: could not find `Wip` in `facet`
 --> src/main.rs:8:26
  |
8 |     let builder = facet::Wip::alloc::<Point>();
  |                          ^^^ could not find `Wip` in `facet`

For more information about this error, try `rustc --explain E0433`.
```

Previously, the `Wip` struct was re-exported from `facet-reflect v0.27.1`.  However, because the main `facet` crate specifies [`facet-reflect` as `0.27.1`](https://github.com/facet-rs/facet/blob/facet-v0.27.1/facet/Cargo.toml#L34-L36), Cargo resolves to the most recent **compatible** version, which is 0.27.12, which _does not_ have the `Wip` builder in `facet-reflect`.

This is unfortunate; it means that users have to pin the entire family of `facet` crates to a particular version, which kinda defeats the purpose of a single crate re-exporting others.

We can't fix this for historical crates, but for new releases, I think `facet` should specify its dependencies by exact version; if `facet 0.27.1` had specified `facet-reflect = { version = "=0.27.1", ...}`, then we wouldn't have this issue.

fasterthanlime:


fasterthanlime:
Let's see how this behaves when cutting a new version!

PR #792, opened 2025-06-17
facet-core: rework impl Debug for Shape

spikespaz:
Something that I'm not sure about is: there are some fields which are optionally shown, but `Some(...)` does not appear in the output to indicate that they are indeed `Option`. For the time-being, I have decided it's okay to be XXX, because if the developer needs the field they will quickly notice the `Option`-ality from the LSP's feedback or the compiler yelling at them. Still, I'm not a fan of lying in the `Debug` formatting.

There are more comments in the code and descriptions on certain commits, but here's an overview:

- `type_identifier` - First, because it tells you what type the `Shape` is for.
  - This position orients the reader immediately.
- `type_params` - Second, because it forms a complete path to the monomorphized type.
  - Distinguishes between `Shape`s with the same `type_identifier`.
  - Only shown if there are generic parameters present, otherwise omitted.
- `type_tag` - Third, as it provides optional information relevant to self-describing or introspective formats.
  - Omitted if it is `None`, shown without `Some(...)`.
- `attributes` - Fourth, because it determines how de/serialization should behave, and is that is high-value in most applications.
  - Only shown if there are attributes, otherwise omitted.
- `inner` - Fifth, because if the type is a wrapper, knowing the type that is wrapped is structurally relevant.
  - However, it's deferred slightly since it can be long, and because it builds upon the context provided by `ty` (adjacent).
  - Omitted if it is `None`, shown without `Some(...)`.
- `ty` - Sixth, even though it's arguably the most important field, it can be quite verbose.
  - Shown after metadata to avoid disrupting visual flow.
- `layout` - Seventh, because memory layout is rarely needed unless debugging ABI or alignment issues.
  - Positioning this directly after `ty` helps link them logically.
- `def` - Eighth, since it's often redundant with `ty`.
  - May want to consider selectively showing this field, depending on `inner` and `ty`.
- `doc` - Last, because it's either irrelevant noise for a summary, or the only thing you care about.
  - In the latter case, it's easy to spot at the bottom.
  - Omitted if there is no documentation.
- `id` - Omitted because it's not very useful.
- `vtable` - Omitted because raw function pointers probably aren't helpful outside of a low-level debugging context.

### DIFF
--- a/facet-core/src/types/mod.rs
+++ b/facet-core/src/types/mod.rs
@@ -406,18 +406,33 @@ impl core::fmt::Debug for Shape<'_> {
         } else {
             let mut debug_struct = f.debug_struct("Shape");
 
-            // Always show the type name
-            debug_struct.field("type", &format_args!("{}", self));
-
-            // Show def if it's not Undefined
-            if !matches!(self.def, Def::Undefined) {
-                debug_struct.field("def", &format_args!("{:?}", self.def));
+            macro_rules! field {
+                ( $field:literal, $( $fmt_args:tt )* ) => {{
+                    debug_struct.field($field, &format_args!($($fmt_args)*));
+                }};
             }
 
-            // Show inner if present
-            if self.inner.is_some() {
-                debug_struct.field("inner", &format_args!("Some(..)"));
-            }
+            field!("id", "{:?}", self.id);
+
+            field!("layout", "{:?}", self.layout);
+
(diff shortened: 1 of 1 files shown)


fasterthanlime:
I read through the rationale and it makes sense to me! Thanks for tackling this.

Re def being redundant with ty, the plan is to get rid of it altogether rather than just in the Debug impl.

Issue #339, opened 2025-04-20
facet-assert


fasterthanlime:
Let's provide facet-assert, like pretty-assertions (and with diffs, colors, etc.) but better!

fasterthanlime:
Been done a while ago! We have `facet-diff`, and `facet-assert`

PR #807, opened 2025-06-22
Allowing to inspect facet objects (get field by path + object diff)

pngouembe:
Hi !

This is an experimentation on allowing facet objects to be diffed.

I make this PR early to gather feedback and make sure that I go the right way.

The key elements of this PR are:
- Introducing Paths allowing to navigate a facet object
- Creating an iterator that will traverse a given facet object and yield a (path, peek) tuple
- using this iterator to create a diff between two facet objects

I have two things in mind with this addition, allowing facet_assert (https://github.com/facet-rs/facet/issues/339) and allowing sending diffs over the network and apply them like patches to existing objects.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -491,6 +491,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "facet-inspect"
+version = "0.1.0"
+dependencies = [
+ "facet",
+ "facet-reflect",
+]
+
+[[package]]
 name = "facet-json"
 version = "0.24.15"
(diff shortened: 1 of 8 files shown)


fasterthanlime:
Looks like a nice start!

fasterthanlime:
I'm happy to merge this as-is, just to have the crate in there — then you can iterate on it as much as you want. Sounds good?

pngouembe:
Sounds nice @fasterthanlime 😊

PR #816, opened 2025-06-24
Reflection perf tinkering

simonask:
Here's my results of tinkering with reflection benchmarks and profilers.

The challenge is hard because it's difficult to know what actually matters, especially for a broad API with many potential use cases. It's super hard to get really conclusive, but these changes are quite conservative and should fall in the "well at least it can't hurt" category.

I've attacked the problem from a couple of angles:

1. Profiling existing benchmarks, especially `json`.
2. Profiling a new microbenchmark, `reflect`.
3. Manually looking at the disassembly for a couple of probably-hot functions to make educated guesses about performance impact.

I've been using Visual Studio and AMD uProf to measure and analyze the results.

## Findings / Guidance

**Key takeaway:** Facet is pretty fast, and there's not a huge amount of low-hanging fruit left. Big improvements mostly have to come from refactoring or redesign of internal structures, and sometimes whole APIs. There is very probably some room for improvement, especially on the happy path, and I imagine that it would mostly look like lifting validation checks out of the way.

### Enable LTO: ~20% improvement for free

I think everyone who cares about performance will be enabling LTO no matter what, and it makes a big difference. The `Deserialize - Nested (depth=15)` benchmark went from consistently ~1.3ms to consistently ~1.1ms with this change on my machine. This PR enables LTO for release builds in the monorepo, which just affects the default baseline for all future benchmarks.

### If you're allocating 1, why not allocate 2?

The stack in `Partial` is always allocated with 1 element, but every type that is not a primitive has a minimum "depth" of at least 2 frames, so you are basically guaranteed to get an immediate reallocation. Meanwhile, there is generally no difference in allocator performance between allocating 64 bytes or 128 bytes, or even 256 bytes.

This PR changes `Partial::alloc_shape()` to preallocate 4 frames instead of 1, cutting single-level struct deserialization time in half (because they are dominated by the allocation). Even for deeper structs, this change should still prevent at least 2 reallocations, because the reallocation policy of `Vec` is to double its capacity, so we get to amortized territory much quicker.

### Minimize indirect calls

LLVM can't see through function pointers in most cases, especially the `fn() -> TypeId` inside of `ConstTypeId`. It can't tell that the function always returns the same value, so a long if-else chain of `id == ConstTypeId::of::<...>()` will result in repeated indirect calls, and will prevent other optimization passes, such as jump tables.

This PR makes `ConstTypeId::get()` public, and changes `ScalarType::try_from_shape()` to call the `fn() -> TypeId` once and then compare against the raw `TypeId` instead, yielding much better codegen - particularly when inlined.

### Inline small functions

This does basically nothing when LTO is enabled, but I have conservatively added `#[inline]` to many obvious candidates. These are small non-generic functions that typically just delegate to another function, potentially through a function pointer. You can say that it doesn't really matter, since LTO will be enabled for everyone who cares, but I do think there are workflows where you semi-care about performance for fast compile times during development (like `--opt-level=1` for the main project and `--opt-level=3` for deps).

I've not had any luck getting anything out of inlining bigger functions like `Partial::end()` or `Partial::begin_nth_field()` - these can be made faster, but not without changing how they work.

### Struct size matters a little bit

Did you know that LLVM sometimes generates `imul` instructions for random-access `slice[index]` lookups? ([godbolt](https://godbolt.org/z/5n3Yobsqj)) It happens when `size_of::<Foo>() >= 56`, and when `size_of::<Foo>() / 8` is uneven after that. We're well into micro-optimization territory here, but I found it interesting.

The `Field` struct is currently 88 bytes, and there are quite a few index-based lookups on those in very hot paths. I would recommend splitting the struct into a hot and a cold part, where the cold part is accessible through a pointer in the hot part. Cold fields would probably be `doc`, `attributes`, `flattened`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,3 +44,7 @@ edition = "2024"
 rust-version = "1.87.0"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/facet-rs/facet"
+
+[profile.release]
+lto = true
+debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,3 @@ edition = "2024"
 rust-version = "1.87.0"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/facet-rs/facet"
-
-[profile.release]
-lto = true
-debug = true
--- a/facet-bench/Cargo.toml
+++ b/facet-bench/Cargo.toml
@@ -26,3 +26,7 @@ harness = false
 [[bench]]
 name = "canada"
 harness = false
+
+[[bench]]
+name = "reflect"
+harness = false
--- a/facet-bench/benches/reflect.rs
+++ b/facet-bench/benches/reflect.rs
@@ -1,3 +1,5 @@
+#![expect(clippy::approx_constant)]
+
 use divan::{Bencher, black_box};
 use facet::Facet;
 use facet_reflect::{Partial, ReflectError};
--- a/facet-bench/benches/reflect.rs
+++ b/facet-bench/benches/reflect.rs
@@ -1,0 +1,50 @@
+use divan::{Bencher, black_box};
+use facet::Facet;
+use facet_reflect::{Partial, ReflectError};
+
+#[derive(Facet)]
+struct SimpleStruct {
+    a: u32,
+    b: i32,
+    c: f32,
+}
+
+#[divan::bench(name = "Populate small struct by field names")]
+fn populate_simple_struct(bencher: Bencher) {
+    fn populate_by_field_name<'shape>(
+        partial: &mut Partial<'_, 'shape>,
+    ) -> Result<(), ReflectError<'shape>> {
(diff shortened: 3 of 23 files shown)


simonask:
This PR should be per-commit reviewable, and if there are any changes you are unhappy about, I can easily subtract them.

simonask:
Disabling LTO from benchmarks fixes the regression in `Serialize - Long String (100KB) - serde`, and still yields a lot of improvement in Facet's own benchmarks - though not nearly as much. Probably some inlining heuristics in `serde_json` are being affected, but who can say.

I would personally choose to always benchmark with LTO, because that's what people enable when they care about performance, but it also muddies the picture a bit, so I can go either way. :-)

fasterthanlime:
In this specific case, where facet is geared towards folks who care about compile times a lot and are willing to trade perf for it, I think benchmarking with lto off is the right move! Happy to see there are still gains from everything else!

PR #818, opened 2025-06-25
Fix `#[facet(default = ...)]` in facet-toml

tristan:
Noticed the toml deserializer wasn't respecting `#[facet(default = ...)]`.

This fixes it by re-ordering the if-else chain to check for `default` before checking if the type is `Option<T>` and also calling `set_field_default` instead of `set_default` when `field.vtable.default_fn` is present.


tversteeg:
Thanks for the PR, LGTM!

fasterthanlime:
Thanks for this!

PR #819, opened 2025-06-26
POC Bugfix Hashmap with numbers to valid json

VanOvermeire:
This is a POC for fixing the issue of `HashMap<u32, u32>` not serializing to valid JSON.

The issue is that JSON keys should be strings, and the current code simple serializes as-is, i.e. `{ 1: 2 }`. Serde on the other hand outputs `{ "1": 2 }`.

In the below code, this is fixed for `i32` by checking the object state. If the last element on the stack is `ObjectItemState::Value`, we're dealing with a number key in an `HashMap` (or similar). So we change the number to a string.

Currently, I've only changed this one method because there might be a better place to fix this? Not sure.

If this is a good place to fix the issue, the PR obviously has to be extended to also transform other number keys (and booleans?) to strings as well.


fasterthanlime (commented):
Well, this feels a bit wrong but I can't think of a better way right now so.

VanOvermeire:
@fasterthanlime do you want me to create another PR to apply the same solution to other numbers? Current PR only covers `i32`

fasterthanlime:
> @fasterthanlime do you want me to create another PR to apply the same solution to other numbers? Current PR only covers `i32`

Mhh sure, let's make sure there's shared code there, like a common function to check that we need to write a string, not a number — I actually am not super fond of `serialize_i32` etc. — I'd rather have an enum, then there would be a single place for us to have a "needs_string_coerce" check. But that's probably too much code, for another PR.

VanOvermeire:
Yes, a common function sounds good - for now. :)

VanOvermeire:
Follow up PR: https://github.com/facet-rs/facet/pull/821

Issue #145, opened 2025-04-11
Use case idea: diffing values


whentze:
Could we use Facet to diff values of the same type, getting a representation of the diff as a value (that is itself `Facet`?) and that can be applied to other values of the type?

Prior art:
- https://crates.io/crates/daft
- https://crates.io/crates/serde-diff

I have looked at Facet for all of like 10 minutes, but I think this could be built on top of just peek and poke. Alternatively, it might even be possible to provide a blanket `Diffable` impl for all `Facet` types so people can use `daft` as is but save themselves an extra derive on every type.

fasterthanlime:
Yep absolutely — I was thinking of this for facet-pretty but it probably should be its own package.

whentze:
Agreed on this being a separate package. `facet-diff` or something. facet-pretty doesn't quite fit because the diff (at least the kind I'm interested in) would not be textual but rather a Rust data structure. And if  `Diff<T> : Facet`, you could then use facet-pretty to pretty-print the diff if you wanted to.

FilipAndersson245:
This could probably be combined into an awesome facet-assert (#339)

pngouembe:
Hello,

I tried to think about this and I think I need some insight from you as I don't really know if what i came out with is relevant ^^'.

I thought about introducing some kind of path allowing to navigate a facet structure.

then the diff structure would contain 2 collections, the structural differences and the data differences.

Let's consider that we do `a.diff(b)`:

- the structural differences would contain the list of paths to add to/delete from `a`'s shape to obtain `b`'s shape.
- the data differences would contain the paths to the data that changed between `a` and `b` associated with the before value (`a`'s) and the after value (`b`'s)

a little example of what I'm thinking about:

```rust
let a = facet_json::from_str::<AStruct>(r#"{"a": 42, "b": "foo"}"#);
let b = facet_json::from_str::<BStruct>(r#"{"a": 40, "b": "foo", "c": "bar"}"#);

let diff = facet_json::to_string(a.diff(&b));

assert_eq!(
    diff,
    r#"{
        types: ["AStruct", "BStruct"],
        "shape_modification": [
            {"this.c": "+"}
        ], 
        "data_modification": [
            {"this.a": {"before": 42, "after": 40}}
        ]
    }"#
)
```

with this kind of structure, we could implement the pretty print on the diff struct that would allow to do the `facet_assert`.

[...]

PR #824, opened 2025-06-30
Add facet-diff

TheLazyDutchman:
This PR goes most of the way to adding diffing capabilities to facet types as discussed in [this issue](https://github.com/facet-rs/facet/issues/145).

There are still features that would be nice to add, but I think it would be best to get this feature out there so that others can take a look at it.


TheLazyDutchman:
I do not know how to fix this lockfile test, which should also fix the msrv test.

If people want me to add tests to `facet-diff`, I can do that, but I consider the `Diff` type and the `Display` output to be quite unstable still.

fasterthanlime:
> I do not know how to fix this lockfile test, which should also fix the msrv test.

I'll take a look :)

fasterthanlime:
Done in #826

PR #833, opened 2025-07-01
Remove args-specific stuff from deserialize

fasterthanlime:


fasterthanlime:
The regression here is fishy... can't explain why that would be the case.

fasterthanlime:
The regression disappears when enabling LTO, and looking at the benchmarks makes me suspect it's all about inlining decisions.

I'm enabling LTO for benchmarks to try to avoid wasting time on these regressions — hopefully performance will be more stable then.

PR #840, opened 2025-07-05
Fix XDR serialization of u8, u16, i8, and i16

### LINKED ISSUE #839: facet-xdr serializes integers below 32-bit as 64-bit
I suspect this is because the default implementations of `serialize_(u|i)(8|16)` in facet-serialize defer to `serialize_(u|i)64` instead of `serialize_(u|i)32` and they are not overwritten in `impl Serializer for XdrSerializer`. This would suggest two ways of fixing it, either override them for `XdrSerializer` or change the default implementations to defer upward through sizes (`(u|i)8` -> `(u|i)16` -> `(u|i)32` -> `(u|i)64`).

<details><summary>Test Program</summary>

```rust
#[derive(facet::Facet, Debug)]
pub struct Foo {
    pub a: u16,
    pub b: u16,
}

fn main() {
    let msg = Foo {
        a: 123,
        b: 45,
    };
    println!("msg: {:?}", msg); // msg: Foo { a: 123, b: 45 }
    let bytes = facet_xdr::to_vec(&msg).unwrap();
    println!("bytes: {:?}", bytes); // bytes: [0, 0, 0, 0, 0, 0, 0, 123, 0, 0, 0, 0, 0, 0, 0, 45]
    let msg2 = facet_xdr::deserialize::<Foo>(&bytes).unwrap();
    println!("msg2: {:?}", msg2); // msg2: Foo { a: 0, b: 123 }
}
```
</details>

RoastVeg:
Fixes #839

### DIFF
--- a/facet-xdr/CHANGELOG.md
+++ b/facet-xdr/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fixed serialization of `u8`, `u16`, `i8`, and `i16`
+
 ## [0.1.19](https://github.com/facet-rs/facet/compare/facet-xdr-v0.1.18...facet-xdr-v0.1.19) - 2025-07-03
 
 ### Other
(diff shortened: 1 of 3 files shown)


fasterthanlime:
Cherry-picked on top of main.

PR #842, opened 2025-07-07
Fix serialization for Option<T> in XDR

### LINKED ISSUE #841: facet-xdr Optional Serialization missing leading bytes
In XDR, a `Some` optional should be serialized with `\0\0\0\1`, followed by the data. Facet handles this while decoding ([facet-xdr/src/lib.rs#L491](https://github.com/facet-rs/facet/blob/1ae8c979feca4ec655fa6e0f8e9e4cf4d2180072/facet-xdr/src/lib.rs#L491)), however when serializing the leading bytes are omitted.

The easiest test case is that [the spec](https://datatracker.ietf.org/doc/html/rfc1832#section-3.19) says that `type-name *identifier;` (`Optional<T>`) is equivalent to `type-name identifier<1>;` (`Vec<T>`), but facet serializes them differently.

```rust
#[derive(Debug, Clone, Copy, facet::Facet)]
struct Foo {
    elem: u32,
}

fn main() {
    let inner = Foo { elem: 57 };

    let option_serialization = facet_xdr::to_vec(&Some(inner)).unwrap();
    let vec_serialization = facet_xdr::to_vec(&vec![inner]).unwrap();
    assert_eq!(
        option_serialization, vec_serialization,
        "Option and Vec Serialization are not identical.",
    );
}

```
```
thread 'main' panicked at src/main.rs:11:5:
assertion `left == right` failed: Option and Vec Serialization are not identical.
  left: [0, 0, 0, 57]
 right: [0, 0, 0, 1, 0, 0, 0, 57]
****
```

RoastVeg:
Fixes #841

### DIFF
--- a/facet-serialize/CHANGELOG.md
+++ b/facet-serialize/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add start_some to serializer trait
+
 ## [0.24.18](https://github.com/facet-rs/facet/compare/facet-serialize-v0.24.17...facet-serialize-v0.24.18) - 2025-07-03
 
 ### Other
--- a/facet-serialize/src/lib.rs
+++ b/facet-serialize/src/lib.rs
@@ -61,6 +61,11 @@ pub trait Serializer<'shape> {
     /// Serialize a `None` variant of an Option type.
     fn serialize_none(&mut self) -> Result<(), Self::Error>;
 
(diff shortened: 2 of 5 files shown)


facet-serialize/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add start_some to serializer trait
```
tversteeg, line 10: The changelogs are automatically generated from commit messages, so all manual changes will be overwritten.

PR #845, opened 2025-07-07
Allow facet-json to build with no-default-features

itsjunetime:
`facet-json` doesn't currently build when `default-features = false` because that disables the `std` features of all the underlying crates, and `no-std`/`no-alloc` doesn't necessarily compile in all of the facet crates atm. 

It's also just not reasonable to expect a general-purpose json serializer/deserializer to work out-of-the-box without std at the moment, so I think this is the correct solution to this problem.

### DIFF
--- a/facet-json/Cargo.toml
+++ b/facet-json/Cargo.toml
@@ -15,9 +15,9 @@ default = ["rich-diagnostics"]
 
 [dependencies]
 facet-core = { version = "0.27.16", path = "../facet-core", default-features = false }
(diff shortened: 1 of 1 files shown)


fasterthanlime:
> It's also just not reasonable to expect a general-purpose json serializer/deserializer to work out-of-the-box without std

I think it is, with just alloc?

fasterthanlime:
Done in https://github.com/facet-rs/facet-json/commit/8cd6b74bcc55cce4bf19267104a1b1cc79d3017a

PR #896, opened 2025-10-18
add ndarray def kind

sarah-quinones:
adds an n-dimensional array type description.

n-dimensional arrays have more or less the same api as lists, but their shape cannot be mutated, so there are no `push`, `init_with_capacity`, etc functions in the vtable.

unlike arrays, ndarrays are rarely contiguous, so they also expose a strided access api to help with the common case

### DIFF
--- a/facet-core/src/types/def/mod.rs
+++ b/facet-core/src/types/def/mod.rs
@@ -35,6 +35,7 @@ pub use ndarray::*;
 #[repr(C)]
 // this enum is only ever going to be owned in static space,
 // right?
+#[non_exhaustive]
 pub enum Def {
     /// Undefined - you can interact with the type through [`Type`] and [`ValueVTable`].
     Undefined,


facet-core/src/types/def/mod.rs
```diff
@@ -62,6 +65,11 @@ pub enum Def {
     /// e.g. `[T; 3]`
     Array(ArrayDef),
 
+    /// n-dimensional array of heterogeneous values, fixed size
```
fasterthanlime: did you mean homogeneous? or did I misunderstand what you're going for

sarah-quinones: oh wait, yeah. i always confuse the two

fasterthanlime (approved):
I guess the only bummer here is that an ndarray cannot be constructed with this vtable, but it can always be added later 🤷

PR #900, opened 2025-10-21
remove Facet::VTABLE, add num-complex support, add type checks

sarah-quinones:
the type checks are for `Peek::partial_eq` and similar functions to make sure that we're not passing any random type to the vtable functions.

### DIFF
--- a/facet-reflect/src/peek/value.rs
+++ b/facet-reflect/src/peek/value.rs
@@ -103,9 +103,14 @@ impl<'mem, 'facet> Peek<'mem, 'facet> {
     /// `false` if equality comparison is not supported for this scalar type
     #[inline]
     pub fn partial_eq(&self, other: &Peek<'_, '_>) -> Result<bool, ReflectError> {
-        if self.shape == other.shape {
-            if let Some(f) = (self.vtable().partial_eq)() {
+        if let Some(f) = (self.vtable().partial_eq)() {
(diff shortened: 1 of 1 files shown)


fasterthanlime:
CleanShot 2025-10-21 at 11 22 47@2x

halp

facet-core/src/spez/mod.rs
```diff
@@ -32,11 +34,24 @@ pub struct Spez<T>(pub T);
...
 
-impl<T> SpezEmpty<T> {
+impl<T: ?Sized> SpezEmpty<T> {
     /// An instance of [`SpezEmpty`].
     pub const SPEZ: Self = Self(PhantomData);
 }
 
+impl<T> SpezEmpty<T> {
+    /// Sized shape builder of `T`
+    pub const BUILDER: ShapeBuilder = Shape::builder_for_sized::<T>();
```
fasterthanlime, line 44: ooh that's evil I like it

facet-core/src/ptr.rs
```diff
@@ -163,18 +163,25 @@ struct Ptr {
     metadata: *const (),
 }
 
+/// the layout of pointers to DST is not guaranteed so we try to detect it in a const-friendly way
```
fasterthanlime, line 166: thanks for the extra comments!

PR #915, opened 2025-11-19
allow `#[facet(opaque)]` attribute for container

nikvoid:
This PR allows using `#[facet(opaque)]` attribute on container. It does the following things:
1. Overrides `ty` with `Type::User(UserType::Opaque)`
2. Removes `Facet` bound from generic parameters
3. Removes generic parameters from type name

This allows using Facet for e.g. following use-case:
```rust
#[derive(Facet)]
#[facet(opaque)]
pub struct CommaSep<T: Display + FromStr>(Vec<T>);
impl <T: Display + FromStr> Display for CommaSep<T> { ... }
impl <T: Display + FromStr> FromStr for CommaSep<T> { ... }
```

Where `T` cannot implement `Facet`, but may be parsed/formatted through wrapper type, maybe with additional custom behaviour, like in this case (comma-separated list of values).

### DIFF
--- a/facet-macros-emit/src/process_struct.rs
+++ b/facet-macros-emit/src/process_struct.rs
@@ -198,7 +198,11 @@ pub(crate) fn process_struct(parsed: Struct) -> TokenStream {
     let struct_name = &ps.container.name;
     let struct_name_str = struct_name.to_string();
 
-    let opaque = ps.container.attrs.facet.iter()
+    let opaque = ps
+        .container
+        .attrs
+        .facet
+        .iter()
         .any(|a| matches!(a, PFacetAttr::Opaque));
 
     let type_name_fn = generate_type_name_fn(struct_name, parsed.generics.as_ref(), opaque);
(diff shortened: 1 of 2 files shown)


fasterthanlime:
that's an interesting addition. I have not looked at the implementation yet. I will when the tests are passing!

nikvoid:
Test fails are caused by moved variable definitions. How do I regenerate codegen snapshots?

fasterthanlime:
> How do I regenerate codegen snapshots?

`cargo insta test`, then `cargo insta review` — https://lib.rs/crates/cargo-insta

facet-macros-emit/src/process_struct.rs
```diff
@@ -502,6 +511,8 @@ pub(crate) fn process_struct(parsed: Struct) -> TokenStream {
         #[automatically_derived]
         unsafe impl #bgp_def ::facet::Facet<'__facet> for #struct_name_ident #bgp_without_bounds #where_clauses {
             const SHAPE: &'static ::facet::Shape = &const {
+                #fields
```
fasterthanlime, line 514: I'm struggling to find the semantic difference in this commit. Can you help me out? Why does `fields` need to be a separate binding now?

nikvoid: In previous commit I integrated it into .ty(...) method call, but formatter struggled with it and gave up. In result there were unreadable long lines of code

nikvoid: So I this difference is only, err, cosmetic? Instead of changing codegen snapshots I changed code generation to be similar to what was before.

fasterthanlime: Oh, that's a completely valid reason to do that. Thank you for the explanation. The reason it was that way in the first place is probably also because of formatting issues. Rust fmt gets very upset very easily.

fasterthanlime:
This shipped in facet v0.31.4

Issue #902, opened 2025-10-22
Add support for (de)serialization of non-facet types when used as fields of facet types.


tristan:
Non-Facet types can be included as fields of Facet types using `#[facet(opaque)]`, but this blocks the type from being able to be reflected on completely, blocking them from (de)serialization.

Being able to provide some way to tell the reflection/(de)serialization machinery how to convert to/from an opaque type would help cover these cases.

Related issues:
- https://github.com/facet-rs/facet/issues/731
- https://github.com/facet-rs/facet/issues/779

tristan:
I have implemented a proof of concept adding a `#[facet(proxy(ProxyType))]` attribute for struct fields here https://github.com/tristan/facet/pull/1.

The basic idea is that the (de)serialization (peek/poke) things see the shape of the field as the shape of the `ProxyType` (which has to implement Facet), and the reflection side of things covers conversion between that and the real type.

This allows using non-facet types via a Newtype that has a (de)serializable shape, and has `TryFrom` methods to convert between the target and proxy types.

e.g.

```
#[derive(facet::Facet, Debug)]
#[facet(transparent)]
pub struct HexString(pub String);

impl TryFrom<HexString> for ::some_external_lib::BigInt {
    type Error = ParseIntError;

    fn try_from(value: HexString) -> Result<Self, Self::Error> {
        value.0.parse()
    }
}

impl From<&::some_external_lib::BigInt> for HexString {
    fn from(value: &::some_external_lib::BigInt) -> Self {
        HexString(format!("{value:0x}"))
    }
}

#[derive(facet::Facet, Debug)]
pub struct Test {
    #[facet(proxy(HexString))]
    pub data: ::some_external_lib::BigInt,
}

let data = r#"{"data":"0xabc"}"#;
let test: Test = facet_json::from_str(data).unwrap();
let d: u64 = test.data.as_u64();
assert_eq!(d, 2748);
let serialized_data = facet_json::to_string(&test);
assert_eq!(serialized_data, data);
```

While this "works", the implementation is far from ideal (serialization is especially not in a happy place) and I'm not confident it's the right track for supporting this.

tristan:
Resolved by https://github.com/facet-rs/facet/pull/924

PR #916, opened 2025-11-21
Add serialize_with and deserialize_with field attributes

tristan:
Closes: https://github.com/facet-rs/facet/issues/902

companions: 
 - https://github.com/facet-rs/facet-serialize/pull/10
 - https://github.com/tristan/facet-deserialize/pull/1
 - https://github.com/tristan/facet-json/pull/1 (only tests!)

Implements the `serialize_with` and `deserialize_with` attributes for fields to allow users to provide functions to do custom (de)serialization of types similar to serde's `(de)serialize_with` methods.

## `serialize_with`

allows a separate shape to be parsed from the input by facet, then transformed to the target type using the `deserialize_with` function provided.

the function takes the form: `Fn(&SourceType) -> Result<TargetShape, ErrorType>`.

The `SourceType` does not have to implement `Facet`, and thus can be used to serialize structs and enums containing types that are marked `opaque`.

The `ErrorType` can be anything that implements Display (i.e. can be passed to `format!("{}", e)`).

e.g.

```rust
#[derive(Facet)]
struct MyStruct {
    #[facet(opaque, serialize_with = str_from_opaque)]
    field: OpaqueType
}

fn str_from_opaque(&String) -> Result<OpaqueType, &'static str> { ... }
```

or 

```rust
#[derive(Facet)]
enum MyEnum {
    TupleVariant(
        #[facet(opaque, serialize_with = str_from_opaque)]
        OpaqueType
    )
    FieldVariant {
        #[facet(opaque, serialize_with = str_from_opaque)]
        field: OpaqueType
    }
}
```

This does not have to be used with opaque types and can be used to provide custom serialization of other shapes

e.g.

```rust
#[derive(Facet)]
struct MyStruct {
    #[facet(serialize_with = into_hex_str)]
    val: u64
}

fn from_hex_str(val: &u64) -> Result<String, &'static str> {
    format!("{val:x}")
}
```

Using the custom serialization is up to the serializer, and can otherwise be ignored for non-serialization cases. e.g. see https://github.com/facet-rs/facet-serialize/pull/10 for the changes to `facet-serialize` to support this. 

## `deserialize_with`

allows a separate shape to be parsed from the input by facet, then transformed to the target type using the `deserialize_with` function provided.

the function takes the form: `Fn(&SourceShape) -> Result<TargetType, ErrorType>`.

The `TargetType` does not have to implement `Facet`, and thus can be used to deserialize structs and enums containing types that are marked `opaque`.

The `ErrorType` can be anything that implements `Display` (i.e. can be passed to `format!("{}", e)`).

e.g.

```rust
#[derive(Facet)]
struct MyStruct {
    #[facet(opaque, deserialize_with = opaque_from_str)]
    field: OpaqueType
}

fn opaque_from_str(&String) -> Result<OpaqueType, &'static str> { ... }
```

or 

```rust
#[derive(Facet)]
enum MyEnum {
    TupleVariant(
        #[facet(opaque, deserialize_with = opaque_from_str)]
        OpaqueType
    )
    FieldVariant {
        #[facet(opaque, deserialize_with = opaque_from_str)]
        field: OpaqueType
    }
}
```

This does not have to be used with opaque types and can be used to provide custom deserialization of other shapes

e.g.

```rust
#[derive(Facet)]
struct MyStruct {
    #[facet(deserialize_with = from_hex_str)]
    val: u64
}

fn from_hex_str(hex: &String) -> Result<u64, std::num::ParseIntError> {
    u64::from_str_radix(hex, 16)
}
```

When used with the facet-json stack, errors will show up like so:
```
thread 'main' panicked at src/main.rs:91:52:
called `Result::unwrap()` on an `Err` value: Error:
   ╭─[ <unknown>:1:26 ]
   │
 1 │ {"OpStrField": {"field": "0xhello"}}
   │                          ────┬────
   │                              ╰────── Custom deserialization of shape 'u64' into 'Opaque' failed: invalid digit found in string
───╯
```

Using the custom deserialization is up to the deserializer, and can otherwise be ignored for non-deserialization cases. e.g. see https://github.com/tristan/facet-deserialize/pull/1 for the changes to `facet-deserialize` to support this. 

## `#[cfg(feature = "alloc")]`

Since using this feature requires an allocator to construct the temporary shapes used during the custom (de)serialization, I decided it was easiest to put the behaviour and all associated parts behind the `alloc` feature. This was also to allow use of `Shape::allocate` and `Shape::deallocate_mut` which are already gated behind the `alloc` feature.

Unfortunately the facet-macros crates don't currently have an alloc, so i left those free for now. They might cause issues for `no_std` people using libraries that use `(de)serialize_with`, so might be worth adding a `cfg` fence to those as well in the future.

## Error handling

I started out with using only `&'static str` for the `message` field of the `ReflectError` variants to avoid adding anything that needed allocation, but since the `#[cfg(feature = "alloc")]` detail means we always have an allocator when using these features, I decided to go with `String` to allow the `(de)serialize_with` methods to return anything that implements `Display` which can be formatted into a `String` (e.g. all `std::error::Error`s). I considered using `Box<dyn Something>`, but `ReflectError` is currently `Clone`, and cloning a `Box<dyn ...>` is already a [fun adventure](https://quinedot.github.io/rust-learning/dyn-trait-clone.html), and also restricts the error type returned by the `(de)serialize_with` methods to be `Clone` as well, which is not a requirement of `std::error::Error`, so it would add extra restrictions to implementers for very little gain (we can't do anything with the `Box<dyn ...>` except format it anyways).

## `OwnedPeek`

the custom serialization process needed somewhere to allocate the data for the shape that is being used for serialization (i.e. the `serialize_with` result). `OwnedPeek` was created for this purpose, whose main purpose is to own the data and drop/deallocate the data when it's dropped. To make it work with existing code in `facet-serialize` the `as_peek` method was added to get a borrowed peek from this. The only difficulty is keeping the iterative style of `serialize_iterative`, as it relies on the caller to own the data. The simple solution to this was to break the iterative style when custom serialization occurs and recurse in that case (see https://github.com/facet-rs/facet-serialize/pull/10). This doesn't feel quite right, but I don't have a better solution yet (without removing the guardrails, which I think is a bad idea).

### DIFF
--- a/facet-reflect/tests/partial/deserialize.rs
+++ b/facet-reflect/tests/partial/deserialize.rs
@@ -237,7 +237,7 @@ fn wip_custom_deserialize_errors() -> Result<(), IPanic> {
         assert_eq!(src_shape, u64::SHAPE);
         assert_eq!(dst_shape, Opaque::<NotDerivingFacet>::SHAPE);
     } else {
-        assert!(false, "expected custom deserialization error");
+        panic!("expected custom deserialization error");
     }
 
     Ok(())
@@ -301,7 +301,7 @@ fn wip_custom_deserialize_zst() -> Result<(), IPanic> {
         assert_eq!(src_shape, u64::SHAPE);
         assert_eq!(dst_shape, <() as Facet>::SHAPE);
     } else {
-        assert!(false, "expected custom deserialization error");
+        panic!("expected custom deserialization error");
     }
 
     Ok(())
--- a/facet-reflect/tests/peek/serialize.rs
+++ b/facet-reflect/tests/peek/serialize.rs
@@ -33,7 +33,7 @@ fn peek_opaque_custom_serialize() -> Result<(), IPanic> {
         .expect("Should have an inner field");
 
     let mut tested = false;
-    for (field, peek) in peek_struct.fields_for_serialize() {
+    if let Some((field, peek)) = peek_struct.fields_for_serialize().next() {
         tested = true;
         assert_eq!(inner_field, peek);
         assert!(field.vtable.serialize_with.is_some());
@@ -88,7 +88,7 @@ fn peek_shaped_custom_serialize() -> Result<(), IPanic> {
         .expect("Should have an inner field");
 
     let mut tested = false;
-    for (field, peek) in peek_struct.fields_for_serialize() {
+    if let Some((field, peek)) = peek_struct.fields_for_serialize().next() {
         tested = true;
         assert_eq!(inner_field, peek);
         assert!(field.vtable.serialize_with.is_some());
@@ -142,7 +142,7 @@ fn peek_opaque_custom_serialize_enum_tuple() -> Result<(), IPanic> {
         .expect("Should have an field");
 
     let mut tested = false;
-    for (field, peek) in peek_enum.fields_for_serialize() {
+    if let Some((field, peek)) = peek_enum.fields_for_serialize().next() {
         tested = true;
         assert_eq!(inner_field, peek);
         assert!(field.vtable.serialize_with.is_some());
@@ -201,7 +201,7 @@ fn peek_opaque_custom_serialize_enum_feels() -> Result<(), IPanic> {
         .expect("Should have an field");
 
     let mut tested = false;
-    for (field, peek) in peek_enum.fields_for_serialize() {
+    if let Some((field, peek)) = peek_enum.fields_for_serialize().next() {
         tested = true;
         assert_eq!(inner_field, peek);
         assert!(field.vtable.serialize_with.is_some());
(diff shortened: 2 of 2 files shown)


fasterthanlime:
I'm going to add a few tests at the facet macros level and then get this merged.

fasterthanlime:
As is my habit now, I'm going to squash and clean up by adding a few more tests and documentation, etc. in another branch and then get this merged. And you're going to retain full credit because of the magic of Git rebase!

Thank you so much for working through this and including feedback that you got through Discord.

I think a lot of people are going to be happy about this. I'm going to follow up on the facet-{serialize,deserialize,json} PRs as well and get this all published.

fasterthanlime:
Closing in favor of https://github.com/facet-rs/facet/pull/924

PR #918, opened 2025-11-24
fix double colon in where clause

cramt:
couldnt get the git hook to work on my nix setup so i hope i didnt need to for
this pr


fasterthanlime (approved):
Well that fix was a lot easier than I thought it would be.

PR #926, opened 2025-11-25
feat(core): add ruint implementations

### LINKED ISSUE #921: Feature: consider implementing the Facet trait for `ruint`
The [`ruint`](https://crates.io/crates/ruint) crate provides bits and uint types of arbitrary size using const generics.

I am trying to derive `Facet` for a struct which contains one of those `Uint` and since the struct is itself code-generated by a macro I don't control, I can't wrap it in a newtype.

I was wondering if it could be thinkable to derive the trait for the `ruint` types here? Or should this be raised with the `ruint` authors instead?

Thanks!

beeb:
This PR implements the `Facet` trait for both of `ruint`'s types: `Bits<BITS, LIMBS>` and `Uint<BITS, LIMBS>`.

Since I am a beginner with facet, any feedback is of course welcome. I am also not very sure of how to test these.

`Uint` has a single field `limbs` which is a `[u64; LIMBS]`, and `Bits` is a newtype tuple struct wrapper around `Uint`.

Closes #921

### DIFF
--- a/facet-core/src/impls_ruint.rs
+++ b/facet-core/src/impls_ruint.rs
@@ -42,7 +42,7 @@ mod tests {
         assert_eq!(SHAPE.type_identifier, "Uint");
         assert!(matches!(SHAPE.layout, ShapeLayout::Sized(..)));
         let layout = SHAPE.layout.sized_layout().unwrap();
-        assert_eq!(layout.size(), 32);
+        assert_eq!(layout.size(), 32); // 4 limbs with type u64 -> 32 bytes
         assert_eq!(layout.align(), 8);
     }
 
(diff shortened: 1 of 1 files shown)


fasterthanlime:
I would like to see some tests, ideally? like asserting basic things about some of the ruint types. I don't know.

beeb:
Thanks for the feedback, I added some tests, let me know if it's enough!

fasterthanlime:
Update, I am happy to see some tests.

I know it can be hard to test facet implementations because the serialization and deserialization code, one of the primary use cases is in a separate repository, but: all we can do our best. 

What I try to do locally is have some tests in, say, facet-json and use Cargo's patch.crates-io feature to pick up my local version of `facet-core`.

Because I'm busy releasing Facet today, I'm going to include that. And if it turns out that it's not useful enough, we can always make a breaking change later.

Issue #931, opened 2025-11-26
`transparent` attribute ignores `opaque` attribute


ManevilleF:
Hello,

To work around the absence of `serialize_with` and `deserialize_with` on containers I'm attempting this:

```rust
#[derive(Facet)]
struct UrlWrapper(
    #[facet(
        opaque,
        serialize_with = custom_serializer,
        deserialize_with = custom_deserializer,
    )]
    MyUrl,
);
```

In this example `MyUrl` is not `Facet` as I want to apply custom parsing logic, transforming `MyUrl` to a stringified version when serialized
This compiles fine until I add

```diff
#[derive(Facet)]
+ #[facet(transparent)]
struct UrlWrapper(
    #[facet(
        opaque,
        serialize_with = custom_serializer,
        deserialize_with = custom_deserializer,
    )]
    MyUrl,
);
```

And now facet requires `Facet` for `MyUrl`, I guess by bypassing the field attributes.

I am not familiar with the facet codebase yet, so I'm not sure what would be the best way to fix this:

- Alter `transparent` proc attribute to not ignore the field attribute in unit tuples
- Directly develop `(de)serialize_with` on the container level, making this workaround obsolete

 I'm using `facet 0.31.7`

tristan:
I think making `(de)serialize_with` work with `transparent` makes sense either way. Feels like a bug that it doesn’t. I'll take a look when I next get a chance (if no one else gets to it first).

Having it on the container level probably makes sense too? I guess it's useful to avoid having to add the annotations every time you use the type? It's a bigger feature to implement though, but now the concept has been established it might not be too much work to add similar metadata to what we added to the fields to the `Shape` itself 🤔.

ManevilleF:
I can try giving it a go, taking inspiration from your work on https://github.com/facet-rs/facet/pull/916 the idea would be to add a similar vtable container for user structs ?

fasterthanlime:
> I can try giving it a go, taking inspiration from your work on [#916](https://github.com/facet-rs/facet/pull/916) the idea would be to add a similar vtable container for user structs ?

That's the idea! And then format crates need to add support for it. Seems rather straightforward.

fasterthanlime:
(well the first part is straightforward — the latter part I'll step in)

ManevilleF:
Doesn't `serialize_with` and `deserialize_with` alter the shape directly ? Shouldn't format crates like `facet-json` support out of the box once supported in `facet` ?

[...]

PR #944, opened 2025-11-28
Support de/serisalize_with inside transparent containers

tristan:
Makes `{de}serialize_with` work in `#facet[transparent]` containers.

Solves part of https://github.com/facet-rs/facet/issues/931

### DIFF
--- a/facet-json/src/serialize.rs
+++ b/facet-json/src/serialize.rs
@@ -94,7 +94,7 @@ fn write_colon<W: crate::JsonWrite>(writer: &mut W, indent: Option<&str>) {
 }
 
 fn serialize_value<'mem, 'facet, W: crate::JsonWrite>(
-    mut peek: Peek<'mem, 'facet>,
+    peek: Peek<'mem, 'facet>,
     maybe_field: Option<Field>,
     writer: &mut W,
     indent: Option<&str>,
@@ -122,9 +122,10 @@ fn serialize_value<'mem, 'facet, W: crate::JsonWrite>(
     {
         let old_shape = peek.shape();
         let ps = peek.into_struct().unwrap();
-        peek = ps.field(0).unwrap();
-        let new_shape = peek.shape();
+        let (field, inner_peek) = ps.fields().next().unwrap();
+        let new_shape = inner_peek.shape();
         trace!("{old_shape} is transparent, let's serialize the inner {new_shape} instead");
+        return serialize_value(inner_peek, Some(field), writer, indent, depth);
     }
 
     trace!(
--- a/facet-macros-emit/src/process_struct.rs
+++ b/facet-macros-emit/src/process_struct.rs
@@ -399,66 +399,70 @@ pub(crate) fn process_struct(parsed: Struct) -> TokenStream {
     // Add try_from_inner implementation for transparent types
     let try_from_inner_code = if ps.container.attrs.is_transparent() {
         if let Some(inner_field) = &inner_field {
-            // Transparent struct with one field
-            let inner_field_type = &inner_field.ty;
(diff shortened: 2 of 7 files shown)


facet-json/src/serialize.rs
```diff
@@ -122,9 +122,10 @@ fn serialize_value<'mem, 'facet, W: crate::JsonWrite>(
     {
         let old_shape = peek.shape();
         let ps = peek.into_struct().unwrap();
-        peek = ps.field(0).unwrap();
-        let new_shape = peek.shape();
+        let (field, inner_peek) = ps.fields().next().unwrap();
+        let new_shape = inner_peek.shape();
         trace!("{old_shape} is transparent, let's serialize the inner {new_shape} instead");
+        return serialize_value(inner_peek, Some(field), writer, indent, depth);
```
tristan, line 128: this recursive call was done to run the `serialize_with` check above on the inner field. we could avoid it if we duplicate the code again, or make it a reusable fn and just call it again.

facet-macros-emit/src/process_struct.rs
```diff
@@ -399,66 +399,70 @@ pub(crate) fn process_struct(parsed: Struct) -> TokenStream {
     // Add try_from_inner implementation for transparent types
     let try_from_inner_code = if ps.container.attrs.is_transparent() {
         if let Some(inner_field) = &inner_field {
-            // Transparent struct with one field
-            let inner_field_type = &inner_field.ty;
-            let bgp_without_bounds = ps.container.bgp.display_without_bounds();
+            if !inner_field.attrs.is_opaque() {
```
tristan, line 402: currently compilation fails if we try add `#[facet(opaque)]` inside a `#[facet(transparent)]`, as the generated try_from code here fails.

we could make it compile by changing `#inner_field_type` to be `::facet::Opaque<#inner_field_type>`, but I'm not sure if it makes any sense? for custom de/ser we want the non-try-from path anyways.

fasterthanlime: I think I need a lot more context to answer that question x)

tristan: For DeSer we prefer the `try_from` vtables to be `None` so that `begin_inner` will hit the [branch that calls `begin_nth_field(0)`](https://github.com/facet-rs/facet/blob/ee0684e50b9faf5b05ab5bff22fe2912c4d3cbef/facet-reflect/src/partial/partial_api/option.rs#L177). then everything else "just works".

Since there should be no one using `#[facet(opaque)]` inside `#[facet(transparent)]` already, because it wouldn't compile before, I think it's safe to go this route, and if someone comes along with a use case for having the try_from implementations on their opaque types called, we can revisit. My current knowledge of how they're used is lacking though :see_no_evil:, so I might be just be missing something.

fasterthanlime: > Since there should be no one using `#[facet(opaque)]` inside `#[facet(transparent)]` already, because it wouldn't compile before, I think it's safe to go this route, and if someone comes along with a use case for having the try_from implementations on their opaque types called, we can revisit. My current knowledge of how they're used is lacking though 🙈, so I might be just be missing something.

Okay, so the way I understand this, 9.31, the user was working around the lack of custom deserialization on containers by making the container transparent, which would then strip the outer shell and use custom deserialization of the child. 

Okay, okay, okay. And now you're saying you added support to use the parent's custom deserialization. You added custom deserialization on containers so that you don't have to use the transparent trick. But it breaks the workaround of the other user, and you're asking if that's okay. And my answer is yes, it is okay, because it's already kind of nonsense.

The workaround is kind of making me queasy. Whereas what you're doing is, I think, how custom deserialization should work.

PR #949, opened 2025-11-28
Add streaming TOML deserializer using toml_parser

fasterthanlime:
cf. #148 

## Summary

- Add new streaming deserializer for facet-toml using toml_parser's event-based API
- Always uses deferred materialization to handle TOML's interleaved keys correctly
- Handles dotted keys like `foo.bar.x = 1; foo.baz = 2; foo.bar.y = 3`
- Includes design documentation explaining the architecture

## Key components

- `SpannedEvent`: bufferable event with span and encoding info
- `EventCollector`: captures toml_parser events for replay  
- `EventIter`: pull-based iteration with peek/rewind
- `StreamingDeserializer`: processes events into `Partial`

## Test plan

- [x] All existing facet-toml tests pass (132 tests)
- [x] New streaming tests pass (simple structs, nested structs, dotted keys, arrays, inline tables, booleans, floats)


epage:
If you haven't already, I highly recommend adding compliance tests first so you have a baseline to compare with.

https://github.com/facet-rs/facet/issues/148#issuecomment-2949782499

fasterthanlime:
> If you haven't already, I highly recommend adding compliance tests first so you have a baseline to compare with.

On it!

fasterthanlime:
(Merged manually — support for a `Value` type punted for another PR — and same thing for the tests.)

(This rewrite is not getting released until these two things happen)

PR #952, opened 2025-11-28
Add facet-value crate for memory-efficient dynamic values

fasterthanlime:
## Summary

- Adds new `facet-value` crate inspired by [ijson](https://lib.rs/crates/ijson)
- Provides a pointer-sized `Value` type using tagged pointers (3 tag bits via 8-byte alignment)
- Includes `VBytes` type for binary format support (MessagePack, CBOR, etc.)
- `no_std` compatible with `alloc` feature

## Design

Following ijson's advice: struct-based API with `destructure()` methods rather than exposed enum (allows backward-compatible representation changes).

**Seven value types:**
| Type | Storage |
|------|---------|
| Null | inline (no alloc) |
| Bool | inline (no alloc) |
| Number | heap (`i64`/`u64`/`f64`) |
| String | heap (UTF-8) |
| Bytes | heap (binary) |
| Array | heap (dynamic) |
| Object | heap (ordered map) |

**Differences from ijson:**
- No string interning (simpler, `no_std` friendly)
- No small number optimization (yet)
- Added `VBytes` for binary formats

## Test plan

- [x] All 39 unit tests pass
- [ ] CI passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -690,6 +690,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "facet-value"
+version = "0.1.0"
+
+[[package]]
 name = "facet-xdr"
 version = "0.31.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ members = [
 
     # utilities
     "facet-diff",
+    "facet-value",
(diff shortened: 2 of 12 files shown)


fasterthanlime:
Merged manually

PR #993, opened 2025-12-02
allow for any type of Cow, not just Cow<str>

cramt:
I might have broken something because the build is now giving this error `warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique`, which i frankly dont even understand what means


fasterthanlime:
> `warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique`,

Ah that's no good. And it means what it means — if you have two function `a` and `b` they might compare equal today and unequal tomorrow. No guarantees. Blocks get merged, split, blah.

I'll review the code to understand why that's happening here.

fasterthanlime:
Wait, 267 commits? That’s no good either. I see a merge commit in there too. This is a no merge zone, rebase only

PR #995, opened 2025-12-02
update nix flake lock and cleanup commit hooks

cramt:
- flake lock was so out of date the rust version it provided was less than the MSRV
- fix commit hooks to not depend on the cargo bin folder being in path, also cleaned up some old ones that didnt get cleaned up from monorepo'ing again

### DIFF
--- a/facet-pretty/.cargo-husky/hooks/pre-commit
+++ b/facet-pretty/.cargo-husky/hooks/pre-commit
@@ -1,2 +1,0 @@
-cargo install --git https://github.com/facet-rs/facet-dev
-facet-dev generate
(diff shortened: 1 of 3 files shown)


fasterthanlime:
This PR is weirdly 200+ commits like the other one, ping me when it’s just 1!

cramt:
wait wtf, how did this happen, git

cramt:
@fasterthanlime is fixed, i also made a meme 
https://cdn.discordapp.com/attachments/1379550209599733837/1445537437999697981/Untitled.png?ex=6930b526&is=692f63a6&hm=3a1bc295d84066723589aa173b24f912530653cc940f0c50a5d80a943b921c5f

PR #1002, opened 2025-12-03
Fix function pointer comparison warning in define_attr_grammar!

### LINKED ISSUE #1001: Warning: function pointer comparisons in define_attr_grammar! derived PartialEq
## Description

When building `facet`, there's a compiler warning about function pointer comparisons:

```
warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
  --> facet/src/lib.rs:218:9
   |
38 | /     crate::define_attr_grammar! {
39 | |         builtin;
40 | |         ns "";
41 | |         crate_path ::facet::builtin;
.. | |
52 | |     }
   | |_____- in this macro invocation
..
18 |           $crate::__make_parse_attr! { $($grammar)* }
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: the address of the same function can vary between different codegen units
   = note: furthermore, different functions could have the same address after being merged together
   = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>
   = note: `#[warn(unpredictable_function_pointer_comparisons)]` on by default
   = note: this warning originates in the derive macro `PartialEq` which comes from the expansion of the macro `crate::define_attr_grammar` (in Nightly builds, run with -Z macro-backtrace for more info)
```

## Root Cause

The `define_attr_grammar!` macro generates an enum (in this case, `builtin::Attr`) that derives `PartialEq`. Some variants of this enum contain function pointers (e.g., `Default(make_t)`, `SkipSerializingIf`, `Invariants`). 

When Rust derives `PartialEq` for such an enum, it compares function pointers directly, which is unreliable because:
1. The same function can have different addresses in different codegen units
2. Different functions could have the same address after being merged together by the linker

## Potential Solutions

1. **Implement `PartialEq` manually** for the generated enum, comparing only the variant discriminant for function pointer variants (treating any two `Default(f1)` == `Default(f2)` as equal regardless of the function pointer, or always unequal)

2. **Remove the `PartialEq` derive** if equality comparison isn't actually needed for these types

3. **Add `#[allow(unpredictable_function_pointer_comparisons)]`** on the generated code if the current behavior is intentional (though this just silences the warning without fixing the underlying issue)

4. **Use a wrapper type** around function pointers that implements `PartialEq` in a defined way (e.g., comparing addresses with `std::ptr::fn_addr_eq` or treating all instances as unequal)

fasterthanlime:
## Summary

The `define_attr_grammar!` macro generates enums with `PartialEq` derived, but some variants contain function pointers (`DefaultInPlaceFn`, `SkipSerializingIfFn`, etc.). Comparing function pointers with `==` is unreliable across codegen units, triggering the `unpredictable_function_pointer_comparisons` warning.

## Solution

Instead of deriving `PartialEq`, generate a manual implementation that:
- Uses regular equality for unit, string, char, and struct variants
- Uses `core::ptr::fn_addr_eq` for function pointer variants
- Uses `core::ptr::eq` for Shape references

This is more explicit about the semantics: function pointer addresses may vary across codegen units, but within the same unit they should be stable.

## Testing

- `cargo build` - no warnings
- `cargo nextest run --workspace` - all 1786 tests pass

Fixes #1001

### DIFF
--- a/facet-macros-impl/src/attr_grammar/make_parse_attr.rs
+++ b/facet-macros-impl/src/attr_grammar/make_parse_attr.rs
@@ -617,11 +617,17 @@ impl ParsedGrammar {
         // For builtin mode, skip deriving Facet entirely because the derive macro
         // generates ::facet:: paths which don't work inside the facet crate.
         // Builtin attrs will have Facet implemented manually via a blanket impl.
+        //
+        // We don't derive PartialEq because some variants contain function pointers.
+        // Instead, we generate a manual impl that uses fn_addr_eq for those variants.
         let derive_attr = if self.builtin {
-            quote! { #[derive(Debug, Clone, PartialEq)] }
+            quote! { #[derive(Debug, Clone)] }
         } else {
-            quote! { #[derive(Debug, Clone, PartialEq, ::facet::Facet)] }
+            quote! { #[derive(Debug, Clone, ::facet::Facet)] }
         };
+
+        // Generate manual PartialEq implementation
+        let partial_eq_impl = self.generate_partial_eq_impl();
 
         quote! {
             #(#attrs)*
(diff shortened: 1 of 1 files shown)


fasterthanlime:
cc @cramt

PR #1043, opened 2025-12-04
feat(facet-args): support short args chaining

cestef:
Adds support for chained short flags like `-abc` instead of `-a -b -c`.
This also extracts the short flag processing logic in to a `Context::process_short_flag` helper.


fasterthanlime:
I would like to see some entries in the facet args showcase showing what error happens if you have a chain of short flags and the middle one is wrong.

```
blah -abc
       ^---- b is not a flag, there's a, c, f, z, h
```

Do you mind if I take over this PR from here and push new things into it? Or do you want to do that yourself?

cestef:
Sure, go ahead !

fasterthanlime:
image

this is what I meant — and now we have it! thanks for your contribution ✨

PR #1055, opened 2025-12-04
feat(facet-json): add from_slice_owned and from_str_owned for owned deserialization

### LINKED ISSUE #1052: Usagetest: Trying to use with axum
Hi again  @fasterthanlime 

So I'm using axum which is quite tightly integrated to serde, but allows me to implement the trait FromRequest for a custom type (I call it FacetJson<T>) so that I can use facet-json instead of serde-json to parse the incoming requets.

I'm probably missing somthing obvious, but I seem unalbe to use facet-json to parse the incoming request however due to lifetime requirements in parse_str. The following code:

```

pub struct FacetJson<T>(pub T);

impl<T, S> FromRequest<S> for FacetJson<T>
where
  T: Facet<'static>,
  S: Send + Sync,
{
  type Rejection = String;

  async fn from_request(req: Request<axum::body::Body>, state: &S) -> Result<Self, Self::Rejection> {
    if !json_content_type(req.headers()) {
      return Err("no json content type".into());
    }

    let bytes = bytes::Bytes::from_request(req, state).await.map_err(|e| {
      error!("could not decode json request: {:?}", e);
      format!("error reading bytes from body")
    })?;

    let a = facet_json::from_slice::<T>(&bytes).map(|v| FacetJson(v));

    let a = a.map_err(|e| {
      error!("could not decode json request: {:?}", e);
      format!("{:?}", e)
    });
    a
  }
}

```

Fails to compile due to the line `let a = facet_json::from_slice::<T>(&bytes).map(|v| FacetJson(v));` since facet_json requires &bytes to *outlive* the lifetime of the Facet in T (expressed as 'static, since what else can I use?)

I.e. I fail to express the lifetimes such that i can Implement FromRequest to return a value of type T (it should be possible, the returned owned value does not have a lifetime, but it "sort of gets on" since the Facet time requires one?)

fasterthanlime:
## Summary

Adds new deserialization functions that don't require the input to outlive the result:
- `from_slice_owned<T: Facet<'static>>(input: &[u8]) -> Result<T>`
- `from_str_owned<T: Facet<'static>>(input: &str) -> Result<T>`

This enables deserializing from temporary buffers, solving the issue reported in #1052 where integrating with axum's `FromRequest` was impossible due to lifetime constraints.

## Changes

- Add `allow_borrow` field to `JsonDeserializer` to track whether borrowing is allowed
- Add `new_owned()` constructor that disables borrowing
- Add runtime checks that error when attempting to deserialize into `&str` with borrowing disabled
- Add `from_slice_owned` and `from_str_owned` public functions with `Facet<'static>` bounds
- Add comprehensive tests for owned deserialization

## Safety

The implementation uses carefully audited `unsafe` code to work around lifetime constraints:
1. The `Facet<'static>` bound on the new functions guarantees the target type contains no borrowed data
2. The `allow_borrow: false` flag causes runtime errors if deserialization tries to borrow into `&str`
3. The transmutes only affect phantom lifetime markers in `Partial` and `HeapValue`, not actual data

## Test plan

- [x] Added tests for basic owned deserialization
- [x] Added tests for escaped strings
- [x] Added test simulating the axum use case (temporary buffer)
- [x] Added tests for nested structs and vectors
- [x] All 307 existing facet-json tests pass
- [x] Pre-push hooks pass (clippy, doc tests)

Fixes #1052

### DIFF
--- a/facet-json/src/deserialize.rs
+++ b/facet-json/src/deserialize.rs
@@ -380,25 +380,42 @@ pub struct JsonDeserializer<'input> {
     tokenizer: Tokenizer<'input>,
     /// Peeked token (for lookahead)
     peeked: Option<Spanned<Token<'input>>>,
+    /// Whether borrowing from input is allowed.
+    /// When false, attempting to deserialize into `&str` will error.
+    allow_borrow: bool,
 }
 
 impl<'input> JsonDeserializer<'input> {
     /// Create a new deserializer for the given input.
+    /// Borrowing from input is allowed by default.
     pub fn new(input: &'input [u8]) -> Self {
         JsonDeserializer {
             input,
             tokenizer: Tokenizer::new(input),
             peeked: None,
+            allow_borrow: true,
+        }
+    }
+
+    /// Create a new deserializer that does not allow borrowing from input.
+    /// Use this when deserializing into owned types from temporary buffers.
+    pub fn new_owned(input: &'input [u8]) -> Self {
+        JsonDeserializer {
+            input,
+            tokenizer: Tokenizer::new(input),
+            peeked: None,
+            allow_borrow: false,
         }
     }
 
     /// Create a sub-deserializer starting from a specific byte offset.
     /// Used for replaying deferred values during flatten deserialization.
-    fn from_offset(input: &'input [u8], offset: usize) -> Self {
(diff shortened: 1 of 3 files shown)


vron:
Nice, my five cents ( but 100% up to you) - i think the "default method" should be the owned, it would be by far the most common use case? And instead have from_slice_borrow

PR #1105, opened 2025-12-05
facet-shapelike

cramt:
introducing Shapelike, its a struct thats a bit like a Shape, this should be treated as the current escape hatch if you really need shape itself to implement Facet, like fx if you wanna diff 2 shapes or if you wanna serialize a shape. Shapelike is a fully owned representation of all the parts of a Shape that it makes sense to serialize (no vtable and typeid), for obvious reason the conversion is a one-way trip.

# things i dont know if i've done right
- should i disable specific features in the dependency on other facet crates?
- should this maybe not be in a seperate crate?
- the name i gave to that function i added to postcard? what should i name it?

### DIFF
--- a/facet-shapelike/src/shape_like.rs
+++ b/facet-shapelike/src/shape_like.rs
@@ -1,0 +1,574 @@
+use std::ptr::NonNull;
+
+use crate::types::{
+    EnumReprLike, FunctionAbi, KnownPointer, PointerFlags, PrimitiveType, ReprLike, ShapeLayout,
+    StructKindLike,
+};
+use facet::{Facet, PtrConst};
+use facet_core::{
+    ArrayDef, Def, EnumType, ExtensionAttr, Field, FunctionPointerDef, ListDef, MapDef, NdArrayDef,
+    OptionDef, PointerDef, PointerType, ResultDef, SequenceType, SetDef, Shape, SliceDef,
+    StructType, Type, UserType, Variant,
+};
+use facet_reflect::Peek;
+
+#[derive(facet::Facet, Clone)]
+#[repr(C)]
+pub struct ShapeLike {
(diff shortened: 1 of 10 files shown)


facet-shapelike/src/shape_like.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub key: String,
+    pub data: Vec<u8>,
+    pub shape: Box<ShapeLike>,
+}
+
+impl From<&ExtensionAttr> for ExtensionAttrLike {
+    fn from(attr: &ExtensionAttr) -> Self {
+        let ptr = PtrConst::new(NonNull::new(attr.data as *mut ()).unwrap());
+        let peek = unsafe { Peek::unchecked_new(ptr, attr.shape) };
+        let data = facet_postcard::ptr_to_vec(peek).unwrap();
```
cramt, line 80: this is kinda cursed but atleast this means its serializable, i kinda wish we could allocate it to a memory arena or something and just copy those bytes itself, im not sure if that would work, but atleast it would mean you could directly reference the type on the other side of the serialization, instead of having to go through the serialization pass twice, (also i just picked postcard cause i had to pick one)

fasterthanlime: Postcard is a good choice. The problem with just copying bytes as is is that they could also contain pointers. and then you would forget to copy the things they point to, whereas if you serialize, then hopefully you get everything.

cramt: yeah i started working on this where i would allocate the shape inside an arena and then recursively go through the shape to allocate the other pointers inside the arena aswell and then copy everything over with special cases for the pointers, tl;dr it was completely unreadable, cursed, and also segfaulted on the simple test i wrote so ehm... yeah, i went with the simple solution

Issue #996, opened 2025-12-03
Derive macro should generate #[allow(dead_code)] to avoid false positive warnings


fasterthanlime:
## Problem

When using `#[derive(Facet)]` on enums/structs that are used through reflection (e.g., `facet_args` for CLI parsing), rustc/clippy reports false positive "dead code" warnings because it doesn't see the variants/fields being constructed directly in code.

Example:

```rust
#[derive(Debug, Facet)]
#[repr(u8)]
enum Command {
    /// Validate all grammar configurations
    Lint,
    
    /// Regenerate crate files
    Gen { name: Option<String> },
    
    /// Build and serve demo
    Serve { port: Option<u16> },
}
```

This produces:
```
warning: variants `Lint`, `Gen`, and `Serve` are never constructed
  --> src/main.rs:31:5
   |
29 | enum Command {
   |      ------- variants in this enum
30 |     Lint,
   |     ^^^^
...
```

The variants ARE used - they're constructed through `facet_args::from_std_args()` via reflection - but rustc doesn't see that.

## Proposed Solution

The `Facet` derive macro should generate `#[allow(dead_code)]` on the type it's deriving for, since facet-derived types are often used through reflection rather than direct construction.

## Workaround

Currently users must manually add the allow attribute:
```rust
#[derive(Debug, Facet)]
#[repr(u8)]
#[allow(dead_code)] // variants used by facet_args derive
enum Command { ... }
```

PR #1118, opened 2025-12-05
Reduce/cordon bloat in facet, introduce bloatbench

fasterthanlime:
Reduce/cordon bloat in facet, introduce bloatbench

Looks into codegen size, compile times etc. - again :) after adding a buuuunch of features.

- [x] Disable miette's default features (derive, which depends on syn) (`b.miette`)
- [x] Don't enable miette fancy by default in facet-json (`b.miette-fancy`)
- [x] Flatten closures in facet-json to avoid generic instantiations (`b.flatten-closures`)
- [x] Get rid of shape_of in Shape derive, use `#field_type` directly (`b.shape-of`)
- [x] Detect known derives in facet-macros, eschew impls! for them (`b.detect-derives`)
- [x] Gate nonzero impls, net impls (`b.gate-impls`)
- [x] Gate doc-comments, vtable-fmt, vtable-cmp, vtable-hash (`b.gate-vtable`)
- [x] Make 'auto trait detection' opt-in (`b.auto-traits-optin`)
- [x] Detect traits from `derive` and from `#[facet(default)]` (`b.derive-traits`)
- [x] Allow manual trait specification via `#[facet(traits(...))]` (`b.manual-traits`)
- [x] Go back to builders (they expand to less code) (`b.builders`)
- [x] VTable restructured with sub-vtables (`format`, `cmp`, `hash`, `markers`) (`b.vtable-substruct`)
- [x] Convert manual impls to ShapeBuilder (`scalar.rs`, `fn_ptr.rs`, `impls_num_complex.rs`) (`b.shapebuilder`)
- [x] **Confirmed**: `#![allow(uncommon_codepoints, nonstandard_style)]` in library = no warnings in consumers (`b.unicode-test`)
- [x] Add `FieldBuilder::new(name, shape_fn, offset)` builder (`b.field-builder`)
- [x] Add `StructTypeBuilder::new(kind, fields)` builder (`b.structtype-builder`)
- [x] Add `VariantBuilder::new(name, data)` builder (`b.variant-builder`)
- [x] Add `EnumTypeBuilder::new(enum_repr, variants)` builder (`b.enumtype-builder`)
- [x] Create `::facet::𝟋` module with type aliases and builder exports (`b.prelude-mod`)
- [x] Update derive macro to use builder pattern for Field, StructType, Variant, EnumType (`b.derive-builders`)

Based on analysis in `facet-bloatbench/MACRO-OPTIMIZATION-ANALYSIS.md`:

- [x] Remove `let mut vtable = const { ... }; vtable` pattern — just use `const { ... }` directly (`b.vtable-let`)
- [x] ~~Remove redundant `{ }` braces around expressions~~ (`b.braces`) — **audited**: remaining braces are from Rust's internal `offset_of!` macro expansion, not our code
- [x] Remove `.display({ None })`, `.debug({ None })` etc. — make these defaults in builder (`b.none-defaults`)
- [x] Shorter lifetime name (`'__facet` → `'ʄ`) (`b.lifetime`)
- [x] Shorter shadow struct names (`__Shadow_RustRepr_Tuple_for_Enum003_V4` → `_TV4`) (`b.shadow-names`)
- [x] Use `loop {}` instead of `::core::panicking::panic(...)` in validation blocks (`b.loop-panic`)
- [x] ~~Eliminate nested `&const { ... &const { ... } }` blocks~~ (`b.nested-const`) — **audited**: inner `&const {}` required for static lifetime promotion of arrays containing closures/fn ptrs
- [x] Use `as _` instead of `as i64` in discriminants (`b.discrim-cast`)
- [x] ~~Remove/simplify `__facet_use_struct` validation blocks~~ (`b.validation`) — **kept**: needed for issue #996 dead code suppression

- [x] Update macro to emit `use ::facet::𝟋::*;` preamble (`b.prelude-use`)
- [x] Replace `::facet::Shape` with `𝟋Shp`, `::facet::Def` with `𝟋Def`, etc. (`b.prelude-types`)
- [x] Replace `::facet::StructKind::*` with `𝟋Sk::*` (`b.prelude-structkind`)
- [x] Replace `::facet::Type/UserType` with `𝟋Ty/𝟋UTy` (`b.prelude-usertype`)
- [x] Replace `::facet::Variant` with `𝟋Var`, builder with `𝟋VarB` (`b.prelude-variant`)
- [x] Replace `::facet::Field` with `𝟋Fld` (`b.prelude-field`)
- [x] Replace `::facet::EnumTypeBuilder` with `𝟋ETyB` (`b.prelude-enumtype`)
- [x] Replace `::facet::StructTypeBuilder` with `𝟋STyB` (`b.prelude-structtype`)
- [x] Empty field array optimization — use `&[]` directly without const block (`b.empty-fields`)
- [x] ~~Add `𝟋NOAT` / `𝟋NODOC` constants for empty attribute/doc slices~~ (`b.prelude-consts`) — **skipped**: `&[]` (3 chars) is shorter than `𝟋NOAT` (6 chars)

- [x] Add ShapeBuilder struct update syntax for default fields (`b.shape-undef`) — **done**: ShapeBuilder now only emits non-default fields
- [x] Add `StructType::UNIT` constant for unit variants (`b.structtype-unit`)
- [x] Add `Repr::C`, `Repr::RUST`, `Repr::TRANSPARENT` constants (`b.repr-consts`)
- [x] ~~Add `ValueVTable::EMPTY` constant~~ (`b.vtable-empty`) — **skipped**: VTables always need `type_name` and `drop_in_place`, can't be truly empty
- [x] Add `𝟋Ph<'a>` type alias for phantom fields (`b.phantom`)

- [x] Gate doc comments behind feature flag — default off, store None (`b.doc-feature`)
- [x] Gate grammar diagnostics (strsim dependency) behind feature — default on (`b.diag-feature`)

- [ ] Add short re-exports for all variants of `Def` enum (`b.def-variants`)
- [ ] Add short re-exports for all variants of `Type` enum (`b.type-variants`)
- [ ] Consider lifting const blocks higher to reduce nesting (`b.const-lift`)

**Test result**: `#![allow(uncommon_codepoints, nonstandard_style)]` in facet crate = **no warnings in consumer crates!**

Module name `𝟋` + all exports prefixed with `𝟋` = no collisions after `use ::facet::𝟋::*;`

| Full Name | 𝟋 Export | Notes |
|-----------|----------|-------|
| Shape | `𝟋Shp` | |
| Field | `𝟋Fld` | |
| Facet | `𝟋Fct` | trait |
| Variant | `𝟋Var` | |
| Type | `𝟋Ty` | |
| UserType | `𝟋UTy` | CamelCase |
| StructType | `𝟋STy` | CamelCase |
| StructKind | `𝟋Sk` | |
| ValueVTable | `𝟋Vt` | |
| Repr | `𝟋Repr` | |
| MarkerTraits | `𝟋Mt` | |
| Def | `𝟋Def` | |
| EnumType | `𝟋ETy` | CamelCase |
| EnumRepr | `𝟋ERpr` | |
| ShapeBuilder | `𝟋ShpB` | builder |
| ValueVTableBuilder | `𝟋VtB` | builder |
| StructTypeBuilder | `𝟋STyB` | builder |
| FieldBuilder | `𝟋FldB` | builder |
| VariantBuilder | `𝟋VarB` | builder |
| EnumTypeBuilder | `𝟋ETyB` | builder |
| PhantomData | `𝟋Ph<'a>` | type alias for phantom fields |

**Why `&const {}` is required**: The pattern `let fields: &'static [Field] = &const {[...]};` is necessary because:
1. Field/Variant structs contain closures (e.g., `|| T::SHAPE`) that create temporaries
2. Attribute arrays may contain function pointers (e.g., `__proxy_convert_in`)
3. These temporaries need promotion to `'static` lifetime, which only `&const {}` provides
4. Named const items (`const FIELDS: [Field; N] = [...]`) don't work because generic parameters can't be referenced in const items

**Empty array optimization works**: For empty arrays, `&[]` works directly without `&const {}` since it's already a static reference.

| Macro | Before | After | Change |
|-------|--------|-------|--------|
| `_impls!` | 1.46MB | 5KB | -99.7% |
| `::facet::value_vtable!` | 1.4MB | 0 | -100% |
| `::facet_core::spez::impls!` | 99KB | 343B | -99.7% |
| `#[derive(Facet)]` | 1.05MB | 806KB | -23% |
| `::core::mem::offset_of!` | 77KB | 57KB | -26% |

**Total macro-generated bytes:** ~4.2MB → ~988KB (roughly **4x reduction**)


fasterthanlime:
image

I'm happy with the result. I had to put some features behind the gate, but I think it's still a useful version of Facet that enables serialization and deserialization. I think I'm gonna bring back the marker trait detection behind a feature flag, just like I did the format traits, the comparison traits and the hash trait.

And then we can actually merge this and just have better performance.

fasterthanlime:
In order to simplify everything and keep a lot of speed, we are going to remove all specialization in Facet.

We’re going to detect traits that are derived, which we already do, and everything else will be explicitly declared.

For all standard library types, we’re going to declare them ourselves, which means facet-diff, facet-assert, and facet-pretty keep working. We don’t really need Display for facet-pretty for composite types, and for new scalars from third-party crates, they can declare their display traits explicitly.

And that’s it.

edit: no, we can have both, see https://github.com/facet-rs/facet/issues/1126

PR #1143, opened 2025-12-06
Experiment: proc-macro plugin registry via shared dylib

fasterthanlime:
# Proc-Macro Plugin Registry Experiment

Testing whether two proc-macro crates can share state via a common `crate-type = "dylib"` dependency.

## The Trick

When multiple proc-macros depend on the same dylib crate, the dynamic linker (ld.so) should load it once and share it between them. This means a static `Mutex<Vec<Plugin>>` in the dylib would be shared across all proc-macros.

```
registry (crate-type = "dylib")
    - static REGISTRY: Mutex<Vec<Plugin>>
    - register_plugin(), list_plugins()
         ↑                    ↑
         |                    |
    plugin-a             plugin-b
    (proc-macro)         (proc-macro)
    registers via        reads registry
    ctor
```

## Why This Matters

If this works, facet could have an **extensible macro system**:

1. `facet-macros` reads a plugin registry during `#[derive(Facet)]`
2. `facet-error` registers an error codegen plugin when loaded
3. `facet-builder` registers a builder codegen plugin when loaded
4. etc.

**Benefits:**
- No double-parsing (parse once in `#[derive(Facet)]`, plugins add their codegen)
- No build.rs required in user crates
- No separate crates for types
- "Just depend on facet-error and it works"

## Testing

```bash
cd experiments/plugin-test/user-crate
cargo build 2>&1 | grep -E '\[registry\]|\[plugin'
```

**Expected if it works:**
```
[registry] registering plugin: plugin-a
[plugin-b] invoke_b called, checking registry...
[registry] listing 1 plugins
[plugin-b] found plugin: plugin-a -> hello from plugin-a
```

**Expected if it doesn't work:**
```
[registry] registering plugin: plugin-a
[plugin-b] invoke_b called, checking registry...
[registry] listing 0 plugins
[plugin-b] NO PLUGINS FOUND - registry not shared!
```

## Open Questions

- Does rustc actually share dylibs between proc-macros?
- What's the load order? Can we control it?
- Does this work on all platforms (Linux, macOS, Windows)?
- Any concerns with hermetic builds / determinism?

## Credit

Idea from @gregmorenz on Bluesky:
> "they should both share the same version of any dynamic library dependency – so a `static registry: Mutex<Vec<Plugin>>` should be able to coordinate between them"


experiments/plugin-test/plugin-b/src/lib.rs
```diff
@@ -0,0 +1,39 @@
+use proc_macro::TokenStream;
+
+type CodegenFn = fn(&str) -> String;
+
+extern "Rust" {
+    fn registry_list() -> Vec<(&'static str, CodegenFn)>;
```
bjorn3, line 6: Passing Vec across the ABI boundary will lead to issues if either side uses `#[global_allocator]` as you would then allocate with one allocator and deallocate with another.

experiments/plugin-test/plugin-b/build.rs
```diff
@@ -0,0 +1,14 @@
+fn main() {
+    // OUT_DIR is something like:
+    // /path/to/target/debug/build/plugin-b-hash/out
+    // We want: /path/to/target/debug
```
bjorn3, line 4: FYI the target directory layout will change in the near future: https://github.com/rust-lang/cargo/issues/15010

experiments/plugin-test/plugin-b/src/lib.rs
fasterthanlime: > if either side uses `#[global_allocator]` as you would then allocate with one allocator and deallocate with another.

Makes sense! In this case you would assume `facet-thirdparty` would... not. x) But yes.

fasterthanlime:
Closing this since, as per @bjorn3 [on Mastodon](https://hachyderm.io/@bjorn3/115674595674513283):

> There is no guarantee (on proc-macro load order). Rust-analyzer will load them in whatever order, while rustc is currently working on parallelization of macro expansion. And we might get caching of macro expansions in the future.

Which makes this trick unusable. It was fun to look into it though!

gmorenz:
I think there's a version of this that works with non-deterministic load order and parallelism, because we always know that the shared dependency is loaded before anything that depends on it. Instead of using a plugin registry (where we don't know that all plugins are done) we could have a cache of shared work that every facet-x crate works off of.

In addition to saving on duplicate work, this can reduce code duplication, e.g. with the shared dep doing something like

```rust
static PARSED_STRUCTS: Mutex<HashMap<Tokens, Arc<OnceLock<ParsedRepr>>>

pub fn parse(tokens: Tokens) -> ParsedRepr {
     let guard = PARSED_STRUCTS.lock();
     if let Some(already_parsing) = PARSED_STRUCTS.lock() {
          let already_parsing = already_parsing.clone();
          drop(guard);
          return already_parsing.wait().clone()
     }
     let pending_result = Arc::new(OnceLock::new());
     PARSED_STRUCTS.insert(tokens, pending_result.clone());
     drop(guard);
     let res = parse(tokens);
     pending_result.set(res.clone());
     res
}

bjorn3:
Wouldn't that cause a memory leak in the context of rust-analyzer as you would never get rid of cache entries?

fasterthanlime:
> Wouldn't that cause a memory leak in the context of rust-analyzer as you would never get rid of cache entries?

well, just like any cache, we could clear it when it gets too big / too stale 🤷

I don't hate this

BUT since "caching of proc macro outputs" is in the cards, I think I'd rather "parse once per derive" or "have Error, Display etc. features on the main facet crate so it does all the work in one fell swoop and not as plugins".

gmorenz:
Good point on rust analyzer.

Now I'm wondering both if rust analyzer periodically unloads and reloads proc macros (to imitate actual compile cycles) and if statics set up destructors properly to be called on unloading. I'm guessing at least the first doesn't happen.

bjorn3:
If I recall correctly rust-analyzer runs a separate proc-macro-server process for every proc-macro. In any case rust analyzer doesn't unload proc macros unless the proc macro crate itself is changed.

fasterthanlime:
> doesn't unload proc macros unless the proc macro crate itself is changed

Self-modifying code seems like the only reasonable option then

experiments/plugin-test/plugin-b/src/lib.rs
rlabrecque: I've been attempting something somewhat similar, and what I was going to look into doing was creating a wrapper allocator that just points to some common central one [somehow].
In my case I control each crate though.

But this is back when I was playing around with rubicon, where the global allocator was the only real blocker for me.

fasterthanlime: mhhhh that feels feasible? I don't see why you couldn't. it might be.. more expensive, due to LazyLock (on first access having to load/init the allocator provided by a shared dylib etc.)

mhh.

PR #1145, opened 2025-12-06
fix(facet-reflect): Use `'mem` as return lifetime of `Peek::get()` to prevent lifetime issue "function was supposed to return data with lifetime `'mem` but it is returning data with lifetime `'a`"

manuel-woelker:
Hi, I ran into the issue that the return lifetime of `Peek::get` was that of &self, rather than `'mem`. This fix attempts to address this issue.

Full disclosure: I haven't done much around `unsafe`, lifetimes and pointer shenanigans, so a good look is definitely warranted.

My understanding is that the lifetime of the underlying data pointer is `'mem` so returning a reference with that lifetime _should_ be ok.
`Peek::as_str()` and `Peek::data()` also return with a `'mem` lifetime, which seems to support this interpretation.

I also ran the test suite and the `just miri` without any errors.

Please let me know if there are any issues.

Note: I did not run the pre-commit hook, since I am currently on my windows machine. I hope the formatting, etc is ok, but feel free to touch it up.

### DIFF
--- a/facet-reflect/src/peek/value.rs
+++ b/facet-reflect/src/peek/value.rs
@@ -217,7 +217,7 @@ impl<'mem, 'facet> Peek<'mem, 'facet> {
     ///
     /// Panics if the shape doesn't match the type `T`.
     #[inline]
-    pub fn get<T: Facet<'facet> + ?Sized>(&self) -> Result<&T, ReflectError> {
+    pub fn get<T: Facet<'facet> + ?Sized>(&self) -> Result<&'mem T, ReflectError> {
         if self.shape != T::SHAPE {
             Err(ReflectError::WrongShape {
                 expected: self.shape,
--- a/facet-reflect/tests/peek/get.rs
+++ b/facet-reflect/tests/peek/get.rs
@@ -1,0 +1,19 @@
+use facet_reflect::Peek;
+
+#[test]
+fn test_peek_get_via_mem() {
+    fn retrieve_via_as_str<'a, 'mem, 'facet>(peek: &'a Peek<'mem, 'facet>) -> &'mem str {
+        peek.as_str().unwrap()
+    }
+
+    fn retrieve_via_get<'a, 'mem, 'facet>(peek: &'a Peek<'mem, 'facet>) -> &'mem str {
+        peek.get::<str>().unwrap()
+    }
+
+    let s = String::from("abc");
+    let peek = Peek::new(s.as_str());
+    let s1 = retrieve_via_as_str(&peek);
+    assert_eq!(s1, "abc");
+    let s2 = retrieve_via_get(&peek);
+    assert_eq!(s2, "abc");
+}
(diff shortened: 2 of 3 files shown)


manuel-woelker:
I just saw that 'facet-reflect::compile_tests test_peek_contravariant_shrinking' now fails, since it compiles, although it shouldn't.

If the type T itself is contravariant, that might throw a wrench into things.
I am wondering if this might also have some bearing on #1131 ... and if there is some way to prevent type from being contravariant...

fasterthanlime:
> I just saw that 'facet-reflect::compile_tests test_peek_contravariant_shrinking' now fails, since it compiles, although it shouldn't.

Yeah I.. am not sure of the implications there. Maybe I should tag @Veykril

manuel-woelker:
I tried to find the reason for Chesterton's 'facet Invariance Fence, both using lifetimes in function arguments and &mut references.
(Those are the ones I found in the [nomicon's chapter on Subtyping and Variance](https://doc.rust-lang.org/nomicon/subtyping.html#:~:text=Here%20is%20a%20table%20of%20some%20other%20generic%20types%20and%20their%20variances))

For function arguments, I basically tried to abuse the covariance to force a non-static str into a `&'static str` using function pointers, but fortunately `Facet` is not implement for those.

The other reason for invariance I could find was &mut references, but AFAICT there's no way to get such a mutable references from a `Peek`.
Interior mutability via `RefCell` and friends is also blocked because `Facet` is not implemented for those either.

My experiments are here if anyone is interested:
https://github.com/manuel-woelker/facet/commit/0afc814dffc30239d6a8d5d3dd2f2f923242dce6

fasterthanlime:
> but fortunately Facet is not implement for those.

It is behind a feature flag!

flag:

https://github.com/facet-rs/facet/blob/48c94835c542f4e7cf7a9622f0d2e8dbe7e10c72/facet-core/Cargo.toml#L55

impl:

https://github.com/facet-rs/facet/blob/24cacb12c1c1f667961bc4bcfb9ec288693fa165/facet-core/src/impls_core/fn_ptr.rs

Veykril:
Yea this should be fine, the important bit is the `'facet` lifetime in the trait bound, the reference itself is just pointing to memory though.

Not sure why the test case now compiles though, sounds like `Peek::new` picks the wrong `'facet` lifetime? I wonder if the derive produces incorrect bounds here?

fasterthanlime:
> Yea this should be fine

Well! apparently it's not - we're gonna start tracking variance: https://github.com/facet-rs/facet/issues/1168

Veykril:
Oh, it's not invariant over 'facet? 😓 yea that is not good

fasterthanlime:
> Oh, it's not invariant over 'facet? 😓

Yeah we recently undid that b/c I couldn't remember why it was bad. should've tagged you then!!

But tracking variance is even better :) means we still can do covariant stuff 90% of the time, without doing UB the other 10%

PR #1170, opened 2025-12-08
feat: variance tracking with automatic computation from field types

### LINKED ISSUE #1168: Track lifetime variance at runtime for safe Peek lifetime manipulation
## Summary

Peek is currently covariant over `'facet`, which combined with certain API patterns can lead to unsound lifetime laundering. We should track variance at runtime and provide variance-aware APIs.

## The Problem

PR #1145 attempted to fix `Peek::get()` to return `&'mem T` instead of `&T`. This is correct (the data lives for `'mem`), but it exposed a soundness hole: Peek's covariance over `'facet` allows "laundering" lifetimes through reflection.

### Proof of Concept (UB confirmed by miri)

```rust
use facet::Facet;
use facet_reflect::Peek;
use std::cell::Cell;

thread_local! {
    static STOLEN: Cell<Option<&'static str>> = Cell::new(None);
}

#[derive(Debug, Facet)]
struct FnWrapper<'a> {
    f: fn(&'a str),  // contravariant in 'a
}

fn store_ref(s: &'static str) {
    STOLEN.with(|cell| cell.set(Some(s)));
}

fn main() {
    let wrapper: FnWrapper<'static> = FnWrapper { f: store_ref };
    
    // Launder FnWrapper<'static> -> FnWrapper<'a> via Peek's covariance
    fn scope<'l, 'a>(token: &'l FnWrapper<'static>) -> &'l FnWrapper<'a> {
        Peek::new(token).get::<FnWrapper<'a>>().unwrap()
    }
    
    {
        let temp = String::from("I will be freed!");
        let laundered = scope(&wrapper);
        (laundered.f)(&temp);  // Stores non-'static ref as 'static!
    }
    
    // Use-after-free!
    STOLEN.with(|cell| {
        if let Some(s) = cell.get() {
            println!("STOLEN: {}", s);  // UB: dangling reference
        }
    });
}
```

Requires `fn-ptr` feature. Miri confirms: `constructing invalid value: encountered a dangling reference (use-after-free)`

## Proposed Solution

Instead of making Peek invariant (which breaks legitimate covariance use cases), track variance at runtime:

### 1. Add `Variance` enum to `facet-core`

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum Variance {
    Covariant,      // Can shrink lifetimes ('static -> 'a)
    Contravariant,  // Can grow lifetimes ('a -> 'static)
    Invariant,      // No lifetime changes allowed
    None,           // No lifetime parameter
}
```

### 2. Add variance field to `Shape`

The derive macro computes variance by analyzing field types:
- `&'a T` → Covariant
- `&'a mut T` → Invariant  
- `fn(&'a T)` → Contravariant
- `fn() -> &'a T` → Covariant
- `Cell<&'a T>` → Invariant
- Mixed → Invariant

### 3. Add variance-aware methods to `Peek`

```rust
impl<'mem, 'facet> Peek<'mem, 'facet> {
    /// Get with exact lifetime (always safe, current behavior)
    pub fn get<T: Facet<'facet>>(&self) -> Result<&'mem T, ReflectError>;
    
    /// Get with shorter lifetime (panics if not covariant)
    pub fn get_shrinking<'shorter, T: Facet<'shorter>>(&self) -> Result<&'mem T, ReflectError>
    where 'facet: 'shorter;
    
    /// Get with longer lifetime (panics if not contravariant)  
    pub fn get_growing<'longer, T: Facet<'longer>>(&self) -> Result<&'mem T, ReflectError>
    where 'longer: 'facet;
}
```

### 4. Make Peek invariant by default

Change from `PhantomData<&'facet ()>` to `PhantomData<fn(&'facet ()) -> &'facet ()>`.

The covariance tests would migrate to use `get_shrinking()`.

## Benefits

- Fixes the soundness hole
- Preserves legitimate use cases via explicit opt-in
- Runtime checks catch variance mismatches
- Reflects (pun intended) actual Rust semantics

### LINKED ISSUE #1171: feat: compute variance automatically from field types
## Summary

Currently, all derived types default to `Variance::Invariant` (the safe choice). We could compute variance automatically by analyzing field types and combining their variances.

## Background

PR #1170 added variance tracking to `Shape` and made `Peek` invariant for soundness. However, the derive macro doesn't actually compute variance - it just defaults to `Invariant`.

## Proposal

Compute variance automatically using these rules:

1. **Covariant fields** (e.g., `&'a T`, `Box<&'a T>`): struct is covariant if ALL fields are covariant
2. **Contravariant fields** (e.g., `fn(&'a T)`): struct is contravariant if ALL fields are contravariant  
3. **Mixed or invariant fields**: struct becomes invariant

The combination follows `Variance::combine()`:
```rust
impl Variance {
    pub const fn combine(self, other: Variance) -> Variance {
        match (self, other) {
            (Covariant, Covariant) => Covariant,
            (Contravariant, Contravariant) => Contravariant,
            _ => Invariant,
        }
    }
}
```

### Option A: Static computation in derive macro
Analyze field types at compile time and set variance statically. Challenge: need to know field type variance at macro expansion time.

### Option B: Lazy computation via function pointer
Store a function that computes variance on demand:

```rust
pub struct Shape {
    // Instead of:
    // pub variance: Variance,
    
    // We could have:
    pub variance: fn() -> Variance,
}
```

This allows each type's variance function to call its field types' variance functions, naturally handling recursive types and avoiding compile-time ordering issues.

### Option C: Lazy with caching
Same as B, but cache the result after first computation to avoid repeated work.

## @fasterthanlime's note

> I'm happy to make variance 'lazy', ie. behind a function that calls a function that calls a function etc.

## Related

- #1168 - Original soundness issue
- #1170 - Variance tracking implementation

fasterthanlime:
## Summary

This PR adds variance tracking to facet and computes variance automatically from field types.

### Initial fix (commits 1439064a, 5c668fd9)
- Add `Variance` enum (`Covariant`, `Contravariant`, `Invariant`) to facet-core
- Add `variance` field to `Shape` struct 
- Fix Peek's soundness by making it invariant over `'facet`

### Automatic variance computation (commit 4664fbb9)
- Change `Shape.variance` from `fn() -> Variance` to `fn(&'static Shape) -> Variance` to allow variance functions to walk type structure without capturing generics
- Add `Shape::computed_variance()` that walks struct fields, enum variants, and inner shapes (Box, Vec, Option, etc.) to compute variance automatically
- Use **thread-local cycle detection** to prevent stack overflow with recursive types - when a cycle is detected, returns `Covariant` (identity for `combine`)
- Update derive macros to use `.variance(Shape::computed_variance)` instead of generating per-type code
- Update `Box<T>` to use computed variance (depends on T's variance)

## Test plan

- [x] All 2058 tests pass
- [x] Added variance tests for recursive types (no stack overflow)
- [x] Compile test `fn_ptr_ub_exploit.rs` verifies the soundness fix

Closes #1168
Closes #1171

### DIFF
--- a/facet-core/src/impls_alloc/boxed.rs
+++ b/facet-core/src/impls_alloc/boxed.rs
@@ -1,4 +1,3 @@
-use crate::Variance;
 use core::{alloc::Layout, ptr::NonNull};
 
 use alloc::boxed::Box;
@@ -121,7 +120,7 @@ unsafe impl<'a, T: ?Sized + Facet<'a>> Facet<'a> for Box<T> {
             type_tag: None,
             inner: Some(T::SHAPE),
             proxy: None,
-            variance: Variance::Invariant,
+            variance: Shape::computed_variance,
         }
     };
 }
--- a/facet-core/src/impls_alloc/cow.rs
+++ b/facet-core/src/impls_alloc/cow.rs
@@ -1,6 +1,5 @@
-use crate::Variance;
 use crate::{
-    Def, Facet, FormatVTable, PtrConst, Shape, Type, UserType, ValueVTable,
+    Def, Facet, FormatVTable, PtrConst, Shape, Type, UserType, ValueVTable, Variance,
     shape_util::vtable_for_ptr,
 };
 use alloc::borrow::Cow;
@@ -73,6 +72,9 @@ where
         type_tag: None,
         inner: None,
         proxy: None,
-        variance: Variance::Invariant,
+        // Cow<'a, T> is covariant in 'a - it can borrow short-lived data
+        // but not extend lifetimes. Since Cow is a wrapper, we conservatively
+        // use INVARIANT as the default.
+        variance: Variance::INVARIANT,
     };
 }
--- a/facet-core/src/impls_alloc/vec.rs
+++ b/facet-core/src/impls_alloc/vec.rs
@@ -1,4 +1,3 @@
-use crate::Variance;
 use core::ptr::NonNull;
 
 use crate::shape_util::vtable_for_list;
@@ -105,9 +104,11 @@ where
             doc: &[],
             attributes: &[],
             type_tag: None,
+            // Vec uses Def::List, not inner, for element access
             inner: None,
             proxy: None,
-            variance: Variance::Invariant,
+            // Vec<T> is covariant in T, computed from the element type
+            variance: Shape::computed_variance,
         }
     };
 }
--- a/facet-core/src/impls_bytes.rs
+++ b/facet-core/src/impls_bytes.rs
@@ -100,7 +100,8 @@ unsafe impl Facet<'_> for Bytes {
             type_tag: None,
             inner: Some(BytesMut::SHAPE),
             proxy: None,
-            variance: Variance::Invariant,
+            // Bytes has no lifetime parameters, so it's covariant
+            variance: Variance::COVARIANT,
         }
     };
 }
@@ -185,7 +186,8 @@ unsafe impl Facet<'_> for BytesMut {
             type_tag: None,
             inner: None,
             proxy: None,
-            variance: Variance::Invariant,
+            // BytesMut has no lifetime parameters, so it's covariant
+            variance: Variance::COVARIANT,
         }
     };
 }
--- a/facet-core/src/impls_camino.rs
+++ b/facet-core/src/impls_camino.rs
@@ -61,7 +61,8 @@ unsafe impl Facet<'_> for Utf8PathBuf {
             type_tag: None,
             inner: Some(<String as Facet>::SHAPE),
             proxy: None,
-            variance: Variance::Invariant,
+            // Utf8PathBuf has no lifetime parameters, so it's covariant
+            variance: Variance::COVARIANT,
         }
     };
 }
@@ -85,7 +86,8 @@ unsafe impl Facet<'_> for Utf8Path {
             type_tag: None,
             inner: None,
             proxy: None,
-            variance: Variance::Invariant,
+            // Utf8Path has no lifetime parameters, so it's covariant
+            variance: Variance::COVARIANT,
         }
     };
 }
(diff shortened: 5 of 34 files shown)


manuel-woelker:
Hi again, glad to see the invariance mystery resolved 👍

I tried integrating this commit into my code base, but I unfortunately hit some snags.

I am wrapping `Peek` values to add origin and provenance info, which means the lifetimes leak into the wrapper type.
I am combining multiple of these wrapped values from various sources (with different lifetimes) into a hashmap with a single lifetime. Because Peek is lifetime invariant, all structs using it are now also invariant, which makes it rather hard to use for my use case (which is plain old data and should be strictly covariant by design).

This means that I would have to `shrink_lifetime()` the whole structure recursively, potentially multiple times, which I would like to avoid since it defeats the purpose of passing around only references.

I thought of the following approaches which all seem to have some drawbacks.

1. I tried to make `Peek` generic over lifetime variance (i.e. adding a VARIANCE type parameter), but AFAICT that would require Higher Kinded Types, so this is probably a no-go at the moment.

2. Another option might be to only enforce the lifetime invariance if the `fn-ptr` feature is enabled (or to opt-out via another feature), which from my understanding should be ok. This has the draw-back of having two cases to test and consider, as well as preventing combining both features in a single project.

3. A third option might be to duplicate `Peek` as `CovariantPeek` (and verify/assert covariance in the constructor). But this code duplication for a single marker trait does not seem elegant either... (maybe using a macro would make it palatable?)

I feel kinda bad bothering you with these issues. I think facet-rs unlocks a boatload of neat functionality, and I hope these improvements help other projects down the line.

Let me know if there is anything I can do to help out. Cheers and thanks for the great work!

fasterthanlime:
> * I tried to make `Peek` generic over lifetime variance (i.e. adding a VARIANCE type parameter), but AFAICT that would require Higher Kinded Types, so this is probably a no-go at the moment.

I kinda like this idea but.. yeah it's probably not gonna plug into the Rust type system _like that_.

> 3\. A third option might be to duplicate `Peek` as `CovariantPeek` (and verify/assert covariance in the constructor). But this code duplication for a single marker trait does not seem elegant either... (maybe using a macro would make it palatable?)

That's not a bad idea. `CovariantPeek` could `Deref<Target=Peek>` to avoid all the duplication ?

PR #1192, opened 2025-12-08
feat(facet-reflect): add CovariantPeek for covariant types

fasterthanlime:
## Summary

Adds a `CovariantPeek` wrapper that is covariant over `'facet`, unlike the invariant `Peek`. This allows storing multiple values with different lifetimes in a single collection without repeated `shrink_lifetime` calls.

`CovariantPeek`:
- Verifies at construction time that the underlying type is actually covariant
- `Deref`s to `Peek`, avoiding code duplication while still providing full access to `Peek` methods
- Provides `new()`, `new_unchecked()`, and `from_ref()` constructors

This addresses the use case discussed in PR #1170 comments by @manuel-woelker and @fasterthanlime.

## Example

```rust
use facet::Facet;
use facet_reflect::{Peek, CovariantPeek};

#[derive(Facet)]
struct Data<'a> {
    value: &'a str,
}

// Data<'a> is covariant in 'a because &'a str is covariant
let data = Data { value: "hello" };
let peek: Peek<'_, 'static> = Peek::new(&data);

// Convert to CovariantPeek - this verifies covariance
let covariant = CovariantPeek::new(peek).expect("Data is covariant");

// Now we can use it where shorter lifetimes are expected
fn use_shorter<'a>(p: CovariantPeek<'_, 'a>) {
    // Full access to Peek methods via Deref
    let _ = p.as_str();
}
use_shorter(covariant);
```

## Test plan

- [x] Unit tests for `CovariantPeek` construction from covariant types
- [x] Test that `Deref` to `Peek` works correctly
- [x] Test that lifetime covariance compiles (passing `'static` where `'a` expected)
- [x] All existing tests pass (354 tests)
- [x] Doctests pass

### DIFF
--- a/facet-reflect/src/peek/value.rs
+++ b/facet-reflect/src/peek/value.rs
@@ -841,6 +841,199 @@ impl<'mem, 'facet> core::hash::Hash for Peek<'mem, 'facet> {
     }
 }
 
+/// A covariant wrapper around [`Peek`] for types that are covariant over their lifetime parameter.
+///
+/// Unlike [`Peek`], which is invariant over `'facet` for soundness reasons,
+/// `CovariantPeek` is **covariant** over `'facet`. This means a `CovariantPeek<'mem, 'static>`
+/// can be used where a `CovariantPeek<'mem, 'a>` is expected.
+///
+/// # When to Use
+///
+/// Use `CovariantPeek` when you need to:
+/// - Store multiple `Peek` values with different lifetimes in a single collection
+/// - Pass `Peek` values to functions expecting shorter lifetimes
+/// - Build data structures that wrap `Peek` without forcing invariance on the wrapper
+///
+/// # Safety
+///
+/// `CovariantPeek` can only be constructed from types that are actually covariant.
+/// The constructor verifies this at runtime by checking [`Shape::computed_variance`].
+/// This ensures that lifetime shrinking is always safe.
+///
+/// # Example
+///
+/// ```
+/// use facet::Facet;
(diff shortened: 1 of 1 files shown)


fasterthanlime:
cc @manuel-woelker something like this?

manuel-woelker:
Hi, this works like a charm! 👍 Thanks for taking the time to address this.

I also did some experiments trying to add a version of `CovariantPeek::new_from(t: &'mem T) -> CovariantPeek<T>` that refuses to compile if T is not covariant, to prevent having the compile time check and having to return an `Option`.
I thought this would be easy after having the compiler going on about this all day.
But when I *wanted* the compiler to complain, I couldn't get it to statically reject invariant types. 🤷 

During these experiments, I was also trying to cross-check with the variance detection for `CovariantPeek`.
To me, it looks like types that *should* be invariant (e.g. fn-ptrs) are detected as covariant. This might also be a misunderstanding on my part due to my still shaky grasp of variance rules.

With that caveat, here's the branch with the failing testcases:
https://github.com/manuel-woelker/facet/tree/feat/covariant-peek-compile-time-check
https://github.com/manuel-woelker/facet/commit/3c1e95eb96bf36a75966590a321c475cb83f740d

fasterthanlime:
> During these experiments, I was also trying to cross-check with the variance detection for `CovariantPeek`.
> To me, it looks like types that _should_ be invariant (e.g. fn-ptrs) are detected as covariant. This might also be a misunderstanding on my part due to my still shaky grasp of variance rules.

sounds like bug material - can you open an issue about it please?

PR #1227, opened 2025-12-10
fix: skip static SHAPE declaration for generic types

### LINKED ISSUE #1225: Usagetest: Deriving facet for struct with generic parameter
From latest master (b11ba96)

The following code 

```rust

#[derive(Clone, Facet)]
pub struct SResults<U> {
  pub date: DateTime<Local>,
  pub name: String,
  pub grid: Vec<NaiveDate>,
  pub res: HashMap<String, U>,
  pub datat: HashMap<String, HashMap<String, HashMap<Arc<str>, f32>>>,
  pub flows: Option<CashflowRes>,
}
```

Gives the error:
```
 missing generics for struct `SResults`
 --> lib.rs:7:12
  |
7 | pub struct SResults<U> {
  |            ^^^^^^^^^^^^^^^^^ expected 1 generic argument
  |
note: struct defined here, with 1 generic parameter: `U`
 --> lib.rs:7:12
  |
7 | pub struct SResults<U> {
  |            ^^^^^^^^^^^^^^^^^ -
help: add missing generic argument
  |
7 | pub struct SResults<U><U> {
  |                             +++
```

But if i remove the derive(Facet) there is no compilation error

fasterthanlime:
## Summary
- Skip generating the static `TYPE_SHAPE` declaration for types with type or const generics
- The static declaration was generating invalid code like `static MY_RESULT_SHAPE: &'static Shape = <MyResult as Facet>::SHAPE;` which is missing the generic arguments

## Test plan
- [x] `cargo nextest run --release -p facet-postcard --test enums generic_enum` passes (this was failing before the fix)
- [x] Full test suite passes in release mode: `cargo nextest run --release` (2196 tests)

Fixes #1225

### DIFF
--- a/facet-macros-impl/src/derive.rs
+++ b/facet-macros-impl/src/derive.rs
@@ -5,7 +5,17 @@ use crate::{LifetimeName, RenameRule, process_enum, process_struct};
 
 /// Generate a static declaration that pre-evaluates `<T as Facet>::SHAPE`.
 /// Only emitted in release builds to avoid slowing down debug compile times.
-pub(crate) fn generate_static_decl(type_name: &Ident, facet_crate: &TokenStream) -> TokenStream {
+/// Skipped for generic types since we can't create a static for an unmonomorphized type.
+pub(crate) fn generate_static_decl(
+    type_name: &Ident,
+    facet_crate: &TokenStream,
+    has_type_or_const_generics: bool,
+) -> TokenStream {
+    // Can't generate a static for generic types - the type parameters aren't concrete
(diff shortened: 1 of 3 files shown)


fasterthanlime:
@vron So the story behind this one is pretty funny. We used to have this static that forced monomorphization of associated const shape for all types, and removing it made debug a lot faster. But it made release slower. So in the latest versions of Facet, it's omitted on debug so that it's faster, and it's emitted on release so that it's faster!

Problem is that apparently it doesn't work with generics. I don't know what we did before that, but it's fine.

PR #1327, opened 2025-12-17
feat: add support for decoding floats in msgpack

RyanSquared:
There's the potential for nuance based on how floats are cast. I figured it was fine that f32 gets cast up to f64 when parsing, then back down to f32 when being placed into the partial. By IEEE 754 specification, there should be no issues - it's just moving bits around; the bits themselves aren't truncated in any way.

The odd case is when a type is f32, but it is decoded from an f64. In that case, edge cases like `f32::MAX != (f64::MAX as u32)` may occur, which may be reflected in msgpack structures generated by systems that only support double-precision floats.

In that case, there's two (acceptable) options: throw a type error, or parse a valid f32 and ensure the value being cast down to f32 and up to f64 does not change the value, in which case, the f32 variant may be safely returned. I went with theb est-attempt to parse a valid f32. Two tests have been added to reflect this behavior: successfully parsing a f32, and a FloatOverflow when a f64 can't be represented as an f32.

I noted that the errors from msgpack were not non_exhaustive, which didn't reflect errors from some other crates. I've added the attribute, but it is technically a breaking change.

### DIFF
--- a/facet-msgpack/src/deserialize.rs
+++ b/facet-msgpack/src/deserialize.rs
@@ -174,6 +174,44 @@ impl<'input> Decoder<'input> {
         }
     }
 
+    /// Decodes a MessagePack-encoded IEEE 754 single precision floating point number
+    /// This is a low-level method used by other decoders.
+    fn decode_f32(&mut self) -> Result<f32, DecodeError> {
+        if self.offset + 4 > self.input.len() {
+            return Err(DecodeError::InsufficientData);
+        }
+        let value =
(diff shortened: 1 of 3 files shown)


fasterthanlime:
> but it is technically a breaking change.

Oh that's nothing - `facet-msgpack` will have to be folded into the `facet-format` ecosystem too, at some point — to support all the same features as the other formats. We'll have to make sure we don't regress there!

Issue #1394, opened 2025-12-23
facet-format-postcard: Tier-2 JIT support for HashMap/BTreeMap


fasterthanlime:
## Summary

The positional struct JIT compiler for postcard doesn't yet support map types (HashMap, BTreeMap).

## Current Status

- 4 tests failing: `btreemap_empty`, `btreemap_ordered`, `hashmap_empty`, `hashmap_single`
- Returns `None` from compilation, falling back to lower tiers

## Implementation Notes

Postcard encodes maps as:
1. Length prefix (varint)
2. Key-value pairs in sequence

The positional compiler needs to:
1. Read the length prefix
2. Loop over key-value pairs
3. Insert into the map

This is similar to list deserialization but with key-value parsing.

## Related

Part of the Tier-2 positional struct compilation work for postcard.

fasterthanlime:
Done in #1400

PR #1400, opened 2025-12-23
facet-format-postcard: Add Tier-2 JIT support for HashMap/BTreeMap

fasterthanlime:
## Summary

Implemented Tier-2 JIT compilation support for HashMap and BTreeMap in postcard format. Maps use the same length-prefix encoding as sequences (varint count followed by key-value pairs), so the implementation reuses sequence state management logic.

All 4 previously failing map tests now pass:
- `collections::btreemap_empty::tier2_format_jit`
- `collections::btreemap_ordered::tier2_format_jit`
- `collections::hashmap_empty::tier2_format_jit`
- `collections::hashmap_single::tier2_format_jit`

## Changes

**facet-format-postcard/src/jit/format.rs**: Implemented 5 map-related JIT format methods in `PostcardJitFormat`:
- `emit_map_begin`: Reads varint length prefix and stores count in state
- `emit_map_is_end`: Checks if remaining count == 0
- `emit_map_read_key`: Delegates to `emit_parse_string`
- `emit_map_kv_sep`: No-op (postcard has no separators)
- `emit_map_next`: Decrements remaining count with underflow check

**facet-format/src/jit/format_compiler.rs**: Added Map support to positional struct compiler:
- Added `Map(&'static Shape)` variant to `PositionalFieldKind` enum
- Updated `classify_positional_field` to recognize maps
- Added Map field handling in two match statements (main loop + Option inner types)

### DIFF
--- a/facet-format-json/tests/jit_deserialize.rs
+++ b/facet-format-json/tests/jit_deserialize.rs
@@ -4,6 +4,9 @@ use facet::Facet;
 use facet_format::jit;
 use facet_format_json::JsonParser;
 
+#[cfg(feature = "jit")]
+use facet_format_json::JsonJitFormat;
+
 #[derive(Debug, PartialEq, Facet)]
 struct SimpleStruct {
     name: String,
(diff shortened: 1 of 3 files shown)


fasterthanlime:
Closes #1394

PR #1413, opened 2025-12-23
facet-xml: attributes reserved to namespace definition are ignored while deserializing

mattiapenati:
XML namespace definitions are reported as unknown attributes if the attribute `deny_unknown_fields` is used. I think that the proper behavior is to use these attributes only for the namespace resolution and to skip them in the deserialization step.

Another issue of the current behavior is that structure defined like follow cannot be deserialized from the output generate from serializer:
```rust
#[derive(Facet)]
#[facet(deny_unknown_fields, xml::ns_all = "http://example.com/ns")]
struct NamespacedRoot {
    #[facet(xml::element)]
    item: String,
}
```

This pull request fix the current behavior, attributes reserved for namespace definition are ignored in deserialization step. A small test is defined to check the correctness of the implementation.

### DIFF
--- a/facet-xml/src/deserialize.rs
+++ b/facet-xml/src/deserialize.rs
@@ -153,6 +153,11 @@ fn get_list_item_shape(shape: &facet_core::Shape) -> Option<&'static facet_core:
     }
 }
 
+/// Check if the attribute is reserved for XML namespace
+fn is_xml_namespace_attribute(name: &QName) -> bool {
+    name.matches_exact("xmlns", None) || name.namespace.as_deref() == Some("xmlns")
+}
+
 // ============================================================================
 // Public API
 // ============================================================================
@@ -715,6 +720,11 @@ impl<'input> EventCollector<'input> {
                 Some(uri) => QName::with_ns(uri, local),
                 None => QName::local(local),
             };
+
+            // Ignore attributes reserved for XML namespace declarations
+            if is_xml_namespace_attribute(&qname) {
+                continue;
+            }
 
             let value = attr
                 .unescape_value()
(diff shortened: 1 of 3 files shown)


fasterthanlime:
Some fresh test failures for you 🫡

Let me know if you need help or if you've got it!

mattiapenati:
Oh no! The reason for the test failures is: all the structures representing the root node in `real_world.rs` contain the field `xmlns`. In particular, `test_xhtml` checks the value of this field, while in all the other tests such field is ignored.

Looking at the outcome of these tests, I have some doubts and I don't know which behavior should be implemented. The options are:

1. Capture all the attributes, even the attributes related to namespace definitions (current implementation). In this case, users can capture namespace definitions using struct fields, but there are at least two different issues:

   - The one presented in this PR: `deny_unknown_fields` cannot be easily used with namespaces.
   - Deserializing equivalent XML documents with different namespace prefixes can produce different outcomes (is it an issue or a feature?). For example:
   ```rust
   #[derive(Facet, Debug, PartialEq)]
   #[facet(rename = "svg")]
   struct Svg {
       #[facet(default, xml::attribute)]
       xmlns: Option<String>,
   }

   assert_eq!(
       Svg { xmlns: Some("http://www.w3.org/2000/svg".to_string()) },
       xml::from_str(r#"<svg xmlns="http://www.w3.org/2000/svg" />"#).unwrap(),
   );

   assert_eq!(
       Svg { xmlns: None },
       xml::from_str(r#"<z:svg xmlns:z="http://www.w3.org/2000/svg" />"#).unwrap(),
   );
   ```
   
2. Ignore all the attributes related to namespace definitions (this PR implementation). In this case, users cannot capture namespace definitions. I don't think it is a regression, because namespaces are related only to the representation of data. Should users be able to distinguish between equivalent but different representations?

3. Capture all the attributes, but errors for unknown fields ascribable to namespace definitions are not reported. This is the most complicated to implement. Is such complexity required?

I'm inclined toward option 2 for its simplicity and adherence to XML semantics. However, I'd appreciate feedback from you on which approach best aligns with the library's design philosophy.

fasterthanlime:
Okay I think I understand better now. I'm on board with option 2.

PR #1481, opened 2025-12-26
facet-xml: check if attribute is namespace definition before namespace resolution

mattiapenati:
This pull request fix how XML namespace definitions are handled during XML deserialization:

* Moved the check for ignoring XML namespace definitions before namespace resolution.
* Updated the `is_xml_namespace_attribute` function to work with `quick_xml::name::QName`.
* Added a new test to verify that attributes with namespace prefixes are properly ignored.


fasterthanlime:
I probably told you that last time too but: facet-format-xml is the future, facet-xml is eventually getting deprecated, we should make some effort to port things to facet-format-xml while it's not at parity yet.

I'll do my best to have parity when switching them but.. I'd hate to see fixes like that go to waste!

PR #1554, opened 2026-01-01
refactor(facet-error): error::source for struct variants

cestef:
This fixes the `error::source` attribute handling and remove legacy directives in `facet-macros-impl`.

Previously, we only checked "variant-level" attributes and this only worked for single field tuple variants. this meant `#[facet(error::source)]` on struct variant fields or multi-field variants was silently ignored.

all error attributes should now work on any field in any variant kind (struct, tuple, unit?).
  
 **Changes** 

- Added `@if_field_attr(ns::key)` directive to find and bind fields with specific attributes
- Added `@variant_pattern_only(ns::key)` directive for selective pattern binding (only binds annotated fields, avoiding unused variable warnings)
- Updated `facet-error` template to use field-level attribute queries instead of variant-level
- Fixed clippy warnings for unreachable code and unused variables in generated code
- Added `#[allow(unused_variables)]` to Display impl where doc comments may not reference all fields

### DIFF
--- a/facet-error/src/lib.rs
+++ b/facet-error/src/lib.rs
@@ -92,6 +92,7 @@ macro_rules! __facet_invoke {
 
                         // Display impl - use doc comments as format strings
                         impl ::core::fmt::Display for @Self {
+                            #[allow(unused_variables)] // not all fields used in format strings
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Good stuff, tysm!

PR #1558, opened 2026-01-01
Fix deserialization of attributes when an element with the same name is present

mattiapenati:
The current implementation of `facet-xml` incorrectly deserializes a missing attribute from an element with the same name, as shown in the following snippet:

```rust
#[derive(Debug, Facet, PartialEq)]
struct Root {
    #[facet(xml::attribute)]
    id: String,
}

assert_eq!(
    facet_xml::from_str::<Root>(r#"<Root><id>value</id></Root>"#).unwrap(),
    Root {
        id: "value".to_string()
    }
);
```

This PR fixes this issue. Additionally, it enables the ability to deserialize both an attribute and an element with the same name without confusion.

### DIFF
--- a/facet-format/src/deserializer.rs
+++ b/facet-format/src/deserializer.rs
@@ -614,6 +614,11 @@ where
         location: FieldLocationHint,
         ns_all: Option<&str>,
     ) -> bool {
+        // === XML/HTML: Fields with xml::attribute match only attributes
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Thanks! Expect a few regressions as the old facet-xml was retired and facet-format-xml was renamed to facet-xml — hopefully not too many!

PR #1559, opened 2026-01-01
fix(soundness): make OxRef::new and OxMut::new unsafe

### LINKED ISSUE #1555: Distinguish pointer-like vs reference-like APIs
As of 0.41, there isn't a clear enough distinction between facet data structures that are pointer-like (require unsafe block before invoking a vtable pointer) vs reference-like (okay to invoke vtable pointers from safe code).

- `PtrConst` seems intended to be pointer-like because it is possible to construct one safely from an arbitrary pointer using `PtrConst::new`/`new_sized`/`new_wide`.

- `OxPtrConst` seems intended to be pointer-like because turning it into `&T` using `OxPtrConst::get` is unsafe and lists the correct set of preconditions.

- `Peek` seems intended to be reference-like because `Peek::partial_eq` is a safe function.

Without a categorical distinction between these though, various APIs unsoundly convert one into the other. For example:

```rust
use facet::{OxRef, PtrConst, Shape, ShapeBuilder, VTableDirect};

#[derive(PartialEq)]
struct MyType(String);

impl MyType {
    const VTABLE: VTableDirect = VTableDirect::builder_for::<Self>()
        .partial_eq(Self::eq)
        .build();
    const SHAPE: Shape = ShapeBuilder::for_sized::<Self>("MyType")
        .vtable_direct(&Self::VTABLE)
        .build();
}

fn main() {
    let shape = &MyType::SHAPE;
    let a = PtrConst::new(0x1111111 as *const MyType);
    let b = PtrConst::new(0x2222222 as *const MyType);
    let _ = OxRef::new(a, shape) == OxRef::new(b, shape);
}
```

```console
Segmentation fault (core dumped)
```

fasterthanlime:
## Summary

Fixes #1555 - Distinguish pointer-like vs reference-like APIs

Before this change, `OxRef::new` was **safe** but accepted arbitrary pointers, allowing use-after-free through safe code like `PartialEq`. This was unsound because safe Rust code could crash by dereferencing invalid pointers.

## Changes

- **Make `OxRef::new` and `OxMut::new` unsafe** with proper safety documentation
- **Add safe constructors**: `OxRef::from_ref` and `OxMut::from_mut` that capture lifetime from actual references
- **Update module documentation** to clearly distinguish:
  - Pointer-like types (`OxPtrConst`, `OxPtrMut`, `OxPtrUninit`) - no validity guarantees
  - Reference-like types (`OxRef<'a>`, `OxMut<'a>`) - **must guarantee** data is valid for `'a`
- **Update all call sites** in facet-core, facet-reflect, and facet-macros-impl
- **Add compile-fail regression test** that verifies the original unsound code is now rejected

## Test plan

- [x] `cargo check --workspace` passes
- [x] `cargo nextest run -p facet-core -p facet-reflect -p facet-json` passes (758 tests)
- [x] Regression test `test_oxref_unsound_from_raw_ptr` verifies unsound code fails to compile

### DIFF
--- a/facet-core/src/types/builtins.rs
+++ b/facet-core/src/types/builtins.rs
@@ -33,7 +33,7 @@
 //!
 //! # Comparison with `Peek`
 //!
-//! [`Peek<'mem, 'facet>`][crate::Peek] in `facet-reflect` follows the same pattern:
+//! `Peek<'mem, 'facet>` in `facet-reflect` follows the same pattern:
 //! - `Peek::new(&'mem T)` - safe, captures lifetime from reference
 //! - `Peek::unchecked_new(PtrConst, Shape)` - unsafe, for internal use
 


facet-core/src/types/builtins.rs
```diff
@@ -452,7 +521,9 @@ impl Ox<'static> {
     /// Take ownership of a boxed value.
     pub fn from_boxed<T>(boxed: Box<T>, shape: &'static Shape) -> Self {
         let ptr = Box::into_raw(boxed);
-        Ox::Owned(OxMut::new(PtrMut::new(ptr), shape))
+        // SAFETY: We just created this pointer from Box::into_raw, so it's valid
+        // and we have exclusive ownership (hence 'static lifetime is fine).
+        Ox::Owned(unsafe { OxMut::new(PtrMut::new(ptr), shape) })
```
dtolnay, line 526: Not sure about this one. `OxMut::new` requires _"The shape correctly describes the pointed-to type"_ which is not known here. This would need to be an unsafe fn to trust the caller's shape.

PR #1567, opened 2026-01-02
facet-kdl: Support primitive types directly with kdl::child attribute

### LINKED ISSUE #1560: facet-kdl: Support primitive types directly with kdl::child attribute
## Summary

When using `#[facet(kdl::child)]` on a primitive field like `bool`, `u64`, or `String`, facet-kdl should automatically extract the first argument from the child node.

## Current Behavior

Currently, to parse KDL like:

```kdl
config {
    enabled #true
    timeout 30
}
```

You need wrapper structs:

```rust
#[derive(Facet)]
struct SingleValue<T> {
    #[facet(kdl::argument)]
    pub value: T,
}

#[derive(Facet)]
struct Config {
    #[facet(kdl::child)]
    enabled: Option<SingleValue<bool>>,
    
    #[facet(kdl::child)]
    timeout: Option<SingleValue<u64>>,
}
```

And then access values via `config.enabled.map(|v| v.value)`.

## Expected Behavior

It would be much more ergonomic to write:

```rust
#[derive(Facet)]
struct Config {
    #[facet(kdl::child, default = true)]
    enabled: bool,
    
    #[facet(kdl::child, default = 30)]
    timeout: u64,
}
```

And have facet-kdl automatically extract the argument from nodes like `enabled #true` or `timeout 30`.

## Rationale

This pattern is extremely common in KDL configs. The current requirement for wrapper types adds boilerplate and makes the Rust types harder to work with (especially when you want direct field access rather than going through `.value`).

The same logic that works for `#[facet(kdl::argument)]` on a primitive inside a struct should work when `#[facet(kdl::child)]` targets a primitive - it should look for a child node with that name and extract its first argument.

fasterthanlime:
## Summary

When using `#[facet(kdl::child)]` on a primitive field like `bool`, `u64`, or `String`, facet-kdl now automatically extracts the first argument from the child node. This eliminates the need for wrapper structs that were previously required.

## Before

```rust
#[derive(Facet)]
struct SingleValue<T> {
    #[facet(kdl::argument)]
    pub value: T,
}

#[derive(Facet)]
struct Config {
    #[facet(kdl::child)]
    enabled: Option<SingleValue<bool>>,
    
    #[facet(kdl::child)]
    timeout: Option<SingleValue<u64>>,
}
// Access via config.enabled.map(|v| v.value)
```

## After

```rust
#[derive(Facet)]
struct Config {
    #[facet(kdl::child, default = true)]
    enabled: bool,
    
    #[facet(kdl::child, default = 30)]
    timeout: u64,
}
// Direct access: config.enabled, config.timeout
```

Both parse KDL like:
```kdl
enabled #true
timeout 60
```

## Changes

- Added `is_scalar_compatible_type()` helper to detect types that can be deserialized from a scalar value (handles `Scalar`, `Option<Scalar>`, `Pointer<Scalar>`, and transparent wrappers)
- Added `deserialize_kdl_child_scalar()` method that extracts the `_arg` field from a KDL node struct and deserializes it into the target scalar type
- Modified struct field deserialization in `deserialize_struct()` to detect `kdl::child` fields with scalar-compatible types and use the special handling
- Added comprehensive tests for primitive `kdl::child` fields, including `Option<primitive>`, nested structs, mixed fields, and roundtrip serialization

## Test plan

- [x] All existing facet-kdl tests pass (185 tests)
- [x] All full test suite passes (2457 tests)
- [x] New tests cover primitive types (`bool`, `u64`, `String`)
- [x] New tests cover `Option<primitive>` types
- [x] New tests cover mixed primitive and struct children
- [x] New tests cover nested primitive children in struct children
- [x] Roundtrip serialization works correctly

Fixes #1560

### DIFF
--- a/facet-format/src/deserializer.rs
+++ b/facet-format/src/deserializer.rs
@@ -592,6 +592,114 @@ where
     ///
     /// This implements format-specific field matching for XML and KDL:
     ///
+    /// Check if a type can be deserialized directly from a scalar value.
+    ///
+    /// This is used for KDL child nodes that contain only a single argument value,
+    /// allowing `#[facet(kdl::child)]` to work on primitive types like `bool`, `u64`, `String`.
+    fn is_scalar_compatible_type(shape: &facet_core::Shape) -> bool {
+        match &shape.def {
+            Def::Scalar => true,
+            Def::Option(opt) => Self::is_scalar_compatible_type(opt.t),
+            Def::Pointer(ptr) => ptr.pointee.is_some_and(Self::is_scalar_compatible_type),
+            _ => {
+                // Also check for transparent wrappers (newtypes)
+                if !matches!(
+                    &shape.def,
+                    Def::List(_) | Def::Map(_) | Def::Set(_) | Def::Array(_)
+                ) && let Some(inner) = shape.inner
+                {
+                    return Self::is_scalar_compatible_type(inner);
+                }
+                false
+            }
+        }
+    }
+
+    /// Deserialize a KDL child node that contains only a scalar argument into a scalar type.
+    ///
+    /// When a KDL node like `enabled #true` is parsed, it generates:
+    /// - `StructStart`
+    /// - `FieldKey("_node_name", Argument)` → `Scalar("enabled")`
+    /// - `FieldKey("_arg", Argument)` → `Scalar(true)`  ← this is what we want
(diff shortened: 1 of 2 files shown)


dougEfresh:
nice! 
when release?

PR #1569, opened 2026-01-02
Fix memory leaks and use-after-free in Partial cleanup

### LINKED ISSUE #1568: facet-html panics (SIGABRT) during error cleanup instead of returning parse error
# facet-html panics during error cleanup

## Description

`facet-html` panics with `SIGABRT` during error cleanup instead of returning the parse error cleanly. When the HTML parser encounters a type mismatch, it crashes in the `Drop` implementation of `Partial<true>` with undefined behavior.

## Error Message

```
thread 'main' (236639874) panicked at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/alloc/src/raw_vec/mod.rs:532:30:
unsafe precondition(s) violated: Layout::from_size_align_unchecked requires 
that align is a power of 2 and the rounded-up allocation size does not exceed isize::MAX

This indicates a bug in the program. This Undefined Behavior check is optional, 
and cannot be relied on for safety.
```

Full stack trace shows the panic occurs during:
```
drop<true> at facet-reflect/src/partial/mod.rs:1100:23
drop_in_place<facet_reflect::partial::Partial<true>>
deserialize_struct_single_flatten<true, facet_html::parser::HtmlParser>
```

### Cargo.toml
```toml
[package]
name = "facet-html-bug"
version = "0.1.0"
edition = "2021"

[dependencies]
facet-html = { git = "https://github.com/facet-rs/facet", branch = "main" }
```

### src/main.rs
```rust
use facet_html as fhtml;

fn main() {
    // This HTML triggers the bug - generated by a markdown processor
    let html = include_str!("../failing.html");
    
    println!("Attempting to parse {} bytes of HTML...", html.len());
    match fhtml::from_str::<fhtml::elements::Html>(html) {
        Ok(_) => println!("✓ Parse succeeded"),
        Err(e) => {
            eprintln!("✗ Parse error: {}", e);
            eprintln!("Bug: should see this message, but process crashes during error cleanup");
        }
    }
    println!("If you see this, the bug is fixed!");
}
```

### failing.html
See attached file (9323 bytes). This is HTML generated by a markdown renderer that triggers the issue.

## Expected Behavior

The parser should return an error:
```
type mismatch: expected struct start, got Scalar(Str("Some features are implemented as separate helper binaries ("))
```

The program should print the error message and continue executing.

## Actual Behavior

The process crashes with `SIGABRT` during the `Drop` implementation when cleaning up the partial parse state. This makes it impossible to handle parse errors gracefully in production code.

## Environment

- facet-html: commit f42a5f59 (main branch)
- rustc: 1.91-aarch64-apple-darwin
- OS: macOS 14 (also reproducible on Linux)

## Impact

This bug prevents using facet-html in any error-handling context where malformed HTML might be encountered. The process crash cannot be caught, making the library unsuitable for production use with untrusted input.

## Workaround

None currently - any HTML that triggers this parse error will crash the process.

fasterthanlime:
## Summary

Fixes multiple memory safety issues in `facet-reflect`'s `Partial` cleanup code, discovered through fuzzing.

### Issues Fixed

1. **Stored frame ordering in `Partial::drop`** - Stored frames in deferred mode were processed in BTreeMap iteration order (alphabetical) instead of deepest-first. This caused parent frames to be cleaned up before children, invalidating child data pointers.

2. **`begin_map`/`begin_list` leak for BorrowedInPlace frames** - These functions called `deinit()` when converting a DynamicValue Scalar to Object/Array, but `deinit()` early-returns for BorrowedInPlace frames without dropping the existing value, causing leaks.

3. **Use-after-free in `deinit_for_replace`** - When dropping a BorrowedInPlace DynamicValue, the parent Object's HashMap entry still points to the now-garbage memory. When the parent Object drops, it tries to access freed memory. Fixed by reinitializing with `set_null` after dropping.

### Changes

- `facet-reflect/src/partial/mod.rs`: Fixed stored frame ordering, added `set_null` reinitialization for BorrowedInPlace DynamicValue
- `facet-reflect/src/partial/partial_api/maps.rs`: Use `deinit_for_replace()` instead of `deinit()`
- `facet-reflect/src/partial/partial_api/lists.rs`: Use `deinit_for_replace()` instead of `deinit()`
- `facet-reflect/tests/partial/option_leak.rs`: Added regression tests for all issues

### Testing

- All 431 facet-reflect tests pass
- All 34 facet-html tests pass
- Fuzzer ran 2.8M iterations without finding new issues
- All previously-crashing/leaking fuzzer artifacts now pass

Fixes #1568

### DIFF
--- a/facet-reflect/src/partial/mod.rs
+++ b/facet-reflect/src/partial/mod.rs
@@ -567,26 +567,12 @@ impl Frame {
             return;
         }
 
-        // For Field frames with simple trackers (Scalar, List, Map, etc.), we must NOT
-        // drop the value directly. The parent struct's deinit will drop all its
-        // initialized fields, so dropping here would cause double-free.
-        //
-        // However, Field frames with composite trackers (Struct, Array, Enum) must still
-        // process their internal iset to drop individually initialized sub-fields,
-        // since the parent only tracks whether the whole field is initialized.
-        if matches!(self.ownership, FrameOwnership::Field { .. }) {
-            match &self.tracker {
-                Tracker::Struct { .. } | Tracker::Array { .. } | Tracker::Enum { .. } => {
-                    // Continue to process iset below - don't return early
-                }
-                _ => {
-                    // Skip dropping for simple trackers
-                    self.is_init = false;
-                    self.tracker = Tracker::Scalar;
-                    return;
-                }
-            }
-        }
+        // Field frames are responsible for their value during cleanup.
+        // The ownership model ensures no double-free:
+        // - begin_field: parent's iset[idx] is cleared (parent relinquishes responsibility)
+        // - end: parent's iset[idx] is set (parent reclaims responsibility), frame is popped
+        // So if Field frame is still on stack during cleanup, parent's iset[idx] is false,
+        // meaning the parent won't drop this field - the Field frame must do it.
 
         match &self.tracker {
             Tracker::Scalar => {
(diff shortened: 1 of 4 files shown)


fasterthanlime:
## Root Cause Analysis: `ManagedElsewhere` Conflates Two Different Ownership Semantics

### The Problem

`FrameOwnership::ManagedElsewhere` is currently used for two fundamentally different scenarios:

1. **Temporary buffers** (safe to drop on overwrite)
   - Created by `begin_key()`, `begin_value()` 
   - Parent has `MapInsertState` with explicit `key_initialized`/`value_initialized` tracking
   - When child is popped, `is_init` is propagated to parent's tracking
   - Parent's cleanup respects this tracking

2. **Existing collection entries** (NOT safe to drop on overwrite)
   - Created by `begin_object_entry()` on existing key, `begin_some()` re-entry
   - Points directly into the collection's storage (e.g., a Value inside an IndexMap)
   - Parent collection has NO per-entry initialization tracking
   - Collection always drops all its entries on cleanup

### The Leak (without fix)

```
begin_key() → parse_from_str("hello") → parse_from_str("world")
```

The second `parse_from_str` calls `deinit()` which skips dropping for `ManagedElsewhere` → "hello" leaks.

### The Double-Free (with naive fix)

Adding explicit drops before overwrite:
```rust
if matches!(frame.ownership, FrameOwnership::ManagedElsewhere) && frame.is_init {
    unsafe { shape.call_drop_in_place(frame.data.assume_init()) };
}
```

This breaks for existing entries:
```
begin_object_entry("key") → set(42i32) → parse_from_str("hello")
```

1. `parse_from_str` drops the `Value::Number(42)` inside the Value object
2. Parse fails or Partial is dropped
3. Parent frame's `deinit()` drops the Value object  
4. `Value::drop` iterates all entries, tries to drop already-freed entry
5. **Double-free!**

### Proposed Solution

Split `ManagedElsewhere` into two variants:

```rust
pub(crate) enum FrameOwnership {
    Owned,
    Field { field_idx: usize },
    
    /// Temporary buffer tracked by parent's MapInsertState.
    /// Safe to drop on overwrite - parent will check key_initialized/value_initialized.
    TrackedBuffer,
    
    /// Pointer into existing collection entry (Value object, Option inner, etc.)
    /// NOT safe to drop without reinitializing - parent has no per-entry tracking.
    BorrowedInPlace,
}
```

Then:
- `begin_key()`, `begin_value()` → `TrackedBuffer`
- `begin_object_entry()` on existing, `begin_some()` re-entry → `BorrowedInPlace`

For `TrackedBuffer`: Safe to drop before overwrite (parent respects `is_init` propagation)
For `BorrowedInPlace`: Must NOT drop unless we successfully reinitialize (accept leak on double-parse as lesser evil than corruption)

### Alternative: Runtime Check

Instead of new variants, check parent's tracker state:
- If parent has `MapInsertState::PushingKey/PushingValue` → safe to drop
- Otherwise → not safe

This avoids changing the enum but requires accessing parent frame during child operations.

PR #1585, opened 2026-01-03
Add `facet_no_doc` cfg for global doc string stripping

### LINKED ISSUE #1583: Find ways to disable documentations
For some reason I saw this in rapace:

Image

Those are not necessary reflection information which contributes to a great part in binary size. We should have a switch to disable it

fasterthanlime:
## Summary

Adds a cfg-based mechanism to globally disable doc strings in facet reflection metadata, similar to tracing's `release_max_level_off` pattern. This allows users to strip docs from release builds without patching dependencies.

Fixes #1583

## Usage

Add to `.cargo/config.toml`:

```toml
[profile.release]
rustflags = ["--cfg", "facet_no_doc"]
```

## How it works

- Feature `doc` disabled → No docs (checked at proc-macro compile time)
- Feature `doc` enabled + `facet_no_doc` cfg → No docs (checked at user crate compile time)  
- Feature `doc` enabled + no cfg → Docs included (default behavior)

## Changes

- Modified `process_struct.rs` to emit conditional cfg blocks for field and container docs
- Modified `process_enum.rs` to emit conditional cfg blocks for container and variant docs

## Test plan

- [x] `cargo check --all-features --all-targets` passes
- [x] `cargo nextest run --all-features` passes (pre-existing XML failures unrelated)
- [x] `RUSTFLAGS='--cfg facet_no_doc' cargo nextest run -p facet struct_doc_comment` fails with `left: []` vs `right: [" yes"]` - confirming docs are stripped

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,6 +134,11 @@ ulid = "^1.2.1"
 unsynn = "^0.3.0"
 uuid = "^1.19.0"
 
+# Allow the facet_no_doc cfg in generated code
+# See: https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(facet_no_doc)"] }
+
 # Fuzzing profile for cargo-bolero
 [profile.fuzz]
 inherits = "dev"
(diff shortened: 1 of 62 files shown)


fasterthanlime:
Well it'll now generate slightly more code at compile time... but it'll all get eliminated if the cfg is enabled.

Issue #1656, opened 2026-01-04
Wrong deserialization of HTML element named extra


mattiapenati:
If an element named `extra` is present then it is deserialized as a map, since it has the same name of the field used capture all the unknwon fields.

The following code can be used to reproduce the problem:
```rust
let html = facet_html::to_string(
    &facet_html::from_str::<facet_html_dom::Html>(r#"<html><extra>zzz</extra></html>"#)
        .unwrap(),
)
.unwrap();
println!("{html}");
```

The output is `<html _tag="extra" _text="zzz"></html>`, the expected one is `<html></html>` (because the unknown elements are not captured).

PR #1657, opened 2026-01-04
Fix for issue #1656

mattiapenati:
This PR contains the fix for the issue #1656. Currently nested flattened map are not properly handled for elements.

### DIFF
--- a/facet-format/src/deserializer.rs
+++ b/facet-format/src/deserializer.rs
@@ -2087,15 +2087,22 @@ where
                         }
                         if let Some((inner_fields, inner_set)) = flatten_info[flatten_idx].as_mut()
                         {
-                            let inner_match = inner_fields.iter().enumerate().find(|(_, f)| {
-                                Self::field_matches_with_namespace(
-                                    f,
-                                    key.name.as_ref(),
-                                    key.namespace.as_deref(),
-                                    key.location,
-                                    ns_all,
-                                )
-                            });
+                            let inner_match =
+                                inner_fields.iter().enumerate().find(|(inner_idx, f)| {
+                                    // Nested flattened map is handled separately, even if its name
+                                    // matches the key name it must be skipped.
+                                    let is_flatten_map =
+                                        Some((flatten_idx, Some(*inner_idx))) == flatten_map_idx;
+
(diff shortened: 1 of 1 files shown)


fasterthanlime:
I'm not very happy with this fix because it feels like it's just disabling something. Like it's disabling invalid behavior without implementing valid behavior. So I think the question is: what is the ideal behavior and then let's just implement that.

mattiapenati:
Good point. After reviewing my implementation more carefully, I found that it breaks the deserialization of unknown attributes. I've added a new test and I'm looking for a better implementation.

mattiapenati:
I've changed the implementation and written a test to catch any regressions with unknown arguments. I find the new version clearer, I hope it works better for you too.

facet-html/tests/minimal_repro.rs
```diff
@@ -634,3 +634,19 @@ Line 3</textarea>"#;
         serialized
     );
 }
+
+#[test]
+fn issue_1656_deserialization_of_extra_element() {
+    let input = r#"<html><extra>value</extra></html>"#;
+    let html: facet_html_dom::Html = facet_html::from_str(input).unwrap();
+    let output = facet_html::to_string(&html).unwrap();
+    assert_eq!(output, r#"<html></html>"#);
```
fasterthanlime, line 643: I don't understand why the extra tags would be stripped from this because facet-html-dom is supposed to support arbitrary tags, just like HTML and it's supposed to be able to roundtrip them so I don't get it. 

dodeca uses arborium for syntax highlighting, which emits `<a-kw/>` tags for example and it definitely needs to be able to parse them and re-emit them.

mattiapenati: The `Html` type captures only `head` and `body` elements. There isn't any `CustomElement` to deserialize the `extra` tag.

fasterthanlime: Ah, mh. I missed that. That makes sense. I was thinking in terms of HTML5 is tag soup and anything is fair game anywhere. But facet-html-dom the crate is trying to make sense of that. So... that makes sense.

PR #1667, opened 2026-01-06
refactor(bytestring): simplify ByteString impl with vtable_direct! macro

gaetschwartz:
- Replace manual VTableIndirect with vtable_direct! macro
- Remove redundant parse, display, and partial_eq implementations
- Simplify try_from signature to use direct pointer type
- Add builder_shape for String to enable built-in parsing


fasterthanlime (approved):
LGTM, thanks!

PR #1705, opened 2026-01-07
fix: prevent exponential variance computation for recursive types

### LINKED ISSUE #1704: Variance computation takes exponentially long for multiple-recursive types
```rust
use facet::{CovariantPeek, Facet, Peek};

#[derive(Facet)]
struct Node(&'static Node, &'static Node, &'static Node, &'static Node);

static NODE: Node = Node(&NODE, &NODE, &NODE, &NODE);

fn main() {
    let begin = std::time::Instant::now();
    let _ = CovariantPeek::new(Peek::new(&NODE));
    println!("{:?}", begin.elapsed());
}
```

```console
$ cargo run --release
30.689778762s
```

fasterthanlime:
## Summary

Fixes #1704 where variance computation took ~30 seconds for types with multiple recursive fields like `Node(&'static Node, &'static Node, &'static Node, &'static Node)`. The algorithm was visiting the same shape 4^depth times, causing exponential blowup.

## Changes

- **Add visited set tracking** using `ConstTypeId` to detect cycles during variance computation
  - When a cycle is detected, return `Covariant` (neutral element - cycles don't add new variance info)
  - This prevents exponential blowup and correctly handles recursive types
- **Add early termination** when `Invariant` is encountered (anything combined with Invariant is Invariant)
- **Respect custom variance functions** before falling back to Def-based lookup
- **Add explicit variance propagation** to container types that were missing it:
  - Arrays, slices, HashSet, BTreeSet, HashMap, BTreeMap, Result

## Key insight

Recursive types like `Node(&'static Node, ...)` are actually **covariant**, not invariant. The old code incorrectly returned Invariant when hitting the depth limit. The fix correctly identifies cycles and returns the proper variance based on the non-cyclic parts of the type structure.

## Testing

Added comprehensive tests in `facet/tests/variance_stack_overflow.rs`:
- Multi-recursive types (exact reproduction of #1704)
- Recursive types with invariant fields
- Early termination on invariant
- Mutually recursive types

All tests verify completion in <100ms (before fix: ~30 seconds).

### DIFF
--- a/facet/tests/variance_stack_overflow.rs
+++ b/facet/tests/variance_stack_overflow.rs
@@ -1,4 +1,10 @@
 //! Test variance computation with recursive types.
+//!
+//! These tests verify:
+//! 1. Cycle detection prevents infinite recursion
+//! 2. Exponential blowup is prevented for multi-recursive types
+//! 3. Variance is computed correctly for recursive types
+//! 4. Early termination works when invariant is detected
 
 use facet::{Facet, Variance};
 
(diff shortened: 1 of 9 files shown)


facet/tests/variance_stack_overflow.rs
```diff
@@ -91,3 +98,173 @@ fn test_struct_with_mut_ptr_is_invariant() {
...
+#[test]
+fn test_early_termination_on_invariant() {
+    let start = std::time::Instant::now();
+    let shape = EarlyTermination::SHAPE;
+    let variance = shape.computed_variance();
+    let elapsed = start.elapsed();
+
+    // Should terminate early when *mut i32 is encountered
+    assert!(
+        elapsed.as_millis() < 100,
```
dtolnay, line 190: Timing-sensitive tests should get a carveout for Miri.

```console
---- test_mutually_recursive_types stdout ----

thread 'test_mutually_recursive_types' (1528187) panicked at facet/tests/variance_stack_overflow.rs:228:5:
Mutually recursive variance took 2.814776383s, expected < 100ms
note: in Miri, you may have to set `MIRIFLAGS=-Zmiri-env-forward=RUST_BACKTRACE` for the environment variable to have an effect
```

```suggestion
        cfg!(miri) || elapsed.as_millis() < 100,
```

Issue #1708, opened 2026-01-07
Consider adding a bivariant variance


dtolnay:
Currently `facet_core::Variance` can only be Covariant, Contravariant, and Invariant.

Some (most?) types are both covariant (<code>T&lt;'long&gt; <b>&lt;:</b> T&lt;'short&gt;</code>) and contravariant (<code>T&lt;'short&gt; <b>&lt;:</b> T&lt;'long&gt;</code>). This includes all types containing none of: lifetimes, references, or interior mutability. For example:

```rust
type T<'a> = <std::str::Bytes<'a> as Iterator>::Item;

fn covariant<'short, 'long: 'short>(value: T<'long>) -> T<'short> {
    value
}

fn contravariant<'short, 'long: 'short>(value: T<'short>) -> T<'long> {
    value
}
```

With only 3 choices of variance, facet cannot use the variance of `T` to accurately determine a variance for `fn(T)`. If `T` is "covariant" (meaning either covariant or bivariant) then `fn(T)` may be either contravariant or "covariant". (Currently facet falls back to invariant in this situation.)

fasterthanlime:
Hopefully #1710 does the trick

PR #1710, opened 2026-01-07
refactor(variance): replace function-based variance with declarative model

fasterthanlime:
## Summary

Replace the `Shape.variance` field from a function pointer to a new `VarianceDesc` struct. This eliminates a class of bugs where custom variance functions could break cycle detection by creating fresh visited sets.

The motivation is issue #1708 - with the old design, types like `&T` and `fn(...)` had custom variance functions (`ref_variance`, `fn_ptr_variance`) that recursively computed variance, but they didn't properly participate in cycle detection because each created its own visited set.

### New types (`facet-core/src/types/variance.rs`)
- `VarianceDesc { base: Variance, deps: &'static [VarianceDep] }` - declarative variance description
- `VarianceDep { position: VariancePosition, shape: &'static Shape }` - a single dependency
- `VariancePosition::Covariant | Contravariant` - position transform for dependencies

### Updated variance computation (`facet-core/src/types/shape.rs`)
- `computed_variance_impl` now interprets `VarianceDesc`:
  1. Uses `base` variance as starting point
  2. Iterates over `deps`, computing and combining each dependency's variance (with position transform)
  3. Falls back to field walking for structs/enums when deps is empty (backward compatibility)

### Updated type implementations
- Invariant types (`*mut T`, `&mut T`, `Opaque<T>`) → `VarianceDesc::INVARIANT`
- Covariant propagating types (`&T`, `Box<T>`, `Vec<T>`, etc.) → `VarianceDesc` with `VarianceDep::covariant(T::SHAPE)`
- Multi-param types (`HashMap<K,V>`, `Result<T,E>`) → multiple covariant dependencies
- Function pointers → contravariant args + covariant return

### Updated derive macro
- `𝟋CV` constant changed from function pointer to `VarianceDesc::BIVARIANT`
- `𝟋VncD` alias added for `VarianceDesc` in prelude

## Test plan

- [x] All 2839 existing tests pass
- [x] Variance stack overflow tests continue to pass (cycle detection still works)
- [x] Bivariance tests pass

### DIFF
--- a/facet-core/src/impls/core/reference.rs
+++ b/facet-core/src/impls/core/reference.rs
@@ -5,7 +5,7 @@ use core::fmt;
 use crate::{
     Def, Facet, HashProxy, KnownPointer, OxPtrConst, OxPtrMut, PointerDef, PointerFlags,
     PointerType, PointerVTable, PtrConst, Shape, ShapeBuilder, Type, TypeNameOpts, TypeOpsIndirect,
-    TypeParam, VTableIndirect, ValuePointerType, Variance,
+    TypeParam, VTableIndirect, ValuePointerType, Variance, VarianceDep, VarianceDesc,
 };
 
 /// Type-erased type_name for &T - reads the pointee type from the shape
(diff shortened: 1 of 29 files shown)


facet-core/src/impls/core/reference.rs
```diff
@@ -300,7 +304,7 @@ unsafe impl<'a, T: ?Sized + Facet<'a>> Facet<'a> for &'a mut T {
             }])
             .inner(T::SHAPE)
             // &mut T is invariant in T
-            .variance(Variance::INVARIANT)
+            .variance(VarianceDesc::INVARIANT)
```
dtolnay, line 307: This is sound but more strict than necessary. A type like `&'a mut i32` is covariant in 'a.

```rust
type T<'a> = &'a mut i32;

fn covariant<'short, 'long: 'short>(value: T<'long>) -> T<'short> {
    value
}
```

The optimal behavior would be something like: if `T` is bivariant then `&mut T` is covariant, otherwise `&mut T` is invariant.

PR #1722, opened 2026-01-09
feat(facet-kdl): add from_reader function

dougEfresh:
- Add reader.rs module with from_reader function
- Add 'reader' feature as alias for 'std'
- Function reads entire input into memory then calls from_str

### DIFF
--- a/facet-json/src/parser.rs
+++ b/facet-json/src/parser.rs
@@ -634,9 +634,10 @@ impl<'de> FormatParser<'de> for JsonParser<'de> {
         // Handle the case where peek_event was called before capture_raw.
         // This happens when deserialize_option peeks to check for null.
         let start_offset = if let Some(event) = self.event_peek.take() {
-            let start = self.peek_start_offset.take().expect(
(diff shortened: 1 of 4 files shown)


fasterthanlime:
I feel like `from_reader` should only be exposed if the crate actually does support streaming deserialization (like the facet-json crate does), I'm inclined to reject this PR as-is (since streaming KDL parsing is not happening soon).

dougEfresh:
@fasterthanlime make sense.  Feel free to close. 

I was parsing from files and figured others would want to do the same, hence this PR.

fasterthanlime:
Note that due to unfortunate events, KDL support is being entirely removed from facet — #1727 

(People are welcome to maintain whatever they want in third-party repositories, I will just not spend time officially supporting it)

dougEfresh:
@fasterthanlime 
Whats the back story behind this? What did Kat do?

fasterthanlime:
> @fasterthanlime Whats the back story behind this? What did Kat do?

All I'm comfortable sharing is here: https://github.com/facet-rs/facet/issues/1727

PR #1723, opened 2026-01-09
feat(pretty): add proxy support for pretty-printing

gaetschwartz:
Fix pretty-printing for types using `#[facet(proxy = ...)]` and `#[facet(transparent)]`.

  Previously, `facet-pretty` did not check for proxy definitions when formatting values, causing types with proxies to fall through to the default case and print `unsupported peek variant: ⟨TypeName⟩`. Additionally, transparent wrapper types were not being unwrapped, so a type like `IntAsString(String)` would print as `IntAsString("value")` instead of just `"value"`.

  This brings `facet-pretty` in line with `facet-format`'s serialization behavior:
  - Add `innermost_peek()` call to unwrap transparent wrappers before formatting
  - Check `shape.proxy` for container-level proxies
  - Check `field.proxy()` for field-level proxies in struct/tuple fields

### DIFF
--- a/facet-pretty/src/printer.rs
+++ b/facet-pretty/src/printer.rs
@@ -1395,7 +1395,7 @@ impl PrettyPrinter {
 
         // Handle proxy types by converting to the proxy representation and formatting that
         if let Some(proxy_def) = shape.proxy {
-            return self.format_via_proxy_unified(
+            let result = self.format_via_proxy_unified(
                 value,
                 proxy_def,
                 out,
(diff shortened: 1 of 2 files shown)


fasterthanlime:
The new code (unfortunately) has to be duplicated into `format_peek_with_spans`, which is an entire /second/ set of formatting routines.

Ideally we'd merge those in the future (implement the non-span-aware format in terms of the span-aware format), that would make a lot more sense.

PR #1733, opened 2026-01-10
fix(facet-pretty): false cycle detection

gaetschwartz:
The cycle detection was flagging sibling fields as cycles when they pointed to the same memory address. 

Added a regression test to make sure this doesn't happen again.


fasterthanlime:
Huh was that also present in another PR? it shows zero changes now + conflicts + 4 commits behind.

gaetschwartz:
Oh I guess we did fix it in the previous proxy support one, closing this one 👍

Issue #1756, opened 2026-01-13
facet-xml: use `Attribute::decode_and_unscape_value` instead of `Attribute::unescape_value`


Erik1000:
Initially, I hit this error because one of my dependencies enabled the `encoding` feature of `quick_xml` which removes [`Attribute::unescape_value`](https://docs.rs/quick-xml/latest/quick_xml/events/attributes/struct.Attribute.html#method.unescape_value) for [_reasons_](https://github.com/tafia/quick-xml/issues/861#issuecomment-2855763810).

Anyways, `quick_xml` states that libraries should use  [`Attribute::decode_and_unescape_value`](https://docs.rs/quick-xml/latest/quick_xml/events/attributes/struct.Attribute.html#method.decode_and_unescape_value)` instead.

Related: https://github.com/tafia/quick-xml/issues/861

fasterthanlime:
Closed by #1757

PR #1757, opened 2026-01-13
facet-xml: use Attribute::decode_and_unscape_value instead of Attribute::unescape_value

Erik1000:


Erik1000:
Fixes #1756

fasterthanlime:
Well that's a weird one. Thanks for the fix!

PR #1768, opened 2026-01-14
feat(facet-core): add iddqd crate support

gaetschwartz:
Add Facet trait implementations for iddqd collection types:
- IdHashMap<T, S> - hash map with single key extracted from value
- IdOrdMap<T> - ordered map (requires std feature)
- BiHashMap<T, S> - bijective map with two keys per value
- TriHashMap<T, S> - trijective map with three keys per value

These types use Def::Set with SetVTable since they're conceptually "sets with extractable keys" - values are stored and keys are derived.

Also includes:
- Feature flag `iddqd` in facet-core and facet crates
- Documentation in ecosystem.md
- Format suite tests for JSON, TOML, YAML, XML, MsgPack, ASN.1

### DIFF
--- a/facet-asn1/tests/format_suite.rs
+++ b/facet-asn1/tests/format_suite.rs
@@ -500,6 +500,18 @@ impl FormatSuite for Asn1Slice {
         CaseSpec::skip("ASN.1 is a binary format, requires binary input not JSON strings")
     }
 
+    fn iddqd_id_ord_map() -> CaseSpec {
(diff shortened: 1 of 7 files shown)


fasterthanlime (changes requested):
Two bugs here:

  * major: Bi/Tri HashMaps only check key1, ignoring key2/key3 (maybe don't cover these if we don't have vtables sophisticated enough for them?)
  * minor: `type_name` only prints `T` type params when we have `T` and `S`

PR #1769, opened 2026-01-14
feat(facet-core): add chrono::Duration support

gaetschwartz:
Follows the same logic as their own serde impl: (de)serialize as (seconds, nanoseconds).


fasterthanlime (changes requested):
I'd like to see negative duration test cases please

PR #1770, opened 2026-01-14
fix(facet-macros): preserve spans for proxy and other attribute values

gaetschwartz:
Fixes span preservation for proxy, default, skip_serializing_if, invariants, and truthy attributes so IDE features like hover and go-to-definition work correctly on attribute values.

### DIFF
--- a/facet-macros-impl/src/process_struct.rs
+++ b/facet-macros-impl/src/process_struct.rs
@@ -98,6 +98,27 @@ impl<'a> TraitSources<'a> {
     }
 }
 
+/// Generate a phantom use statement that links the attribute span to its Attr variant.
+///
+/// This enables IDE hover documentation when users mouse over attribute names like `proxy`
+/// in `#[facet(proxy)]`. By generating `{ use facet::builtin::Attr::Proxy as _; }` where
+/// `Proxy` has the same span as the source `proxy`, the IDE can follow the connection and
+/// show the Attr variant's documentation.
+pub(crate) fn phantom_attr_use(
+    attr: &PFacetAttr,
+    facet_crate: &TokenStream,
+) -> Option<TokenStream> {
+    if !attr.is_builtin() {
+        return None;
+    }
+    let binding = attr.key_str();
+    let variant_name = RenameRule::PascalCase.apply(&binding);
+
+    // Create ident with the attribute's source span - this is the key to IDE hover!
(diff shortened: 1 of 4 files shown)


facet/Cargo.toml
```diff
@@ -135,6 +135,7 @@ facet-showcase = { path = "../facet-showcase" }
 facet-testhelpers = { path = "../facet-testhelpers" }
 num-complex = "0.4.6"
 owo-colors = "4.2.2"
+trybuild = "1.0.114"
```
fasterthanlime, line 138: we've been specifically avoiding trybuild in the past (we have `compile_tests` that just.. do not use it) but.. the benefits outweigh the cost for now.

facet-macros-impl/src/process_struct.rs
```diff
@@ -885,11 +885,8 @@ pub(crate) fn gen_field_from_pfield(
                         default_value = Some(DefaultKind::FromTrait);
                     } else {
                         // #[facet(default = expr)] - use custom expression
-                        // Parse `= expr` to get just the expr
-                        let args_str = args.to_string();
-                        let expr_str = args_str.trim_start_matches('=').trim();
-                        let expr: TokenStream = expr_str.parse().unwrap_or_else(|_| args.clone());
-                        default_value = Some(DefaultKind::Custom(expr));
```
fasterthanlime, line 892: wow that code looks suspicious... I guess it never was needed after the migration to the attr grammar parser?

facet-macros-impl/src/process_struct.rs
```diff
@@ -98,6 +98,27 @@ impl<'a> TraitSources<'a> {
     }
 }
 
+/// Generate a phantom use statement that links the attribute span to its Attr variant.
+///
+/// This enables IDE hover documentation when users mouse over attribute names like `proxy`
+/// in `#[facet(proxy)]`. By generating `{ use facet::builtin::Attr::Proxy as _; }` where
+/// `Proxy` has the same span as the source `proxy`, the IDE can follow the connection and
+/// show the Attr variant's documentation.
```
fasterthanlime, line 106: that's devilish. hopefully it doesn't impact compile times too much! the DX gain is awesome though

fasterthanlime:
Rebasing this to another branch

fasterthanlime:
Merged in #1779

Issue #1806, opened 2026-01-16
Python generate


vron:
Hi,

I will need to generate python types from facet, and would like to give it a shot to contribute something.

Before getting started I'd like to check if a PR to add the following would be accepted (@fasterthanlime )?

Structured similar to facet-typescript, but instead of generating typescript code generate python types.

I would suggest to generate tye types using non-total Typed dicts, as an example:

```python

class ExecLog(TypedDict, total=False):
    context: Required[list[str]
    level: Required["LogLevel"]

LogLevel = Literal["info"] | Literal["warn"] | Literal["err"]
```

I would suggest to always sort the output types alphabetically to ensure stable output and smaller diffs.

Does this sound good or do you not think it belongs in facet core?

vron:
I know I asked, but in the meantime did some work anyway - feel free to say "does not fit here" anyway

fasterthanlime:
I'm not against it! facet is good with cross-language stuff

PR #1807, opened 2026-01-16
Introduce facet-python for generating python types

vron:
A first verions structured similartly to the typescript crate. Partial solution to issue #1806 but not yet ready to be merged.

### DIFF
--- a/facet-python/src/lib.rs
+++ b/facet-python/src/lib.rs
@@ -186,10 +186,16 @@ impl PythonGenerator {
         // Collect all generated code in sorted order (BTreeMap iterates in key order)
         let mut output = String::new();
 
-        // Write imports if requested and any were used
-        if write_imports && !self.imports.is_empty() {
-            let imports: Vec<&str> = self.imports.iter().copied().collect();
-            writeln!(output, "from typing import {}", imports.join(", ")).unwrap();
+        // Write imports if requested
+        if write_imports {
+            // Always emit __future__ annotations for postponed evaluation
+            // This allows forward references and | syntax without runtime issues
+            writeln!(output, "from __future__ import annotations").unwrap();
+
+            if !self.imports.is_empty() {
+                let imports: Vec<&str> = self.imports.iter().copied().collect();
+                writeln!(output, "from typing import {}", imports.join(", ")).unwrap();
+            }
             output.push('\n');
         }
 
@@ -398,15 +404,15 @@ impl PythonGenerator {
             }
             StructKind::TupleStruct if variant.data.fields.len() == 1 => {
                 self.generate_newtype_variant(variant_name, &pascal_variant_name, variant);
-                format!("\"{}\"", pascal_variant_name)
+                pascal_variant_name.to_string()
             }
             StructKind::TupleStruct => {
                 self.generate_tuple_variant(variant_name, &pascal_variant_name, variant);
-                format!("\"{}\"", pascal_variant_name)
+                pascal_variant_name.to_string()
             }
             _ => {
                 self.generate_struct_variant(variant_name, &pascal_variant_name, variant);
-                format!("\"{}\"", pascal_variant_name)
+                pascal_variant_name.to_string()
             }
         }
     }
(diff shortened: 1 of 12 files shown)


vron:
Note ready to be merge yet

vron:
Should be ready for review / merging (if accepted) now

fasterthanlime:
Have you tested this in actual python? I think that sorting thing by alphabetical order will produce name errors for enums with data variants.

gpt 5.2 says:

[P1] Emit union alias before defining variant classes — facet-python/src/lib.rs:97-109

    Enums with data variants generate a union alias keyed by the enum name, but in finish the output is assembled by iterating the BTreeMap of generated fragments in sorted key order. Because the alias
    key (e.g., Message) sorts between some variant class names (e.g., ImageUpload and TextMessage), the emitted union line can precede the class definitions it references, leading to a NameError as
    soon as the generated Python is executed. Any enum whose name alphabetically falls before at least one of its variant wrapper classes (common for multiple variants) will produce invalid Python. The
    output order should preserve dependency order so referenced classes are defined first

vron:
That’s a fair point.. no i have only tested a simple example in python so far. I will look into that and fix it - but will be afk until Sunday at the earliest - but will device an alternative sorting that is consistent and compiles then

fasterthanlime:
> That’s a fair point.. no i have only tested a simple example in python so far. I will look into that and fix it - but will be afk until Sunday at the earliest - but will device an alternative sorting that is consistent and compiles then

Excited for that! Have a good weekend!

vron:
So this should now be good to go. Turns out that there was another error (missing Union) that I have fixes.

Regarding the order the order is unchanged, turns out that if we refer to types as "TypeA" instead of TypeA python accepts any order. I have tested in python and tested in an editor (zed) that "goto definiton" works even if we refer to types as "TypeA"

fasterthanlime:
Should the README state this is for Python 3.12+ ?

Plus some P2 stuff from codex:

```
  - [P2] Fix README example to pass required write_imports arg to to_python — /home/amos/bearcove/facet/facet-python/README.md:28-31
    facet-python/README.md calls to_python::<User>(), but the actual API is to_python::<T>(write_imports: bool) (facet-python/src/lib.rs:35), so copying the README
    snippet won’t compile; update the README (or provide a no-arg convenience wrapper) to match the published API.
  - [P2] Fix README example to pass required write_imports arg to finish — /home/amos/bearcove/facet/facet-python/README.md:50-56
    facet-python/README.md shows let py = gen.finish();, but PythonGenerator::finish requires finish(write_imports: bool) (facet-python/src/lib.rs:85), so the multi-
    type example won’t compile as written; update the snippet to call finish(false) (or similar) or adjust the API to provide a default.
  - [P2] Align README “Output” example with actual generated Python — /home/amos/bearcove/facet/facet-python/README.md:34-41
    The README’s “Output” snippet shows email: Optional[str] = None, but the generator currently emits TypedDict fields using Required[...] for non-Option keys and
    unwraps Option<T> fields to just T (see facet-python/src/snapshots/facet_python__tests__optional_field.snap), so the README example is misleading for users trying
    to understand the produced typing surface.
  - [P3] Remove duplicate docs.rs metadata table in facet-python/Cargo.toml — /home/amos/bearcove/facet/facet-python/Cargo.toml:13-17
    facet-python/Cargo.toml defines both [package.metadata.docs.rs] and [package.metadata."docs.rs"], which are distinct TOML keys (nested vs quoted dotted key) and
    can cause confusion/drift; only the standard [package.metadata.docs.rs] is expected to be honored by Cargo/docs.rs tooling.
```
    
Let me know if you're annoyed by the codex review copypaste but in python land I.. need a little helper.

vron:
Thats fine, Ill get on it, currently using it from my fork to pythonify my entire codebase to see if I find any additional errors

vron:
As a heads up, still quite a bit to be done here, running into all sort of cases that needs to be fixed when I use it in anger..

vron:
So now this works for my entire codebase to generate several 1000 lines of python types that all "compiles" with python.

The one remaining issue is that e.g. a chrono::NaiveDate will generate Any as the type. However, I think that is fine for a first version and I will consider maybe adding a "config" later that allows the user to specify "I want to map NaiveDate to str" and similar.


So, this should be ready for review, have however become quire a number of commits as I work, do you want me to squash it all into a single commit "Add facet-python" to keep your history shorter? (@fasterthanlime )

vron:
Do you want me to merge this into one commit or are you happy as is?

fasterthanlime:
> Do you want me to merge this into one commit or are you happy as is?

No worries, I squashed it at merge time

PR #1832, opened 2026-01-20
XML documents with a DOCTYPE now work

softdevca:
Previously any XML document with a `DOCTYPE` would fail.  There is no good place to put the DOCTYPE data in the deserialized object so it's skipped, which should be okay for the vast majority of usages.


fasterthanlime:
Well I'm not thrilled about this, but we're not actually losing any functionality by merging this so... until next PR, this'll do!

softdevca:
I feel the same.  Perhaps  

```rust
#[derive(Facet)]
struct Thing {
    #[facet(xml::doctype)]
    doctype: String
}
```

Issue #1724, opened 2026-01-09
Flatten and enum expects struct


vron:
Hi, I suspect this may be related to #1721 but the error message is very different so opening as a sepperate issue / test case.

I would expect the following to run without panic, but it complains that it expects a "struct start" instead of a string in "ty".

```rust
use facet::Facet;

#[derive(Facet, Clone)]
pub struct O {
  #[facet(flatten)]
  pub p: Pd,
}

#[derive(Facet, Clone)]
#[facet(tag = "ty")]
#[repr(C)]
pub enum Pd {
  A(Ai),
}

#[derive(Facet, Clone)]
pub struct Ai {
  pub pi: String,
}

fn main() {
  let json = r#"
      {
          "ty": "A",
          "pi": "1000"
      }"#;
  let _: O = facet_json::from_str(json).unwrap();
}

```

vron:
The error message is ```Err` value: TypeMismatch { expected: "struct start", got: "Scalar(Str(\"A\"))", span: None, path: None }```

Issue #1721, opened 2026-01-08
Combination of flatten and tag with the same name fails


vron:
The below example fails with "ambigious match despite it actually is well-defined (the same structure works with Serde)

```rust

use facet::Facet;
use facet_pretty::*;

#[derive(Clone, Facet)]
#[facet(tag = "model")]
#[repr(C)]
pub enum Mod {
  A { s: f64 },
  B { s: f64 },
}

#[derive(Clone, Facet)]
pub struct Outer {
  pub c: String,
  #[facet(flatten)]
  pub model: Mod,
}

fn main() {
  let json = r#"
    {
        "c": "example",
        "s": 0,
        "model": "B",
      }
  "#;
  let d: Outer = facet_json::from_str(json).unwrap();
  eprintln!("{}", d.pretty());
  eprintln!("{}", facet_json::to_string(&d).unwrap());
}


```

vron:
So I no longer think this is related to #1724, i *think* the problem might be here that the solver is not able do distinguish between them, i.e. that the solver does not look at the tag but only at the fields of the struct alternatives (and they are here the same)?

But as to what to do about it, I have a hacky change that seems to pass my test case - buts its a combination of me and llm generation - not very large but im not 100% if its the correct / best way to fix it.

Should I send a pr for you to review @fasterthanlime or will that only be noice?

fasterthanlime:
> Should I send a pr for you to review [@fasterthanlime](https://github.com/fasterthanlime) or will that only be noice?

Feel free to open it!

vron:
Ok, will do!

PR #1838, opened 2026-01-21
Support flattening of internally-tagged enums in json

vron:
This is a combination of human and llm work. Added tests and changes to support the two reported errors in issues #1721 and #1724.

The test cases intentionally start from strings and not from types to ensure that we parse the strings (a round trip test from a struct would ahve worked before too)

### DIFF
--- a/facet-format/src/deserializer/struct_with_flatten.rs
+++ b/facet-format/src/deserializer/struct_with_flatten.rs
@@ -83,10 +83,11 @@ where
         // so we can look for tag fields and use their values to select variants.
         if solver.candidates().len() > 1 {
             for ev in &evidence {
-                if let Some(ScalarValue::Str(variant_name)) = &ev.scalar_value {
-                    if solver.hint_variant(variant_name) && solver.candidates().len() == 1 {
-                        break;
-                    }
+                if let Some(ScalarValue::Str(variant_name)) = &ev.scalar_value
+                    && solver.hint_variant(variant_name)
+                    && solver.candidates().len() == 1
+                {
+                    break;
                 }
             }
         }
--- a/facet-json/tests/issue_1721_1724.rs
+++ b/facet-json/tests/issue_1721_1724.rs
@@ -2,23 +2,24 @@
 
 use facet::Facet;
 use facet_json::{from_str as from_json, to_string};
+use std::collections::HashMap;
 
 #[test]
 fn test_deserialize_flattened_enum() {
-    #[derive(Facet, Clone, Debug, PartialEq)]
+    #[derive(Facet, Debug, PartialEq)]
     pub struct O {
         #[facet(flatten)]
         pub p: Pd,
     }
 
-    #[derive(Facet, Clone, Debug, PartialEq)]
+    #[derive(Facet, Debug, PartialEq)]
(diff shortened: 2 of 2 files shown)


fasterthanlime:
clippy's failing, please address :) (AI Labeler / perf-publish can fail that's ok)

fasterthanlime:
codex is unhappy with this one:

```
• The changes introduce scenarios where deserialization can select the wrong enum variant
  based on unrelated string values and can silently accept/ignore inconsistent
  internally-tagged enum tags. Serialization can also silently drop payloads for certain
  internally-tagged flattened enum variants.

  Full review comments:

  - [P1] Avoid hinting enum variants from arbitrary string scalar values — /home/amos/
    bearcove/facet/facet-format/src/deserializer/struct_with_flatten.rs:84-92
    In facet-format/src/deserializer/struct_with_flatten.rs
    (deserialize_struct_with_flatten, lines 84-92), the new disambiguation loop calls
    solver.hint_variant(variant_name) for any ScalarValue::Str found in the probe
    evidence, regardless of which key it came from; since hint_variant only checks
    whether any candidate resolution contains that variant name (not whether the string
    came from the enum tag field), this can incorrectly select the wrong resolution when
    an unrelated string field equals a variant name (e.g., {"c":"B","s":0.0,"model":"A"}
    with variants A/B), producing wrong deserialization results.
  - [P1] Validate internally-tagged enum tag value instead of skipping it — /home/amos/
    bearcove/facet/facet-format/src/deserializer/struct_with_flatten.rs:286-294
    In facet-format/src/deserializer/struct_with_flatten.rs (lines 286-294), the
    internally-tagged enum branch unconditionally skip_value() for the tag field and then
    selects the variant implied by the chosen resolution; this accepts invalid inputs
    silently (wrong type for the tag, or a tag value that conflicts with the rest of the
    object, e.g. fields match variant B but tag says A) and can deserialize to a
    different variant than the tag claims, which is surprising and differs from typical
    internally-tagged semantics.
  - [P2] Prevent silent data loss when flatten-serializing internal-tag enums — /home/
    amos/bearcove/facet/facet-format/src/serializer.rs:830-866
    In facet-format/src/serializer.rs (lines 830-866), the new flattened internally-
    tagged enum handling emits the tag and then only serializes content if
    field_value.into_struct() or field_value.into_enum() succeeds; for internally-tagged
    newtype variants whose payload isn’t a struct/enum (e.g., A(i32)), this path will
    emit only the tag and then continue, dropping the payload entirely—this should likely
    either error out or fall back to a representable encoding rather than silently
    omitting data.
    ```

fasterthanlime:
I can try to tackle it separately if you don't feel like fixing up this PR.

vron:
I f you have time to fix it I think that would be better yes - I have struggled to reach this points where it passes the tests I have defined :-p

I'll send a updated PR with some extra test cases and clippy warning fixed momentarily

vron:
superseeded by the new pull request that should fix clippy + more tests

vron:
@fasterthanlime repoened and pushed, I will close the "new one" then

fasterthanlime:
Closing in favor of #1845 which includes these changes plus bug fixes found during code review.

PR #1839, opened 2026-01-21
Supported flattened enums

vron:
This is a combination of human and llm work. Added tests and changes to support the two reported errors in issues https://github.com/facet-rs/facet/issues/1721 and https://github.com/facet-rs/facet/issues/1724.

The test cases intentionally start from strings and not from types to ensure that we parse the strings (a round trip test from a struct would ahve worked before too)

### DIFF
--- a/facet-format/src/deserializer/struct_with_flatten.rs
+++ b/facet-format/src/deserializer/struct_with_flatten.rs
@@ -83,10 +83,11 @@ where
         // so we can look for tag fields and use their values to select variants.
         if solver.candidates().len() > 1 {
             for ev in &evidence {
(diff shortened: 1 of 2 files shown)


vron:
@fasterthanlime this is then the same change as in the earlier pull reqeust but now with clippy fixed.

fasterthanlime:
> @fasterthanlime this is then the same change as in the earlier pull reqeust but now with clippy fixed.

So ideally you would've just pushed to the existing PR, now the review lives in the previous PR which is.. not ideal.

vron:
aaah, should i try to open the old one and force push?

fasterthanlime:
> aaah, should i try to open the old one and force push?

That would be convenient, thank you.

vron:
On it

PR #1841, opened 2026-01-21
Make proxy structs work in self-describing formats

vron:
This is to fix issue 1236 where previously facet-json failed to parse a json using a proxy defined as a tuple struct


vron:
So this is to fix 1236

The one thing I'm not sure about if there are any additional formats apart from asn1 that needs the non self-describing flag set - but that was the only format where existing tests failed as a result of this change at least.

fasterthanlime:
Closing in favor of #1843

PR #1856, opened 2026-01-22
facet-typescript: Fix untagged enums, tuple structs and proxies

vron:
Several fixes to typescript generation including tests.

Structured as three sepperate, self contained commits.

These are all (to my mind) bugs that I found while genereating typescript types for my codebase

### DIFF
--- a/facet-typescript/src/lib.rs
+++ b/facet-typescript/src/lib.rs
@@ -99,6 +99,129 @@ impl TypeScriptGenerator {
         }
     }
 
+    /// Unwrap through options, pointers, transparent wrappers, and proxies to get the effective shape.
+    ///
+    /// Returns the unwrapped shape along with a flag indicating whether an `Option` was encountered.
+    fn unwrap_to_inner_shape(shape: &'static Shape) -> (&'static Shape, bool) {
+        // Unwrap Option<T> first so we can mark fields as optional.
+        if let Def::Option(opt) = &shape.def {
+            let (inner, _) = Self::unwrap_to_inner_shape(opt.t);
+            return (inner, true);
+        }
+        // Unwrap pointers (Arc, Box, etc.)
+        if let Def::Pointer(ptr) = &shape.def
+            && let Some(pointee) = ptr.pointee
+        {
+            return Self::unwrap_to_inner_shape(pointee);
+        }
+        // Unwrap transparent wrappers
+        if let Some(inner) = shape.inner {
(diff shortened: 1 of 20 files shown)


fasterthanlime:
That one needs work:

  Full review comments:

  - [P1] Apply proxy shape for non-struct/enum proxy targets — /home/amos/bearcove/facet/facet-typescript/src/lib.rs:118-133
    In facet-typescript/src/lib.rs TypeScriptGenerator::generate_shape, proxy handling only takes effect when the proxy shape is a user struct/enum; if #[facet(proxy = ...)] points to a scalar/tuple/collection/etc, the code falls through and
    generates TS from the original shape, which will misrepresent the serialized form (e.g., a newtype proxied to String will still emit the original newtype’s TS rather than string).
  - [P2] Preserve type doc comments for proxied types — /home/amos/bearcove/facet/facet-typescript/src/lib.rs:118-137
    In facet-typescript/src/lib.rs TypeScriptGenerator::generate_shape, the new proxy fast-path returns before the doc-comment emission block, so any type with shape.proxy will silently drop its /// docs from the generated TypeScript output.
  - [P2] Mark Option fields optional in untagged struct-variant objects — /home/amos/bearcove/facet/facet-typescript/src/lib.rs:286-295
    In facet-typescript/src/lib.rs TypeScriptGenerator::generate_enum untagged-enum path, struct variants are rendered as inline object types with field: <type> for every field, so Option<T> becomes field: T | null instead of matching struct
    generation’s field?: T; this is a behavior change specific to untagged enums with struct variants and can over-constrain/incorrectly model “field omitted” vs “field present but null”.
  - [P3] Remove newly introduced #[allow(dead_code)] in tests — /home/amos/bearcove/facet/facet-typescript/src/lib.rs:653-663
    Several newly added tests in facet-typescript/src/lib.rs introduce #[allow(dead_code)], which conflicts with the repo’s AGENTS.md guidance; if these are only present to satisfy the dead_code lint for unconstructed enum variants, it’s typically
    better to construct and drop(...) representative values in the test instead of suppressing the lint.

vron:
Right, hopefully these changes adress your concerns - a lot more test cases included now to.

I know your comment said not to use allow(dead_code) - but i really dont see any other viable alternative when all I do is use the "type" and never create a value?

fasterthanlime:
Superseded by https://github.com/facet-rs/facet/pull/1865 (same fixes rebased & updated). Please review/merge the new PR instead.

PR #1863, opened 2026-01-22
feat(facet-xml): add DOCTYPE declaration support

### LINKED ISSUE #1834: facet-xml: Preserve DOCTYPE
`facet-xml` currently ignores the DOCTYPE in XML documents.   I suggest adding  `xml::doctype` as shown in the attribute below to capture the information.
  
```rust
#[derive(Facet)]
struct Thing {
    #[facet(xml::doctype)]
    doctype: Option<String>
}
```

`xml::doctype` should be ignored on structs that aren't root because the same struct may be reused as a root somewhere else.

fasterthanlime:
## Summary

Add support for XML DOCTYPE declarations through a new `#[facet(xml::doctype)]` attribute. This allows structs to capture and emit DOCTYPE declarations during XML serialization/deserialization, enabling proper handling of DTD references for both SYSTEM and PUBLIC identifiers.

## Changes

- **Core**: Added `is_doctype()` method to `Field` type to identify DOCTYPE fields
- **DOM deserializer**: Handle DOCTYPE events during struct deserialization, storing the declaration content in the designated field
- **DOM serializer**: Emit DOCTYPE declarations before the root element when present
- **XML integration**: Expose DOCTYPE support through the `facet-xml` API
- **Tests**: Comprehensive test coverage for DOCTYPE roundtrip, optional fields, and various identifier formats (SYSTEM, PUBLIC)

## Test Plan

- Added `facet-xml/tests/doctype.rs` with tests covering:
  - Basic DOCTYPE reading from XML
  - Roundtrip serialization/deserialization with DOCTYPE
  - Optional DOCTYPE fields (None case)
  - PUBLIC identifiers (XHTML example)

Closes #1834

### DIFF
--- a/facet-core/src/types/ty/field.rs
+++ b/facet-core/src/types/ty/field.rs
@@ -256,6 +256,15 @@ impl Field {
             || self.has_attr(Some("html"), "tag")
     }
 
+    /// Returns true if this field captures the DOCTYPE declaration (for XML documents).
+    ///
+    /// Checks for `xml::doctype` attribute.
+    /// Used to store the DOCTYPE declaration during deserialization.
+    #[inline]
+    pub fn is_doctype(&self) -> bool {
+        self.has_attr(Some("xml"), "doctype")
+    }
+
     /// Returns true if this field stores metadata.
     ///
     /// Metadata fields are excluded from structural hashing and equality.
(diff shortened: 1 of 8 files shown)


softdevca:
Awesome!  Thanks!

PR #1874, opened 2026-01-23
facet-typescript: Handle default

vron:
Handle default fields similar to optional ones


vron:
This should be ready for review, allows facet-typescript to handled fields with defaults by adding a ? to signal that they are optional

fasterthanlime:
> This should be ready for review

Please ping me when CI is green!

vron:
Fixing clippy now

vron:
@fasterthanlime  ping

Issue #1904, opened 2026-01-23
facet-json / facet-solver: Fails to parse valid json with nested enum


vron:
This is ( I belive ) a case that should parse correctly - at least it used to do with serd and I think looking at it that it should since facet-json by default happily should ignore extra fields.

```rust
use facet::Facet;

#[derive(Facet, Clone, PartialEq, PartialOrd)]
#[facet(tag = "type")]
#[repr(C)]
pub enum NestedEnum {
  NestedVariant { value: f64, label: String },
}

#[derive(Facet, Clone, PartialEq, PartialOrd)]
pub struct VariantA {
  pub foo: String,
  pub bar: String,
  pub count: f64,
  pub nested: Option<NestedEnum>,
}

#[derive(Facet, Clone, PartialEq, PartialOrd)]
pub struct VariantB {
  pub name: String,
  pub extra: String,
}

#[derive(Facet, Clone, PartialEq, PartialOrd)]
#[facet(tag = "type")]
#[repr(C)]
pub enum InnerEnum {
  VariantA(VariantA),
  VariantB(VariantB),
}

#[derive(Facet, Clone, PartialEq, PartialOrd)]
pub struct Outer {
  #[facet(flatten)]
  pub inner: InnerEnum,
  pub id: Option<String>,
}

fn main() {
  let json = r#"{"type":"VariantA","foo":"a","bar":"b","count":1,"nested":null,"id":null,"extra":""}"#;
  let _: Outer = facet_json::from_str(json).unwrap();
}

```

@fasterthanlime I beleive this is an issue with facet-solve, but do not know where to start to solve it

vron:
so i *think* that the reason this fails is the extra "extra" in the string to be parsed, but is should not affect it since the enum variant should be selected from the"type" field

PR #1915, opened 2026-01-25
Change order of tag vs fields selection for enums

vron:
Fixes issue #1904 which under the earlier order could not parse a valid
 json string. Also added test case to check that it still returns an
 error when in strict mode.


vron:
So i *think* this patch has no downside apart from somtimes running the same check twice - but it should be relatively cheap. I see no other obvious way to solve this and avoid that.

PR #2008, opened 2026-02-02
Fix Vec deserialization for flattened internally-tagged enum tuple variants

### LINKED ISSUE #2007: Vec deserialization fails for structs with flattened tagged enum tuple variants
Im pretty sure this used to work but no longer does:

```rust
// Facet bug: deserializing Vec<Outer> fails with "missing field `0`" but single Outer works fine.
// Bug occurs when Outer has a flattened tagged enum with tuple variants.

use facet::Facet;

#[derive(Facet, Clone, PartialEq, Debug)]
pub struct Inner {
  pub value: f64,
}

#[derive(Facet, Clone, PartialEq, Debug)]
#[facet(tag = "type")]
#[repr(C)]
pub enum Tagged {
  TypeA(Inner),
  TypeB(Inner),
}

#[derive(Facet, Clone, PartialEq, Debug)]
pub struct Outer {
  #[facet(flatten)]
  pub tagged: Tagged,
}

#[cfg(test)]
mod tests {
  use super::*;

  #[test]
  fn test_facet_vec_deserialization_bug() {
    let json_single = r#"{"type":"TypeA","value":42.0}"#;
    let json_vec = r#"[{"type":"TypeA","value":42.0},{"type":"TypeB","value":99.0}]"#;

    let single: Outer = facet_json::from_str(json_single).expect("single should work");
    assert!(matches!(single.tagged, Tagged::TypeA(_)));

    let vec_result: Result<Vec<Outer>, _> = facet_json::from_str(json_vec);
    assert!(vec_result.is_ok(), "Vec deserialization should work but fails with: {:?}", vec_result.err());
  }
}


```

fasterthanlime:
## Summary

- Fixes Vec deserialization failing for structs with flattened internally-tagged enum tuple variants
- The bug caused "missing field `0`" errors when deserializing `Vec<Outer>` where `Outer` has a `#[facet(flatten)]` field containing an internally-tagged enum with tuple variants like `TypeA(Inner)`

## The Problem

For internally-tagged enums with flattened fields, the tag field (`type`) is processed first. Previously, after selecting the variant, we closed the enum field, triggering immediate validation. But the variant's fields (like the tuple field in `TypeA(Inner)`) haven't been deserialized yet.

This bug only manifested inside `Vec` because frames inside "movable allocations" (like `ListSlot`) cannot be stored for deferred processing. Without storage, the enum frame gets validated immediately upon closing.

## The Fix

Keep the enum segment open after processing the tag field by calling `keep_final_open()` instead of `close_final()`. This allows subsequent variant fields to be deserialized into the same enum frame.

## Test plan

- [x] Added regression test in `facet-json/tests/integration/issue_2007.rs` covering:
  - Single flattened tagged enum deserialization
  - Vec with multiple elements (different variants)
  - Vec with single element
  - Vec with struct variants (not just tuple variants)
  - Roundtrip serialization/deserialization
- [x] All 339 facet-json tests pass
- [x] All 463 facet-reflect tests pass
- [x] All facet-toml and facet-yaml tests pass

Fixes #2007

### DIFF
--- a/facet-format/src/deserializer/struct_with_flatten.rs
+++ b/facet-format/src/deserializer/struct_with_flatten.rs
@@ -201,10 +201,13 @@ impl<'parser, 'input, const BORROW: bool> FormatDeserializer<'parser, 'input, BO
                                 let wip = nav.take_wip();
                                 nav.return_wip(wip.select_variant_named(variant_name)?);
 
-                                // Only close final if we opened one (may be None if field was already open)
-                                if nav_result.final_segment.is_some() {
-                                    nav.close_final(nav_result.final_is_option)?;
-                                }
+                                // For internally-tagged enums, keep the enum segment open so that
+                                // subsequent fields of the variant can be deserialized into it.
+                                // Without this, the enum frame would be closed and validated here,
+                                // but variant fields haven't been deserialized yet, causing
+                                // "missing field" errors for tuple variants like TypeA(Inner).
+                                // See https://github.com/facet-rs/facet/issues/2007
+                                nav.keep_final_open(&nav_result);
                                 fields_set.insert(field_info.serialized_name);
                                 continue;
                             }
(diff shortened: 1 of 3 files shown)


vron:
Lovely

PR #2033, opened 2026-02-07
feat: JSON Schema Output Omit Null Fields

dougEfresh:
# JSON Schema Output: Omit Null Fields

Hey, the JSON schema generation works perfectly but the output is pretty verbose—every field includes all the null properties even when they're not set. This makes the schemas harder to read and unnecessarily large.

For example, a simple struct like:

```rust
#[derive(Facet)]
struct Simple {
    name: String,
}
```

Currently generates:

```json
{
  "$schema": null,
  "$ref": null,
  "$defs": null,
  "type": "object",
  "properties": {
    "name": {
      "$schema": null,
      "$ref": null,
      "$defs": null,
      "type": "string",
      "properties": null,
      "required": null,
      "additionalProperties": null,
      "items": null,
      "enum": null,
      "minimum": null,
      "maximum": null,
      "oneOf": null,
      "anyOf": null,
      "allOf": null,
      "description": null,
      "title": null,
      "const": null
    }
  },
  ...
}
```

Would be much cleaner as:

```json
{
  "type": "object",
  "properties": {
    "name": {
      "type": "string"
    }
  },
  "required": ["name"],
  "additionalProperties": false,
  "title": "Simple"
}
```

## JSON Schema Spec Says This Is Fine

According to the [JSON Schema 2020-12 spec](https://json-schema.org/draft/2020-12/json-schema-core), omitting keywords is explicitly allowed and has the same behavior as their default values:

- Section 10.2.2.4: "Omitting this keyword has the same behavior as an empty object."
- Section 10.3.1.1: "Omitting this keyword has the same assertion behavior as an empty array."
- Section 10.3.1.2: "Omitting this keyword has the same assertion behavior as an empty schema."

The spec never requires null fields to be present—they can be omitted entirely.

## Proposed Fix

Add `#[facet(skip_serializing_if = Option::is_none)]` to all `Option<T>` fields in the `JsonSchema` struct. This will skip serializing fields that are `None`, making the output much cleaner while remaining spec-compliant.

This is purely a quality-of-life improvement—the current implementation is correct, just unnecessarily verbose.

### DIFF
--- a/facet-json-schema/src/lib.rs
+++ b/facet-json-schema/src/lib.rs
@@ -34,71 +34,65 @@ use facet_core::{Def, Field, Shape, StructKind, Type, UserType};
 /// This is a simplified representation that covers the most common cases.
 /// It can be serialized to JSON using facet-json.
 #[derive(Debug, Clone, Facet)]
+#[facet(skip_all_unless_truthy)]
 pub struct JsonSchema {
     /// The JSON Schema dialect
-    #[facet(rename = "$schema", skip_serializing_if = Option::is_none)]
+    #[facet(rename = "$schema")]
     pub schema: Option<String>,
 
     /// Reference to another schema definition
-    #[facet(rename = "$ref", skip_serializing_if = Option::is_none)]
+    #[facet(rename = "$ref")]
     pub ref_: Option<String>,
 
     /// Schema definitions for reuse
-    #[facet(rename = "$defs", skip_serializing_if = Option::is_none)]
+    #[facet(rename = "$defs")]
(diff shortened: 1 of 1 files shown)


fasterthanlime:
> `#[facet(skip_serializing_if = Option::is_none)]`

Can you switch to a container-level `#[facet(skip_all_unless_truhyt)]` ?

dougEfresh:
@fasterthanlime now using `skip_all_unless_truthy`

fasterthanlime:
Thanks!

PR #2040, opened 2026-02-10
Fix E0401 for generic extension attr payload dispatch

### LINKED ISSUE #2039: Extension attrs on generic containers fail with E0401 due to nested static payload generation
## Summary

`define_attr_grammar!` currently fails on generic container attrs when payload expressions reference outer generic parameters, due to generated nested `static __ATTR_DATA` items.

This blocks a real integration pattern: attaching shape-local typed metadata to generic wrappers (for example `Predict<S>`) so discovery can read metadata directly from shape attrs.

## Why we need this

In downstream frameworks (including ours), generic wrappers need to carry typed callback/metadata payloads on the shape itself. Without this, we have to use global runtime registries keyed by shape id/type-name string matching, which is brittle (registration order and name coupling) and loses the shape-local contract.

## Repro

```rust
use facet::Facet;
use facet_testattrs as testattrs;

#[derive(Facet)]
#[facet(testattrs::generic_size = core::mem::size_of::<S>())]
struct Predict<S> {
    marker: core::marker::PhantomData<S>,
}
```

Current error:

```
error[E0401]: can't use generic parameters from outer item
  = note: a `static` is a separate item from the item that contains it
```

## Root cause

`facet-macros-impl` emits nested `static __ATTR_DATA` for non-unit payload paths in generated `__attr!` arms.

Key locations:
- `facet-macros-impl/src/attr_grammar/make_parse_attr.rs` non-unit dispatch path around the `VariantKind::Newtype(_) | ... | Struct(_) | FnPtr(_)` arms.
- `facet-macros-impl/src/attr_grammar/make_parse_attr.rs` optional-str path (`Option<&'static str>`) also uses nested static initializers.

Those nested statics cannot capture outer generic params, which triggers `E0401`.

## Proposed direction

Replace nested `static __ATTR_DATA` references with inline const references (`&const { ... }`) in those generated macro arms.

This preserves `'static` data for `Attr::new` while avoiding the nested-item generic capture boundary.

## Expected behavior

Generic payload expressions that are valid in const context should compile in container/field attrs for extension grammars (and corresponding builtin optional-str path), including monomorphized payload values.

## Notes

I can submit a PR with regression coverage. I already validated locally that this approach fixes the repro and passes existing `facet` integration tests.

darinkishore:
## Summary

Fixes generic extension-attribute payloads that currently fail with `E0401` because generated `__attr!` macro arms create nested `static __ATTR_DATA` items.

This change updates the non-unit dispatch path to use inline const references (`&const { ... }`) instead of nested statics, so payload expressions can reference outer generic parameters in generic containers.

Closes #2039.

## What changed

- `facet-macros-impl/src/attr_grammar/make_parse_attr.rs`
- In generated `__attr!` arms for:
  - `VariantKind::Newtype(_)`
  - `VariantKind::NewtypeOptionChar`
  - `VariantKind::ArbitraryType(_)`
  - `VariantKind::Struct(_)`
  - `VariantKind::FnPtr(_)`
- Replaced `static __ATTR_DATA: ... = ...; &__ATTR_DATA` with `&const { ... }`.

## Regression coverage

- Added `testattrs::GenericSize(Option<usize>)` in `facet-testattrs`.
- Added integration test `extension_attr_generic_payload.rs` that uses:

```rust
#[derive(Facet)]
#[facet(testattrs::generic_size = core::mem::size_of::<S>())]
struct Predict<S> { ... }
```

and verifies payload decoding at runtime.

## Validation

- `cargo test -p facet --test main`
- `cargo clippy -p facet-macros-impl -p facet-testattrs -p facet --tests -- -D warnings`

### DIFF
--- a/facet-macros-impl/src/attr_grammar/make_parse_attr.rs
+++ b/facet-macros-impl/src/attr_grammar/make_parse_attr.rs
@@ -1776,111 +1776,68 @@ impl ParsedGrammar {
                         let crate_path = self.crate_path.as_ref().expect(
                             "crate_path is required for non-unit variants; add `crate_path ::your_crate;` to the grammar"
                         );
+                        let dispatch_no_args = quote! {
+                            ::facet::__dispatch_attr!{
+                                @crate_path { #crate_path }
+                                @enum_name { #enum_name }
+                                @variants { #(#variants_meta),* }
+                                @name { #key_ident }
+                                @rest { }
+                            }
+                        };
+                        let dispatch_with_args = quote! {
+                            ::facet::__dispatch_attr!{
+                                @crate_path { #crate_path }
+                                @enum_name { #enum_name }
+                                @variants { #(#variants_meta),* }
+                                @name { #key_ident }
+                                @rest { $($args)* }
+                            }
+                        };
+                        let static_attr_body = |dispatch: TokenStream2| {
+                            quote! {
+                                static __ATTR_DATA: #crate_path::Attr = #dispatch;
+                                ::facet::Attr::new(#ns_expr, #key_str, &__ATTR_DATA)
+                            }
+                        };
+                        let const_attr_body = |dispatch: TokenStream2| {
+                            quote! {
+                                ::facet::Attr::new(#ns_expr, #key_str, &const { #dispatch })
+                            }
+                        };
+                        let static_no_args_body = static_attr_body(dispatch_no_args.clone());
+                        let static_with_args_body = static_attr_body(dispatch_with_args.clone());
+                        let const_no_args_body = const_attr_body(dispatch_no_args);
+                        let const_with_args_body = const_attr_body(dispatch_with_args);
                         quote! {
                             // Field-level: @ns { path } attr { field : Type } or with args
                             // Note: $field is tt not ident because tuple struct fields are literals (0, 1, etc.)
                             (@ns { $ns:path } #key_ident { $field:tt : $ty:ty }) => {{
-                                static __ATTR_DATA: #crate_path::Attr = ::facet::__dispatch_attr!{
-                                    @crate_path { #crate_path }
-                                    @enum_name { #enum_name }
-                                    @variants { #(#variants_meta),* }
-                                    @name { #key_ident }
-                                    @rest { }
-                                };
-                                ::facet::Attr::new(#ns_expr, #key_str, &__ATTR_DATA)
+                                #static_no_args_body
                             }};
                             (@ns { $ns:path } #key_ident { $field:tt : $ty:ty | $($args:tt)* }) => {{
-                                static __ATTR_DATA: #crate_path::Attr = ::facet::__dispatch_attr!{
-                                    @crate_path { #crate_path }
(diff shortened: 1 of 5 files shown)


darinkishore:
Update after maintainer feedback re compile-time tradeoff (`static` vs `const`):

I changed this PR to **hybrid codegen** instead of always-`const`.

### What changed
- Threaded a generic-context marker from derive emission:
  - `process_struct.rs` / `process_enum.rs` now pass `has_type_or_const_generics` into extension attr emission.
  - `extension.rs` now emits `@const` for generic type/const-param contexts.
  - `__ext!` forwards optional `@const` into `ns::__attr!`.
- In `make_parse_attr.rs` (non-unit dispatch path in this PR):
  - Default arms use `static __ATTR_DATA` (preserves previous compile-time behavior for non-generic contexts).
  - `@const` arms use `&const { ... }` (generic-safe, fixes E0401).
  - Added a fallback `@const` forwarding arm so non-special variants continue normal behavior.

### Why this addresses the concern
- Non-generic derives keep static storage (no extra const-eval churn there).
- Generic derives switch to const only where needed to cross the nested-static generic boundary.

### Validation
- `cargo test -p facet --test main`
- `cargo clippy -p facet-macros-impl -p facet-testattrs -p facet --tests -- -D warnings`

### Compile-time benchmark (Apple Silicon, local)
Compared against `main` in a separate worktree with clean builds:

- Command: `cargo check -p facet-macros-impl -p facet-testattrs -p facet --tests`
- `main`: `7.119s ± 0.029`
- this branch: `7.075s ± 0.036`

Incremental (no clean, same command):
- `main`: `97.9ms ± 2.5`
- this branch: `102.2ms ± 6.4`

Net: clean build is effectively neutral/slightly better; incremental shows a small delta (single-digit ms).

darinkishore:
Addressing the compile-time concern from @fasterthanlime directly:

We implemented a hybrid codegen strategy so we only use const dispatch where static dispatch is invalid.

- Derive computes `needs_const_dispatch = has_type_or_const_generics`.
- `needs_const_dispatch = false` (non-generic context): we keep the original static `__ATTR_DATA` path.
- `needs_const_dispatch = true` (type/const generics present): we emit `@const`, and `__attr!` builds payloads via `&const { ... }`.

So this is exactly “codegen one or the other depending on whether generics are used”:
- static by default for compile-time friendliness,
- const only for generic cases to avoid E0401.

Why this is reliable:
- The switch is deterministic from parsed generics (not heuristic).
- It is threaded through builtin + extension attrs, and container/field/variant emission paths.
- Regression coverage now includes type-generic + const-generic payloads on containers/fields/variants, including tuple-field paths.

Validation:
- `cargo test -p facet --test main`
- `cargo clippy -p facet-macros-impl -p facet-testattrs -p facet --tests -- -D warnings`

Downstream context (DSRs): generic wrappers (e.g. `Predict<S>`) carry typed shape-local metadata/callback payloads directly on attrs, avoiding fragile global registries keyed by type-name strings.

darinkishore:
@fasterthanlime we're ready boss!

PR #2041, opened 2026-02-10
Fix E0401 for generic optional-string extension attrs

### LINKED ISSUE #2039: Extension attrs on generic containers fail with E0401 due to nested static payload generation

darinkishore:
## Summary

Fixes generic extension-attribute payloads for `Option<&'static str>` variants that fail with `E0401` due to nested `static __ATTR_DATA` generation.

This updates the optional-string branch to use inline const references (`&const { ... }`) so generic payload expressions on container attrs can monomorphize cleanly.

Closes #2039.

## What changed

- `facet-macros-impl/src/attr_grammar/make_parse_attr.rs`
- In `VariantKind::OptionalStr` generated `__attr!` arms (field/container, with/without args):
  - replaced nested `static __ATTR_DATA` usage with `&const { ... }`.

## Regression coverage

- Added `testattrs::GenericName(Option<&'static str>)` in `facet-testattrs`.
- Added integration test `extension_attr_generic_opt_str.rs`:

```rust
#[derive(Facet)]
#[facet(testattrs::generic_name = <S as HasGenericName>::NAME)]
struct GenericNamed<S: HasGenericName> { ... }
```

and verifies the decoded payload equals `"u32"` for `GenericNamed::<u32>`.

## Validation

- `cargo test -p facet --test main`
- `cargo clippy -p facet-macros-impl -p facet-testattrs -p facet --tests -- -D warnings`

## Notes

This is intentionally separate from the non-unit dispatch fix (PR #2040) to keep review scope tight.

### DIFF
--- a/facet-macros-impl/src/attr_grammar/make_parse_attr.rs
+++ b/facet-macros-impl/src/attr_grammar/make_parse_attr.rs
@@ -1565,46 +1565,72 @@ impl ParsedGrammar {
                         quote! {
                             // Field-level: no args → None
                             (@ns { $ns:path } #key_ident { $field:tt : $ty:ty }) => {{
-                                ::facet::Attr::new(
-                                    #ns_expr,
-                                    #key_str,
-                                    &const { #crate_path::Attr::#variant_name(𝟋None) }
-                                )
+                                static __ATTR_DATA: #crate_path::Attr = #crate_path::Attr::#variant_name(𝟋None);
+                                ::facet::Attr::new(#ns_expr, #key_str, &__ATTR_DATA)
                             }};
                             // Field-level with `= "value"` → Some(value)
                             (@ns { $ns:path } #key_ident { $field:tt : $ty:ty | = $val:expr }) => {{
-                                ::facet::Attr::new(
-                                    #ns_expr,
-                                    #key_str,
-                                    &const { #crate_path::Attr::#variant_name(𝟋Some($val)) }
-                                )
+                                static __ATTR_DATA: #crate_path::Attr = #crate_path::Attr::#variant_name(𝟋Some($val));
+                                ::facet::Attr::new(#ns_expr, #key_str, &__ATTR_DATA)
                             }};
                             // Field-level with just expr → Some(value)
                             (@ns { $ns:path } #key_ident { $field:tt : $ty:ty | $val:expr }) => {{
-                                ::facet::Attr::new(
-                                    #ns_expr,
-                                    #key_str,
-                                    &const { #crate_path::Attr::#variant_name(𝟋Some($val)) }
-                                )
+                                static __ATTR_DATA: #crate_path::Attr = #crate_path::Attr::#variant_name(𝟋Some($val));
+                                ::facet::Attr::new(#ns_expr, #key_str, &__ATTR_DATA)
                             }};
                             // Container-level: no args → None
                             (@ns { $ns:path } #key_ident { }) => {{
+                                static __ATTR_DATA: #crate_path::Attr = #crate_path::Attr::#variant_name(𝟋None);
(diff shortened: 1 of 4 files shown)


darinkishore:
Update after maintainer feedback about compile-time cost:

I changed this PR to **hybrid codegen** instead of always-`const` in generic+non-generic cases.

### What changed
- Added generic-context plumbing from derive emission:
  - `process_struct.rs` / `process_enum.rs` pass `has_type_or_const_generics` into extension attr emission.
  - `extension.rs` emits optional `@const` marker.
  - `__ext!` forwards that marker to `ns::__attr!`.
- In `make_parse_attr.rs` for `VariantKind::OptionalStr` (this PR’s scope):
  - Default arms now use `static __ATTR_DATA` (non-generic-friendly path).
  - `@const` arms use `&const { ... }` (generic-safe path that fixes E0401).
  - Generic `@const` fallback arm added at the macro level.

### Why this is better than always-const
- Keeps existing static behavior in non-generic derives.
- Uses const only where generic capture would otherwise fail with nested statics.

### Validation
- `cargo test -p facet --test main`
- `cargo clippy -p facet-macros-impl -p facet-testattrs -p facet --tests -- -D warnings`

### Compile-time benchmark (Apple Silicon, local)
Compared against `main` in a separate worktree.

Clean check command: `cargo check -p facet-macros-impl -p facet-testattrs -p facet --tests`
- `main`: `7.059s ± 0.069`
- this branch: `7.083s ± 0.027`

Incremental (same command, warm cache):
- `main`: `94.8ms ± 2.8`
- this branch: `101.5ms ± 4.7`

Additional clean test build (`cargo test -p facet --test main --no-run`):
- `main`: `8.825s ± 0.057`
- this branch: `8.891s ± 0.064`

So this is near-neutral on clean builds, with a small incremental delta.

PR #2042, opened 2026-02-10
Improve default-conversion panic messages with Display-aware types

fasterthanlime:
Picks up the approach from #2037 with the Display improvements from #2038 in place.

## What this changes
- Points default-conversion panics at the field span (including tuple fields).
- Improves panic messages to include field name plus expected/found type display.
- Uses the same clear message path for unsupported conversions.

## Why
When  yields a value that cannot be converted to the field type, the previous panic was hard to trace and less actionable.

Related context: https://github.com/facet-rs/facet/pull/2037#issuecomment-3879332122


fasterthanlime:
@anson-vandoren I think this captures the improvements you wanted and thinking about it again, I don't think the quantity of code gen changes significantly. I just got confused by the moves.

PR #2052, opened 2026-02-20
fix: json schema number types produces strings, not JSON numbers

dougEfresh:
The `minimum` and `maximum` fields on `JsonSchema` were `Option<i128>` and `Option<u128>`, which facet-json serializes as strings. JSON Schema spec says these MUST be numbers (sections 6.2.2, 6.2.4).

Changed both to `Option<i64>` so they serialize as proper JSON numbers.

Also stopped auto-populating `maximum` with the type's upper bound (e.g. `u32::MAX`) -- it's just noise and not useful schema information.

### DIFF
--- a/facet-json-schema/src/snapshots/facet_json_schema__tests__simple_struct.snap
+++ b/facet-json-schema/src/snapshots/facet_json_schema__tests__simple_struct.snap
@@ -1,6 +1,5 @@
 ---
 source: facet-json-schema/src/lib.rs
-assertion_line: 567
 expression: schema
 ---
 {
@@ -8,8 +7,7 @@ expression: schema
   "properties": {
     "age": {
       "type": "integer",
-      "minimum": "0",
-      "maximum": "4294967295"
+      "minimum": 0
     },
     "name": {
       "type": "string"
--- a/facet-json-schema/src/lib.rs
+++ b/facet-json-schema/src/lib.rs
@@ -70,10 +70,10 @@ pub struct JsonSchema {
     pub enum_: Option<Vec<String>>,
 
     /// For numbers: minimum value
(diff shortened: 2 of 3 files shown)


fasterthanlime:
> were Option<i128> and Option<u128>, which facet-json serializes as strings
uh oh

facet-json-schema/src/snapshots/facet_json_schema__tests__simple_struct.snap
```diff
@@ -1,15 +1,13 @@
...
-assertion_line: 567
 expression: schema
 ---
 {
   "type": "object",
   "properties": {
     "age": {
       "type": "integer",
-      "minimum": "0",
-      "maximum": "4294967295"
```
fasterthanlime, line 12: wait, why did maximum just disappear

dougEfresh: Yeah , had my doubts about this. 

If you want max, we have to cap it at 64 bits.

Otherwise many parsers will break.

fasterthanlime:
> The `minimum` and `maximum` fields on `JsonSchema` were `Option<i128>` and `Option<u128>`, which facet-json serializes as strings. JSON Schema spec says these MUST be numbers (sections 6.2.2, 6.2.4).
> 
> Changed both to `Option<i64>` so they serialize as proper JSON numbers.
> 
> Also stopped auto-populating `maximum` with the type's upper bound (e.g. `u32::MAX`) -- it's just noise and not useful schema information.

Mhh alright if you say so.

fasterthanlime:
Rebased and merged in #2086

PR #2070, opened 2026-02-23
feat(format): cache TypePlanCore plans in format deserializer

### LINKED ISSUE #2069: facet-format: implicit TypePlan static cache (leaked plans)
## Summary

Add implicit TypePlan building at the `facet-format` layer, backed by a global static cache.

## Motivation

We need plan access in format code (for opaque adapter and related paths) without introducing a `facet-core -> facet-reflect` dependency.

A pragmatic approach is:
- build `TypePlanCore` on demand from a shape,
- store it in a global cache,
- return `'static` plan refs by leaking allocations.

## Proposed design

Implement a process-global cache in `facet-format`:

```rust
static PLAN_CACHE: Mutex<FxHashMap<Shape, &'static TypePlanCore>>;
```

Behavior:
1. On lookup, check cache for plan by shape.
2. On miss, build plan (`TypePlanCore::from_shape(...)`).
3. Leak the plan allocation to obtain `&'static TypePlanCore`.
4. Insert into cache and return cached reference.

## Why here (and not facet-core)

`facet-core` should remain independent of `facet-reflect` internals.
Keeping implicit plan building in `facet-format` avoids layering violations while still enabling plan-aware format behavior.

## Tradeoffs

- **Pros**
  - No crate layering break.
  - Simple runtime API at call sites.
  - Fast repeated access after first build.

- **Cons**
  - Intentional memory growth (leaked plans).
  - Global mutex contention risk.
  - Process-lifetime cache with no eviction.

## Open questions

- Cache key should remain `Shape` as proposed, or use `ConstTypeId`/`&'static Shape` internally?
- Do we need sharded locking / lock-free read path up front, or only if contention appears?
- Should we expose cache metrics (entries, misses, build time) for observability?

## Acceptance criteria

- [ ] Add global plan cache in `facet-format` with implicit build-on-miss.
- [ ] Miss path leaks `TypePlanCore` and stores `&'static TypePlanCore`.
- [ ] All call sites needing plan can retrieve it through a single helper.
- [ ] Add tests for cache hit/miss behavior and basic concurrency safety.
- [ ] Document intentional leak behavior and rationale.

fasterthanlime:
## Summary
- add a process-global TypePlan cache in facet-format keyed by root shape
- reuse cached plans across `FormatDeserializer` entry points instead of building a new plan each call
- add cache hit/miss and concurrency tests in facet-format
- add `Partial::alloc_with_plan` / `Partial::alloc_owned_with_plan` to allocate from cached plans

## Testing
- cargo check --all-features --all-targets --message-format=short
- cargo nextest run -p facet-format --all-features

Fixes #2069

### DIFF
--- a/facet-format/src/deserializer/mod.rs
+++ b/facet-format/src/deserializer/mod.rs
@@ -282,7 +282,7 @@ impl<'parser, 'input> FormatDeserializer<'parser, 'input, true> {
     where
         T: Facet<'input>,
     {
-        let wip = Partial::alloc_with_plan(cached_type_plan_arc(T::SHAPE)?)?;
+        let wip = Partial::alloc_with_plan(cached_type_plan_arc::<T>()?)?;
         let partial = self.deserialize_into(wip, MetaSource::FromEvents)?;
         // SpanGuard must cover build() and materialize() which can fail with ReflectError.
         // Created AFTER deserialize_into so last_span points to the final token.
@@ -308,7 +308,7 @@ impl<'parser, 'input> FormatDeserializer<'parser, 'input, true> {
     where
         T: Facet<'input>,
     {
-        let wip = Partial::alloc_with_plan(cached_type_plan_arc(T::SHAPE)?)?;
+        let wip = Partial::alloc_with_plan(cached_type_plan_arc::<T>()?)?;
         let wip = wip.begin_deferred()?;
         let partial = self.deserialize_into(wip, MetaSource::FromEvents)?;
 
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Benchmark comparison on this branch (`facet-json` / `typeplan_reuse`) using:

- With cache: current PR code
- Without cache: temporarily commented out cached allocation in `facet-format/src/deserializer/mod.rs::FormatDeserializer::deserialize` (owned path) and used `TypePlan::<T>::build()?.partial_owned()?`

Command:
`cargo bench -p facet-json --bench typeplan_reuse`

Mean times:

- `batch_1000_fresh_typeplan`: 4.414 ms (with) vs 5.377 ms (without) => ~17.9% faster
- `company_fresh_typeplan`: 6.5 us (with) vs 7.98 us (without) => ~18.5% faster
- `person_fresh_typeplan`: 4.293 us (with) vs 5.228 us (without) => ~17.9% faster
- `point_fresh_typeplan`: 1.125 us (with) vs 1.58 us (without) => ~28.8% faster

As expected, `*_reused_typeplan` benchmarks were roughly unchanged (already explicitly reusing a plan in the benchmark harness).

PR #2086, opened 2026-02-24
fix: json schema number types produces strings, not JSON numbers

fasterthanlime:
This is a rebased version of #2052 on top of current `main`.

The `minimum` and `maximum` fields on `JsonSchema` were `Option<i128>` and `Option<u128>`, which facet-json serializes as strings. JSON Schema spec says these must be numbers.

Changes:
- Use `Option<i64>` for both `minimum` and `maximum` so they serialize as JSON numbers.
- Stop auto-populating `maximum` with type upper bounds (for example `u32::MAX`) to avoid noisy schema output.

Supersedes #2052.

### DIFF
--- a/facet-json-schema/src/lib.rs
+++ b/facet-json-schema/src/lib.rs
@@ -70,10 +70,10 @@ pub struct JsonSchema {
     pub enum_: Option<Vec<String>>,
 
     /// For numbers: minimum value
-    pub minimum: Option<i128>,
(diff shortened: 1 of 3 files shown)


dougEfresh:
@fasterthanlime 
Not urgent and just curious when cut release? I've been using main for a while now without issues.
I can assist with testing specific features.

PR #2102, opened 2026-02-27
facet-pretty: Handle Def::Result

Mrmaxmeier:
Results previously fell through to the "unsupported" case:

`unsupported peek variant: Ok(42) (Result<i32, String>, User(Opaque))`

They're now printed as `Result<i32, String> Ok(42)`.

I've also included a commit that moves to the insta inline snapshots (they're nice if the snapshots are small), but I'm not particularly attached to that change :smiley:


fasterthanlime:
Thanks!

PR #2106, opened 2026-02-27
Make OptionVTable callbacks extern "C" and FFI-safe

### LINKED ISSUE #2105: Make OptionVTable function pointers extern "C"
## Context

fad (bearcove/fad) is a JIT deserializer that emits machine code via dynasmrt. To deserialize `Option<T>`, we need to call `init_none` and `init_some` from JIT-generated code.

## Problem

The current `OptionVTable` function pointer types use Rust ABI:

```rust
pub type OptionInitSomeFn = unsafe fn(option: PtrUninit, value: PtrMut) -> PtrMut;
pub type OptionInitNoneFn = unsafe fn(option: PtrUninit) -> PtrMut;
pub type OptionIsSomeFn = unsafe fn(option: PtrConst) -> bool;
// etc.
```

JIT-generated code can only call functions with a known calling convention. Rust ABI is unstable and platform-specific — we can't emit correct call sequences for it. We need `extern "C"` so the calling convention is well-defined.

## Proposed change

Change all `OptionVTable` function pointer types to `extern "C"`:

```rust
pub type OptionInitSomeFn = unsafe extern "C" fn(option: PtrUninit, value: PtrMut) -> PtrMut;
pub type OptionInitNoneFn = unsafe extern "C" fn(option: PtrUninit) -> PtrMut;
pub type OptionIsSomeFn = unsafe extern "C" fn(option: PtrConst) -> bool;
pub type OptionGetValueFn = unsafe extern "C" fn(option: PtrConst) -> Option<PtrConst>;
pub type OptionReplaceWithFn = unsafe extern "C" fn(option: PtrMut, value: Option<PtrMut>);
```

And update the monomorphized implementations in `facet-core/src/impls/core/option.rs` to match:

```rust
unsafe extern "C" fn option_init_some<T>(option: PtrUninit, value: PtrMut) -> PtrMut { ... }
unsafe extern "C" fn option_init_none<T>(option: PtrUninit) -> PtrMut { ... }
// etc.
```

## Concern: Option<PtrConst> and Option<PtrMut> in extern "C"

`OptionGetValueFn` returns `Option<PtrConst>` and `OptionReplaceWithFn` takes `Option<PtrMut>`. These are repr(Rust) types containing repr(Rust) inner types, which is technically not FFI-safe. However:

- `PtrConst`/`PtrMut` are pointer-sized (they wrap `*const u8` / `*mut u8`)
- `Option<*const u8>` is guaranteed NPO by Rust
- In practice this works, but we may want to flatten these to raw pointers with null=None for strict correctness

The simplest fix might be to change these two signatures to use raw pointers (null = None) instead of Option.

## Who's affected

All callers of these vtable functions — currently facet-reflect's Partial API and fad's JIT. The Partial API calls these through normal Rust code, so `extern "C"` works fine there too (just slightly less optimizable by the compiler, though these are called through function pointers anyway so the ABI overhead is negligible).

fasterthanlime:
## Summary
Make `OptionVTable` callbacks callable from JIT-generated code by using a stable C ABI.

## Changes
- Changed `OptionVTable` function pointer aliases to `unsafe extern "C" fn`
- Updated option callback implementations in `facet-core` to `extern "C"`
- Kept `TypeOpsIndirect::is_truthy` on Rust ABI via a small wrapper split
- Flattened `OptionGetValueFn` and `OptionReplaceWithFn` to nullable raw pointers for strict FFI-safety:
  - `get_value` now returns `*const u8` (`null` = `None`)
  - `replace_with` now takes `*mut u8` (`null` = `None`)
- Updated all current call sites in `facet-core`/`facet-reflect` to use null-pointer semantics
- Updated JIT helper casts to match `extern "C"` option init callbacks

## Test Plan
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo nextest run`

Closes #2105

### DIFF
--- a/facet-format/src/jit/helpers.rs
+++ b/facet-format/src/jit/helpers.rs
@@ -937,9 +937,8 @@ pub unsafe extern "C" fn jit_result_init_ok_from_value(
     init_ok_fn: *const u8,
 ) {
     // Call init_ok(result, value_ptr)
-    // ResultInitOkFn uses Rust ABI, not extern "C" - matters on Windows x64
     use facet_core::{PtrMut, PtrUninit};
-    type InitOkFn = unsafe fn(PtrUninit, PtrMut) -> facet_core::PtrMut;
+    type InitOkFn = unsafe extern "C" fn(PtrUninit, PtrMut) -> facet_core::PtrMut;
     let init_ok: InitOkFn = unsafe { std::mem::transmute(init_ok_fn) };
     unsafe { init_ok(PtrUninit::new(out), PtrMut::new(value_ptr)) };
 }
@@ -957,9 +956,8 @@ pub unsafe extern "C" fn jit_result_init_err_from_value(
     init_err_fn: *const u8,
 ) {
     // Call init_err(result, value_ptr)
-    // ResultInitErrFn uses Rust ABI, not extern "C" - matters on Windows x64
     use facet_core::{PtrMut, PtrUninit};
-    type InitErrFn = unsafe fn(PtrUninit, PtrMut) -> facet_core::PtrMut;
+    type InitErrFn = unsafe extern "C" fn(PtrUninit, PtrMut) -> facet_core::PtrMut;
     let init_err: InitErrFn = unsafe { std::mem::transmute(init_err_fn) };
     unsafe { init_err(PtrUninit::new(out), PtrMut::new(value_ptr)) };
 }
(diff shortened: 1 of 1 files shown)


fasterthanlime:
Looking at the full picture, **every** vtable fn type in facet-core uses Rust ABI. For fad we'll eventually need extern "C" on:

- **Option**: `init_some`, `init_none`, `is_some`, `get_value`, `replace_with`
- **List** (Vec): `init_in_place_with_capacity`, `push`, `reserve`, `set_len`, `as_mut_ptr_typed`, `len`, `as_ptr`, `as_mut_ptr`, `capacity`
- **Map** (HashMap): `init_in_place_with_capacity`, `insert`, `len`, `contains_key`, `get_value_ptr`, `from_pair_slice`
- **Set** (HashSet): `init_in_place_with_capacity`, `insert`, `len`, `contains`, `from_slice`
- **Result**: `is_ok`, `get_ok`, `get_err`, `init_ok`, `init_err`
- **Pointer** (Box/Arc): `new_into`, `borrow`
- **Array**: `as_ptr`, `as_mut_ptr`
- **Slice**: `len`, `as_ptr`, `as_mut_ptr`
- **Iter**: `init_with_value`, `size_hint`, `dealloc`

Should this PR just do the blanket change across all vtable fn types? The `extern "C"` ABI works fine for callers using normal Rust code too (just called through function pointers, so no inlining difference either way). The only concern is `Option<PtrConst>` / `Option<PtrMut>` in signatures — those aren't FFI-safe strictly speaking, though they're pointer-sized and NPO'd in practice.

Issue #2096, opened 2026-02-26
serialization: no metadata for tuple


wiiznokes:
Consider this code:

```rs
#[derive(Facet)]
struct TupleStruct(bool, bool);
```
The functions called when using facet-format is `begin_seq` and `end_seq`. So there is no way to differentiate a tuple from an array. This would be useful when working with the [ron](https://github.com/ron-rs/ron) format for example, because tuple and array have a different syntax.

PR #2110, opened 2026-02-28
call `struct_metadata` for tuple

wiiznokes:
fix https://github.com/facet-rs/facet/issues/2096


fasterthanlime:
Makes sense, thanks!

PR #2125, opened 2026-03-05
Fix internally-tagged enum newtype

### LINKED ISSUE #2124: facet-json: Nested internally tagged types not working
Hi,

Just encountered a "bug" / unsupported feature that my test cases had not caught before the ser/deser path was used with data stressing it.

In particular the following two test cases do not work - but I think they should?

```rust
use facet::Facet;

#[derive(Facet, Clone, PartialEq, Debug)]
pub struct Filter {
  pub name: String,
}

#[derive(Facet, Clone, PartialEq, Debug)]
#[facet(tag = "add_type")]
#[repr(C)]
pub enum AddOp {
  Full,
  Filtered { filter: Filter },
}

#[derive(Facet, Clone, PartialEq, Debug)]
#[facet(tag = "inner_type")]
#[repr(C)]
pub enum Inner {
  Include(AddOp),
  Exclude(Filter),
}

#[derive(Facet, Clone, PartialEq, Debug)]
#[facet(tag = "outer_type")]
#[repr(C)]
pub enum Outer {
  Nested(Inner),
  Simple { value: f64 },
}

#[cfg(test)]
mod tests {
  use super::*;

  #[test]
  fn inner_exclude_roundtrip() {
    let val = Inner::Exclude(Filter { name: "x".into() });
    let json = facet_json::to_string(&val).unwrap();
    let back: Inner = facet_json::from_str(&json).unwrap();
    assert_eq!(val, back);
  }

  #[test]
  fn outer_nested_roundtrip() {
    let val = Outer::Nested(Inner::Include(AddOp::Filtered { filter: Filter { name: "f".into() } }));
    let json = facet_json::to_string(&val).unwrap();
    let back: Outer = facet_json::from_str(&json).unwrap();
    assert_eq!(val, back);
  }
}


```

vron:
Refactor internally-tagged enum deserialization to properly handle nested newtype chains and fix error handling for ambiguous duplicate tag keys and unsupported scalar newtype flattening.

Added regression test for issue 2124 in facet-json/tests/issue_2124.rs as well as related tests to check the new code.

Fixes #2124

### DIFF
--- a/facet-format/src/deserializer/eenum.rs
+++ b/facet-format/src/deserializer/eenum.rs
@@ -467,154 +467,242 @@ impl<'parser, 'input, const BORROW: bool> FormatDeserializer<'parser, 'input, BO
             return Ok(wip);
         }
 
-        // Check if this is a single-field tuple variant containing an internally-tagged enum
-        // In this case, the inner enum's fields are flattened into the outer object
+        // Check if this is a single-field tuple (newtype) variant.
+        // The inner value's fields are flattened into the enclosing tagged object.
+        // Three cases are handled:
+        //   1. Inner value is a struct — its fields appear alongside the tag.
+        //   2. Inner value is an internally-tagged enum — its tag and fields are
+        //      flattened too (and its variants may themselves be newtypes, so we
+        //      recurse).
+        //   3. Inner value is a scalar / unsupported — error.
         if matches!(
             variant.data.kind,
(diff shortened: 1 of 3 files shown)


vron:
@fasterthanlime is there anything you want me to adress here, As I understand it the comment from copilot was only stylistic in the test file (I can fix that if you want)

fasterthanlime:
> is there anything you want me to adress here,

Nope looks good

PR #2164, opened 2026-04-13
refactor(reflect): single source of truth for Map pending-entry ownership

fasterthanlime:
## Summary

- Replace ad-hoc coordination between `Tracker::Map::pending_entries`, stored MapKey/MapValue frames in `stored_frames`, `sever_parent_pending_for_path`, and `impl Drop for Partial`'s `PathStep::MapValue` arm with a single `owned_by_stored_frame: bool` flag on each pending entry.
- Delete `sever_parent_pending_for_path`'s Map branch and its `stored_map_key_paths: &BTreeSet<Path>` parameter (added in #2163). The flag answers "who drops this key/value buffer?" at the source; Map's `deinit` consults it; no snapshot or parent-pointer disarming is needed for Map.
- `impl Drop for Partial`'s `PathStep::MapValue` arm becomes a no-op: the stored MapValue frame's own `deinit`/`dealloc` drops the value buffer iset-aware, and Map's `deinit` skips the pending entry.

Net diff: **+113 / −151 lines.**

## Motivation

Over the last few months the deferred-mode cleanup paths accumulated four targeted memory-safety fixes (`fix(reflect): drop map key when severing failed MapValue pending entry`, `fix(reflect): prevent double-free in List deinit when Vec is initialized`, `fix(reflect): drain rope before dropping Vec in List frame deinit`, and the recently merged #2163 `fix(reflect): avoid UAF when stored MapKey frame co-owns pending-entry key`). Each was locally reasonable but treated a symptom: the same raw buffer pointer was co-owned by a child `Frame` AND one or more parent-tracker pending slots, and every cleanup path had to independently disarm one side to avoid double-drop. The next nesting combination found a different unarmed pair.

This PR establishes a single invariant for the Map case: the pending entry itself records which side owns drop responsibility. No more cross-data-structure coordination for Map.

## Scope

**In scope:** Map's `pending_entries`.
**Out of scope (follow-up PRs):**
- SmartPointer's / Option's `pending_inner` (same invariant should apply — still handled by the narrowed `sever_parent_pending_for_path` for now)
- `DynamicValue` `pending_entries` / `pending_elements`
- `List` rope (downstream `bearcove/kajit`'s `asm.repr.styx` SIGABRT is via this path and is **not fixed** by this PR — the List-rope ownership coordination is a separate cleanup)

## Test plan

- [x] `cargo nextest run -p facet-reflect` — 513 / 513 pass
- [x] Full ASAN sweep: `RUSTFLAGS=\"-Zsanitizer=address\" cargo +nightly test -p facet-reflect --test main -Zbuild-std --target aarch64-apple-darwin -- --test-threads=1` — 461 / 461 pass
- [x] Downstream `bearcove/kajit` 3 existing map-repro tests pass (`asm_repro_flatten_map_missing_field`, `asm_repro_nested_failure_severs_templates_key`, `asm_repro_option_indexmap_documented_key`)
- [x] Regression tests landed in #2163 (`deferred_map_value_missing_required_field`, `deferred_map_value_deep_box_enum_missing_field`) still pass under ASAN

## Design notes

The `MapInsertState::*_frame_on_stack` / `*_frame_stored` fields are **retained**. They serve a different purpose: they disambiguate the in-progress-entry case (the Map holds a key/value pointer in `insert_state` but no pending entry yet, because the current key or value frame is still being built). `owned_by_stored_frame` addresses the completed-but-deferred-entry case. Both are necessary.

I considered introducing a `FrameOwnership::Transferred` variant, but that would require the parent to wholesale-drop partially-initialized children on error — which is UB. The iset-aware drop has to happen at the child frame; the parent just needs a \"skip me\" signal. The per-entry flag provides that signal.

### DIFF
--- a/facet-reflect/src/partial/mod.rs
+++ b/facet-reflect/src/partial/mod.rs
@@ -388,6 +388,16 @@ pub(crate) struct Frame {
     pub(crate) type_plan: typeplan::NodeId,
 }
 
+/// A key/value pair queued for insertion into a Map's real container at
+/// finalization time. See the doc comment on `Tracker::Map::pending_entries`
+/// for the meaning of `owned_by_stored_frame`.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct PendingMapEntry {
+    pub key_ptr: PtrUninit,
+    pub value_ptr: PtrUninit,
+    pub owned_by_stored_frame: bool,
+}
+
 #[derive(Debug)]
 pub(crate) enum Tracker {
     /// Simple scalar value - no partial initialization tracking needed.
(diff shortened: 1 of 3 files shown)


fasterthanlime:
Closing — wrong approach. Adding a boolean to dodge the partial-init wholesale-drop problem keeps the redundant tracking that motivated the refactor in the first place. Starting over with the real single-source-of-truth fix (parent pending slots only hold pointers to consumed frames; stored frames own their own drop).

PR #2170, opened 2026-04-14
feat(reflect): add Poke API parity with Peek

### LINKED ISSUE #2153: facet_reflect: `Poke` API parity with `Peek`
A `Peek` can be converted to more specific types for further APIs.

Currently available in `Peek`

- [x] PeekDynamicValue
- [x] PeekEnum
- [x] PeekMap
- [x] PeekSet
- [x] PeekStruct
- [x] PeekList
- [x] PeekNdArray
- [x] PeekListLike
- [x] PeekPointer
- [x] PeekOption
- [x] PeekResult
- [x] PeekTuple

Currently available in `Poke`:

- [ ] PokeDynamicValue
- [x] PokeEnum
- [ ] PokeMap
- [ ] PokeSet
- [x] PokeStruct
- [x] PokeList
- [ ] PokeNdArray
- [ ] PokeListLike
- [ ] PokePointer
- [ ] PokeOption
- [ ] PokeResult
- [ ] PokeTuple

fasterthanlime:
## Summary

Closes #2153 — adds the 9 `Poke*` types that were missing relative to their `Peek*` counterparts. Each mirrors the existing Peek API shape and, where the underlying vtable allows, adds mutation.

New types:
- `PokeDynamicValue` — scalar setters, `set_array`/`set_object`, `push_array_element`/`insert_object_entry` (shape-checked), `end_array`/`end_object`, `object_get_mut`
- `PokeMap` — `len`, `contains_key[_peek]`, `get[_peek]`, `insert`
- `PokeSet` — `len`, `contains[_peek]`, `insert`
- `PokeOption` — `is_some`/`is_none`, `value`/`value_mut`, `set_some`/`set_none`
- `PokeResult` — `is_ok`/`is_err`, `ok[_mut]`/`err[_mut]`, `set_ok`/`set_err`
- `PokeTuple` — `field`/`field_mut`, POD-checked `set_field`
- `PokePointer` — `borrow_inner` (read-only; `PointerVTable` has no mutating ops today)
- `PokeListLike` — unified List/Array/Slice mutation with `get_mut`/`iter_mut`
- `PokeNdArray` — `get_mut`, `as_mut_ptr`, `byte_stride`

Each type also provides `def()`/`ty()`, `into_inner()` back to `Poke`, and `as_peek_*()` for a read-only view. Matching `Poke::into_map`/`into_set`/`into_option`/`into_result`/`into_tuple`/`into_pointer`/`into_list_like`/`into_ndarray`/`into_dynamic_value` entry points are added on `Poke`.

Three vtable shapes drove the implementation:
1. **In-place mut** (Map/Set/List/NdArray, Option's `replace_with`) — pass `PtrMut` directly, with `ManuallyDrop` around values the vtable will `ptr::read`.
2. **Init-only** (Result, DynamicValue scalars) — `call_drop_in_place` then wrap as `PtrUninit` and call the `init_*` fn.
3. **Read-only** (Pointer) — wrapper exists for symmetry; mutations are punted until the core vtable grows them.

## Bonus fixes

- **facet-core `Result<T, E>` drop fix**: `result_drop` previously located the inner value via `Result::as_ref().ok()/err()` (retagged as `SharedReadOnly` under Stacked Borrows) and then dropped through it — UB, caught by miri once `PokeResult::set_ok`/`set_err` were exercised. Switched to a typed `result_drop<T, E>` that calls `core::ptr::drop_in_place` on the full `Result<T, E>`, wired through a per-monomorphization `build_type_ops<T, E>()` (same pattern `num_complex` uses).
- **`leaks` nextest profile**: moved the `just leaks` runner from a per-target cargo runner env var to a proper `[profile.leaks]` in `.config/nextest.toml` that wraps each test process with `leaks -quiet --atExit --`.

## Breaking changes

- `PokeList::iter_mut` and `PokeListLike::iter_mut` now return `Result<_, ReflectError>` and fail if the list has no contiguous `as_mut_ptr`. The previous infallible API fell back to synthesising a mut iterator from the list's `iter_vtable`, which yields `PtrConst` items backed by shared references — casting to `PtrMut` and writing through them is UB. A `get_mut`-per-index fallback would also be unsound (each call re-retags `container` as Unique, invalidating previously yielded items), so the only sound mut iteration path requires `as_mut_ptr`. Use `get_mut(i)` manually for non-contiguous lists.
- The `impl IntoIterator for PokeList` was removed alongside the signature change.

## Test plan
- [x] `cargo nextest run -p facet-reflect` (533 tests pass, incl. 13 new inline + 7 new PokeTuple integration tests)
- [x] `cargo miri nextest run -p facet-reflect -E 'test(/poke::/)'` — 84 tests, all clean
- [x] `just leaks` (macOS native `leaks --atExit`) — 0 leaks
- [x] `RUSTFLAGS="-Zsanitizer=address" cargo +nightly nextest run -p facet-reflect` — clean
- [x] `cargo build --workspace --all-features`
- [x] `cargo clippy -p facet-reflect --all-targets` (clean)

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -43,6 +43,11 @@ command = 'valgrind --error-exitcode=1 --track-origins=yes --read-var-info=yes -
 # Note: nextest runs from crate dirs, so we need workspace-relative path
 command = '../scripts/lldb-wrapper.sh'
 
+[scripts.wrapper.leaks]
+# Run tests under macOS `leaks -quiet --atExit`
+# Fails (non-zero exit) if any allocation is unreachable at process exit
+command = 'leaks -quiet --atExit --'
+
 # Valgrind profile - runs tests under valgrind
 [profile.valgrind]
 
@@ -67,6 +72,14 @@ run-wrapper = 'valgrind-lean'
 platform = 'cfg(target_os = "macos")'
 filter = 'all()'
 run-wrapper = 'lldb'
+
+# Leaks profile - runs each test process under macOS `leaks --atExit`
+[profile.leaks]
+
+[[profile.leaks.scripts]]
+platform = 'cfg(target_os = "macos")'
+filter = 'all()'
+run-wrapper = 'leaks'
 
 # Debug profile - for attaching debugger with crash handler
 [profile.debug]
--- a/Justfile
+++ b/Justfile
@@ -74,16 +74,10 @@ asan *args:
 # macOS-only: run facet-reflect's tests under `leaks --atExit` to catch
 # memory leaks at process exit. Native speed, so it's orders of magnitude
 # faster than miri for leak regressions. Exits non-zero if any allocation
-# is unreachable at exit. Uses the host target triple so the runner env
-# var is picked up (cargo only applies --target runners when --target is
-# explicit).
+# is unreachable at exit.
 leaks *args:
-    CARGO_TARGET_DIR=target/leaks \
-    MallocStackLogging=1 \
-    CARGO_TARGET_AARCH64_APPLE_DARWIN_RUNNER="leaks -quiet --atExit --" \
-    CARGO_TARGET_X86_64_APPLE_DARWIN_RUNNER="leaks -quiet --atExit --" \
-        cargo test -p facet-reflect \
-        --target "$(rustc -vV | sed -n 's|host: ||p')" -- {{ args }}
+    MallocStackLogging=1 CARGO_TARGET_DIR=target/leaks \
+        cargo nextest run --profile leaks -p facet-reflect {{ args }}
 
 fuzz-smoke-value:
     cargo fuzz run fuzz_value -- -runs=1000
--- a/facet-core/src/impls/alloc/smallvec.rs
+++ b/facet-core/src/impls/alloc/smallvec.rs
@@ -327,6 +327,7 @@ where
                 get_mut: Some(smallvec_get_mut::<A>),
                 as_ptr: Some(smallvec_as_ptr::<A>),
                 as_mut_ptr: Some(smallvec_as_mut_ptr::<A>),
+                swap: None,
             }
         }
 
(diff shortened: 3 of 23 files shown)


fasterthanlime:
@Erik1000 how does that look to you?

Erik1000:
Looks great!
One thing that is missing are the `is_*` methods on poke that can be used to check if it e.g. is an map. This is espeically needed on Poke since once you try to turn it into something more specific, you cannot safely get a Poke back (and only a ReflectError) is returned.
image

Erik1000:
Additionally, it would be great if one, for example in PokeOption, could have a erased method to `set_some` which only requires a `Poke` value (or HeapValue i guess?) and not a `T`. Same goes for PokeResult, PokeMap etc. Almost all methods require a type that implements Facet, but in the world of reflections, you sometimes only have a shape and data via Poke, Peek, HeapValue or similar and cannot possibily get an `T` back.

Erik1000:
Something like you did for `PokeMap` with `contains_key_peek` is what im talking about 👍  but also for `insert` etc

Erik1000:
I think the best way would be to have methods like `insery_dynamic` that take an `HeapValue`, check the shape and other safety guarantees and therefore expose a safe wrapper around the vtable methods

Erik1000:
`PokeEnum` and `PokeStruct` would be happy about `set_field_from_heap`.
`PokeList(Like)` would like to have a push and pop method (and from_heap variant) (if available e.g. on Vec, should fail on slice/array) 
dont know about `PokeNdArray` but it misses mutation stuff as well

Erik1000:
Ah, the ListVTable does not contain a pop nor push. Thats kinda bad. Leads to nasty workarounds: https://github.com/Erik1000/facet-egui/blob/806b84b0e4d92289d74b2b8070480e8f172b44cb/facet-egui/src/probe.rs#L1157-L1192
I think the vtable should have optional methods for push and pop. also maybe swap

Erik1000:
ah, PokeTuple also needs set_field_from_heap

Erik1000 (approved):
Looks perfect 👍

fasterthanlime:
> Looks perfect 👍

Well this is a breaking change so — expect some version churn! Here we go.

Erik1000:
Thank you! I'll update facet-egui to it in a few days and let you know

PR #2261, opened 2026-06-19
facet-json-schema: check for field-level proxy before resolving shape

### LINKED ISSUE #2260: facet-json-schema not respecting `proxy` for `jiff`
I am trying out Facet after watching the [talk at RustWeek](https://www.youtube.com/watch?v=11m5HRMvPmU), and wanted to generate a JSON schema from my types, and use my types to add business logic.

My issue is with `jiff` for the timestamps, they do not appear in the final JSON schema. I tried to do what is said in the [Proxy documentation](https://facet.rs/reference/field-attributes/#proxy) (see the `TimestampProxy` struct below). Fwiw I also tried by delegating to FromStr and Display as said in the docs, but I get the same result.

Here's a way to reproduce:

```toml
[dependencies]
facet = { version = "0.46.5", features = ["jiff02"] }
facet-json-schema = "0.46.1"
jiff = "0.2"
```

```rust
use facet::Facet;

#[derive(Facet)]
#[facet(transparent)]
struct TimestampProxy(String);

impl TryFrom<TimestampProxy> for jiff::Timestamp {
    type Error = jiff::Error;

    fn try_from(proxy: TimestampProxy) -> Result<Self, Self::Error> {
        proxy.0.parse()
    }
}

impl TryFrom<&jiff::Timestamp> for TimestampProxy {
    type Error = std::convert::Infallible;

    fn try_from(ts: &jiff::Timestamp) -> Result<Self, Self::Error> {
        Ok(TimestampProxy(ts.to_string()))
    }
}

#[derive(Facet)]
struct Record {
    #[facet(proxy = TimestampProxy)]
    created_at: jiff::Timestamp,
}

#[derive(Facet)]
struct RecordExpected {
    created_at: TimestampProxy,
}

fn main() {
    let schema = facet_json_schema::to_schema::<Record>();
    println!("=== Actual (proxy ignored) ===");
    println!("{}", schema);

    println!("\n=== Expected (proxy consulted -> string) ===");
    let expected = facet_json_schema::to_schema::<RecordExpected>();
    println!("{}", expected);
}
```

And here's the result:
```
[nix-shell:/tmp/facet-proxy-schema-repro]$ cargo run
   Compiling facet-proxy-schema-repro v0.1.0 (/private/tmp/facet-proxy-schema-repro)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.29s
     Running `target/debug/facet-proxy-schema-repro`
=== Actual (proxy ignored) ===
{
  "type": "object",
  "properties": {
    "created_at": {}
  },
  "required": [
    "created_at"
  ],
  "additionalProperties": false,
  "title": "Record"
}

=== Expected (proxy consulted -> string) ===
{
  "type": "object",
  "properties": {
    "created_at": {
      "type": "string"
    }
  },
  "required": [
    "created_at"
  ],
  "additionalProperties": false,
  "title": "RecordExpected"
}
```

I may be misusing it, or completely misunderstood something, please let me know if it is obvious! As a workaround, I'll use the `TimestampProxy` struct directly.

deadbaed:
Fixes #2260 

Previously the proxy type was not consulted during JSON Schema generation.

When `#[facet(proxy = P)]` is set, the proxy type's Shape is used instead of the declared field type's Shape, producing correct schemas.

For example, `{"type": "string"}` would be produced instead of `{}`.

Added snapshot test based off the website

## AI

I used the model `big-pickle` from [OpenCode](https://opencode.ai/docs/zen) for research and development. I reviewed the code and am confident enough to send this PR.

The test was added manually.

Let me know if I did something wrong.

### DIFF
--- a/facet-json-schema/src/lib.rs
+++ b/facet-json-schema/src/lib.rs
@@ -742,10 +742,9 @@ mod tests {
             }
         }
 
-        impl TryFrom<&CustomId> for CustomIdProxy {
-            type Error = std::convert::Infallible;
-            fn try_from(id: &CustomId) -> Result<Self, Self::Error> {
-                Ok(CustomIdProxy(format!("ID-{}", id.0)))
+        impl From<&CustomId> for CustomIdProxy {
+            fn from(id: &CustomId) -> Self {
+                CustomIdProxy(format!("ID-{}", id.0))
             }
         }
 
--- a/facet-json-schema/src/lib.rs
+++ b/facet-json-schema/src/lib.rs
@@ -393,13 +393,20 @@ impl SchemaContext {
             }
             StructKind::TupleStruct if fields.len() == 1 => {
                 // Newtype - serialize as the inner type
-                self.schema_for_shape(fields[0].shape.get())
+                // If the field has a proxy, the proxy type's shape takes precedence
+                let field = &fields[0];
+                let field_shape = field.proxy_shape().unwrap_or_else(|| field.shape.get());
+                self.schema_for_shape(field_shape)
             }
             StructKind::TupleStruct | StructKind::Tuple => {
                 // Tuple struct as array - collect items for prefixItems
                 let _items: Vec<JsonSchema> = fields
                     .iter()
-                    .map(|f| self.schema_for_shape(f.shape.get()))
+                    .map(|f| {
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Thanks!

PR #2327, opened 2026-06-23
Fix serde_styx deserialization of enumerations

Treeniks:
Noticed that I couldn't get enums working when using `serde_styx` so I looked into it. Found that it expected tags like `@tag` to be split into `TokenKind::At` and `TokenKind::BareScalar`, when it is actually a single `TokenKind::Tag`.

I added a couple tests to check enumerations in `styx-compat-tests`.


fasterthanlime:
Merged in https://github.com/facet-rs/facet/pull/2380

PR #2349, opened 2026-06-24
Add facet-json Weavy fuzz oracle

fasterthanlime:
## Summary

- add a CI-friendly shape-bank oracle for facet-json Weavy deserialization
- compare default facet-json, Weavy interpreter, and JIT-requested Weavy plans across generated valid values and arbitrary byte inputs
- add a standalone facet-json AFL harness with seed corpus for long-running malformed-input fuzzing
- narrow a stale scanner float helper cfg so non-JIT builds stay warning-free

## Notes

The useful fuzz input here is target shape plus JSON bytes plus backend. The new tests keep the shape bank fixed and vary values/bytes, which lets the harness compare against the default facet-json path instead of treating every random byte failure as equivalent.

The AFL crate lives outside the workspace like the postcard harness. Its generated Cargo.lock is ignored and not included.

## Verification

- cargo nextest list -p facet-json -E 'test(weavy_oracle_fuzz)'
- cargo nextest run -p facet-json -E 'test(weavy_oracle_fuzz)'
- cargo nextest list -p facet-json --features jit -E 'test(weavy_oracle_fuzz)'
- cargo nextest run -p facet-json --features jit -E 'test(weavy_oracle_fuzz)'
- cargo check -p facet-json
- cargo check -p facet-json --features jit
- cargo check -p facet-json --features fast,jit
- cargo check --manifest-path facet-json/fuzz-afl/Cargo.toml
- cargo nextest run -p facet-json


fasterthanlime:
The regression here is due to correctness, so I'm fine with merging it.

Issue #2342, opened 2026-06-23
JSON `#[facet(other)]` giving `No variant found with the given name` for deeper type


TeamDman:
```rust
//! Azure DevOps-shaped reproduction: adjacently tagged enum with a RawJson catch-all.
use facet::Facet;
use facet_json::{RawJson, from_str};

#[derive(Debug, Facet)]
#[facet(tag = "scheme", content = "parameters")]
#[repr(C)]
enum Authorization {
    ServicePrincipal(ServicePrincipalParameters),
    #[facet(other)]
    Other(RawJson<'static>),
}

#[derive(Debug, Facet)]
#[facet(rename_all = "camelCase")]
struct ServicePrincipalParameters {
    service_principal_id: String,
    tenant_id: String,
}

#[test]
fn azure_devops_known_adjacent_tagged_variant_deserializes() {
    let json = r#"{"scheme":"ServicePrincipal","parameters":{"servicePrincipalId":"spn","tenantId":"tenant"}}"#;

    let parsed: Authorization = from_str(json).unwrap();

    let Authorization::ServicePrincipal(parameters) = parsed else {
        panic!("expected Authorization::ServicePrincipal");
    };
    assert_eq!(parameters.service_principal_id, "spn");
    assert_eq!(parameters.tenant_id, "tenant");
}

#[test]
#[ignore = "Facet currently errors before trying the #[facet(other)] RawJson fallback"]
fn azure_devops_unknown_adjacent_tagged_variant_falls_back_to_raw_json() {
    let json = r#"{"scheme":"Token","parameters":{"token":"redacted"}}"#;

    let parsed: Authorization = from_str(json).unwrap();

    let Authorization::Other(raw) = parsed else {
        panic!("expected Authorization::Other");
    };
    assert_eq!(raw.as_str(), json);
}

```

gives
```
thread 'integration::azure_devops_adjacent_raw_json_other_repro::azure_devops_unknown_adjacent_tagged_variant_falls_back_to_raw_json' (25292) panicked at facet-json\tests\integration\azure_devops_adjacent_raw_json_other_repro.rs:43:48:
called `Result::unwrap()` on an `Err` value: DeserializeError { span: [0..0), path: <root>, kind: Operation failed on shape Authorization: No variant found with the given name }
stack backtrace:
   0: std::panicking::panic_handler
             at /rustc/ac68faa20c58cbccd01ee7208bf3b6e93a7d7f96/library\std\src\panicking.rs:689
[...]

PR #2365, opened 2026-06-25
JSON #[facet(other)] for deep types

TeamDman:
for #2342

currently just adds tests


TeamDman:
this one is a bit wackier.

some additional context

> Me: does it make sense for the test to assert that Other(raw) is eq to the full json string instead of just the `{"token":"redacted"}` part?

> 5.5 extra high: Yes, I think the full JSON object is the right assertion for this repro.
> For `Other(RawJson)`, the fallback is representing the unknown enum value at the enum boundary. In this case the enum value is:
> ```json
> {"scheme":"Token","parameters":{"token":"redacted"}}
> ```
> If it only captured:
> ```json
> {"token":"redacted"}
> ```
> then we’d lose the `scheme`, which is the actual unknown discriminator. That makes the fallback non-roundtrippable and much less useful for Azure DevOps-style payloads where `parameters` only makes sense together with `scheme`.
> The content-only behavior would make sense for a different shape, something like an `other` variant with explicit tag/content fields. But for plain `Other(RawJson)`, full object is the valuable contract.


this is the original code of mine

using serde
https://github.com/AAFC-Cloud/Cloud-Terrastodon/blob/bbad69e5fa7eab79d2ba6c3071ff6ba42df10c04/crates/azure_devops_types/src/azure_devops_service_endpoint_authorization.rs

using facet with rawjson proxy to avoid the issue this pr fixes
https://github.com/AAFC-Cloud/Cloud-Terrastodon/blob/a5e1dffc4fdb59abad07868dbd8143a0e389924d/crates/azure_devops_types/src/azure_devops_service_endpoint_authorization.rs


https://github.com/serde-rs/serde/issues/912#issuecomment-1868785603 may also be relevant, though that doesn't mention adjacently tagged enums

fasterthanlime:
Merged in https://github.com/facet-rs/facet/pull/2380

PR #2397, opened 2026-06-27
Extend facet urlencoded with partial updates, flattened fields and simple enum support

antonWetzel:
This MR contains some changes I required for creating a `Configuration {A, B, C}` per CLI and create a `Request {A, B}` by updating the `Configuration` with the query. 

### New Features + Tests
- Make `from_str_value` public for partial updates
- Support flattened fields
- Support simple enums

### Open Questions
- This contains separable features, should I split this?
- Should `from_str_value` be public and/or should the doc comment be expanded?
- I don't know the internal enum logic and only tested with simple enums, is this enough enough for enums at the moment?
- The other code changes are not required, should I try to minimize the diff?

### Other Code Changes
- Use `(flat/nested).iter()` instead of `keys()` and `get(key).unwrap()`
- Combine `deserialize_value` and `deserialize_nested_field`
  - only difference was logging/error message with the key


fasterthanlime:
Merged manually, thanks!

PR #2404, opened 2026-06-30
Added tracing-wide to ecosystem

yawn:
`tracing-wide` is a (new) crate that enables wide, typed events in the `tracing` ecosystem including catalogue generation. Has first-class, opt-in `facet` support (for catalogue generation and event deserialization when using native subscribers).

See the following examples:

- https://github.com/yawn/tracing-wide/blob/main/tracing-wide/examples/catalogue-facet.rs or
- https://github.com/yawn/tracing-wide/blob/main/tracing-wide/examples/subscriber-facet.rs

### DIFF
--- a/docs/content/ecosystem/_index.md
+++ b/docs/content/ecosystem/_index.md
@@ -113,6 +113,14 @@ Reusable runtime pieces for incremental systems and lowered programs.
 | [`fable`](/fable/) | Evaluates a tiny typed language over Facet-reflected Rust values. | [facet-rs/facet](https://github.com/facet-rs/facet/tree/main/fable) |
 | [`weavy`](/weavy/) | Provides a lowered-program substrate for interpreters and copy-and-patch backends. | [facet-rs/facet](https://github.com/facet-rs/facet/tree/main/weavy) |
 
+## Observability
+
+Emit and inspect telemetry from reflected Rust types.
+
+| Crate | What it does | Source |
+|-------|--------------|--------|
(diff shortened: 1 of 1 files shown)


yawn:
Clippy wasn't me!

fasterthanlime:
I'm curious about the motivation there. In terms of extending tracing, which parts of tracing do you actually reuse? If you have your own serialization, and you have your own emission, then how is it actually hooked into tracing?

yawn:
Great question! This is meant as a companion crate for tracing. It enables mostly logging _wide_ messages (containing all the data required to understand what they mean semantically) expressed as ordinary Rust structs (as opposed to ad-hoc message with ad-hoc k/v pairs).

These messages can also be collected from a crate + all linked crates and made available for inspection or serialization e.g. for documentation purposes. This "message catalogue" can be a great tool to configure upstream systems (e.g. extract all messages tagged for analytics and configure how they should be mapped to e.g. Hubspot or Posthog) as well as stakeholder communication (e.g. rendered as a PDF with Typst when serializing to TOML of YAML) in the sense of "this is the sum of all messages our system can emit".

In order to make this upstream use case work (for analytics, database etc. - everything that is not just a log) the crate also has an internal subscription mechanism which does yield the actual type of the message (which would need reconstruction at the tracing level).

Facet is one of the serialization options here (next to serde). It can be used to a) inspect arbitrary messages that derive facet and enable routing on message _content_ (not just on message type tags) and b) as a serialization driver for the message catalogue.

Addendum: right now the crate supports just canonical _messages_ but soon it will support canonical _fields_ as well. This can then be used to get type safe access to fields e.g. in instrument annotations (~ `#[instrument(fields(field(MyMessage::USER_ID) = "hello" ...`) which can be useful for ambient autocapture (Option-al fields in a wide message that can get autopopulated through span values).

PR #2411, opened 2026-06-30
Add aliases for named args and subcommands

### LINKED ISSUE #2408: Support repeatable args::alias on named CLI arguments in figue
## Summary

`figue` already supports repeatable `args::env_alias` attributes for config/env
lookups, but it does not currently support repeatable CLI aliases for named
arguments.

It would be useful to support compatibility aliases on named CLI flags so a
field can accept both its canonical long name and one or more older spellings.

## Example

```rust
#[derive(facet::Facet)]
struct Args {
    #[facet(
        args::named,
        args::alias = "drive-letter-pattern",
        args::alias = "path-pattern",
    )]
    drive: Option<String>,
}
```

Desired behavior:

- `--drive`, `--drive-letter-pattern`, and `--path-pattern` should all parse
  into the same field.
- The canonical flag name should remain `--drive`.
- Aliases should be treated as compatibility spellings, not separate fields.

## Motivation

This helps with CLI migrations where a project wants to rename a flag without
breaking existing scripts immediately.

It is also useful for compatibility spellings that are not full renames, such as:

- accepting plural forms alongside singular forms
- accepting common colloquialisms or team-specific spellings alongside the canonical flag name

The feature seems like a good fit for `figue` because:

- `figue` already models repeatable env aliases via `args::env_alias`
- the schema layer already tracks canonical CLI names separately from other
  metadata
- the behavior is local and composable compared to broader parser changes

## Proposed scope

For a first pass, support this only on named CLI arguments.

That means:

- supported: `#[facet(args::named, args::alias = "old-name")]`
- not part of this issue: subcommand aliases
- not part of this issue: short aliases

### Parsing

- The canonical long flag and all configured aliases should map to the same
  `ArgSchema` / insertion path.
- Boolean negation should work for aliases too, e.g. `--no-colour` when
  `colour` is an alias for `color`.
- Parent/subcommand lookup should honor aliases the same way it honors the
  canonical flag.

### Validation

Schema building should reject:

- duplicate aliases on the same field
- alias collisions with another field's canonical long name
- alias collisions with another field's alias
- use of `args::alias` on non-named arguments

### UX surfaces

- Help output should show the canonical flag first, followed by aliases.
- Shell completions should include aliases.
- Unknown-flag suggestions should consider aliases.

## Notes on implementation

There is already a close precedent in `figue` for repeatable alias extraction:
`args::env_alias` in `figue/src/schema/from_schema.rs` and `env_aliases` in the
schema types.

A similar approach for CLI aliases could likely work by:

- extracting repeated `args::alias` attributes during schema construction
- storing them on `ArgSchema`
- teaching CLI lookup / help / completions / suggestions to consult the alias
  list in addition to the canonical flag name

## Why this first

This is a relatively small and self-contained upstreaming slice compared to:

- optional-value named args
- richer help UX
- schema-driven `to_args`
- broader roundtrip behavior changes

So it seems like a good first feature branch for porting behavior from
`teamy-figue` into monorepo `figue`.

## Draft note

This issue draft was prepared with LLM assistance and then reviewed locally.

### LINKED ISSUE #2409: Support repeatable args::alias on subcommand variants in figue
## Summary

`figue` currently supports one canonical CLI spelling per subcommand variant,
but it does not currently support repeatable long-form aliases for subcommands.

It would be useful to support compatibility aliases on subcommand variants so a
command can accept both its canonical name and one or more older or alternate
spellings.

## Example

```rust
#[derive(facet::Facet, Debug)]
#[repr(u8)]
enum Command {
    #[facet(args::alias = "profiles")]
    Profile,
}
```

Desired behavior:

- `profile` and `profiles` should both select the same enum variant.
- The canonical subcommand name should remain `profile`.
- Aliases should be treated as compatibility spellings, not separate
  subcommands.

## Motivation

This helps with CLI migrations where a project wants to rename a command
without breaking existing scripts immediately.

It is also useful for compatibility spellings that are not full renames, such as:

- accepting plural forms alongside singular forms
- accepting common colloquialisms or team-specific spellings alongside the canonical subcommand name

The feature seems like a good fit for `figue` because:

- subcommand matching already has a canonical CLI-name layer distinct from Rust
  variant names and effective names
- the behavior is local and composable compared to broader parser changes
- it pairs naturally with named-argument alias support without depending on it

## Proposed scope

For a first pass, support this only on long-form subcommand spellings.

That means:

- supported: `#[facet(args::alias = "old-name")]` on subcommand variants
- not part of this issue: short subcommand aliases beyond existing `short`
- not part of this issue: named-arg aliases

### Parsing

- The canonical subcommand name and all configured aliases should resolve to the
  same enum variant.
- Aliases should work at any nesting level where the subcommand appears.
- Parser state after an aliased subcommand should behave exactly like the
  canonical subcommand path.

### Validation

Schema building should reject:

- duplicate aliases on the same variant
- alias collisions with another variant's canonical subcommand name
- alias collisions with another variant's alias
- use of `args::alias` on non-subcommand enum variants in contexts where it is
  not meaningful

### UX surfaces

- Help output should show the canonical subcommand first, followed by aliases.
- Shell completions should include aliases.
- Unknown-subcommand suggestions should consider aliases.

## Notes on implementation

A likely implementation approach would be:

- extract repeated `args::alias` attributes from subcommand variants during
  schema construction
- store them on `Subcommand`
- teach subcommand lookup / help / completions / suggestions to consult the
  alias list in addition to the canonical CLI name

## Why this in parallel

This is the sibling feature to named CLI argument aliases:

- both are compatibility-spelling features
- both are relatively small and self-contained
- both help real CLI migrations without requiring larger parser redesigns

Drafting them in parallel should make it easier to review and upstream the work
as two independent slices.

## Draft note

This issue draft was prepared with LLM assistance and then reviewed locally.

TeamDman:
## Summary
- add repeated `#[facet(args::alias = "...")]` support for named args and subcommand variants
- teach schema building, CLI parsing, suggestions, help, and shell completions about aliases
- add regression coverage for collisions, negated bool aliases, and subcommand alias behavior

## Testing
- cargo test -p figue

Closes #2408
Closes #2409

### DIFF
--- a/figue-attrs/src/lib.rs
+++ b/figue-attrs/src/lib.rs
@@ -43,6 +43,18 @@ facet::define_attr_grammar! {
         ///
         /// Usage: `#[facet(args::short = 'v')]` or just `#[facet(args::short)]`
         Short(Option<char>),
+        /// Adds an additional accepted CLI alias.
+        ///
+        /// For named fields, this adds an extra long-form flag spelling such as
+        /// `--scopes` alongside the canonical `--scope`.
+        ///
+        /// For subcommand variants, this adds an extra subcommand spelling such as
+        /// `profiles` alongside the canonical `profile`.
+        ///
+        /// Multiple aliases can be specified by repeating the attribute.
+        ///
+        /// Usage: `#[facet(args::alias = "scopes")]`
+        Alias(&'static str),
         /// Marks a field as a subcommand.
         ///
         /// The field type must be an enum where each variant represents a subcommand.
--- a/figue/src/completions.rs
+++ b/figue/src/completions.rs
@@ -333,21 +333,41 @@ fn generate_zsh_function(
             ""
         };
 
-        for long_name in std::iter::once(&flag.long).chain(flag.aliases.iter()) {
-            if let Some(short) = flag.short {
-                // Short and long are mutually exclusive in completion
+        let long_names: Vec<&str> = std::iter::once(flag.long.as_str())
+            .chain(flag.aliases.iter().map(|alias| alias.as_str()))
+            .collect();
+        let synonymous = flag
+            .short
+            .into_iter()
+            .map(|short| format!("-{short}"))
+            .chain(long_names.iter().map(|long_name| format!("--{long_name}")))
+            .collect::<Vec<_>>();
+        let exclusion_group = if synonymous.len() > 1 {
+            Some(format!("'({})'", synonymous.join(" ")))
+        } else {
+            None
+        };
+
+        if let Some(short) = flag.short {
+            if let Some(exclusion_group) = exclusion_group.as_deref() {
                 out.push_str(&format!(
-                    "        '(-{short} --{long})'-{short}'[{escaped_desc}]{value_spec}'\n",
-                    long = long_name,
-                ));
-                out.push_str(&format!(
-                    "        '(-{short} --{long})'--{long}'[{escaped_desc}]{value_spec}'\n",
-                    long = long_name,
+                    "        {exclusion_group}-{short}'[{escaped_desc}]{value_spec}'\n",
                 ));
             } else {
                 out.push_str(&format!(
-                    "        '--{long}[{escaped_desc}]{value_spec}'\n",
-                    long = long_name,
+                    "        '-{short}[{escaped_desc}]{value_spec}'\n",
+                ));
+            }
+        }
+
+        for long_name in long_names {
+            if let Some(exclusion_group) = exclusion_group.as_deref() {
+                out.push_str(&format!(
+                    "        {exclusion_group}--{long_name}'[{escaped_desc}]{value_spec}'\n",
+                ));
+            } else {
+                out.push_str(&format!(
+                    "        '--{long_name}[{escaped_desc}]{value_spec}'\n",
                 ));
             }
         }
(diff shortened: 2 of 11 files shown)


TeamDman:
A bit of design context on why this PR adds `figue`-level `args::alias` instead of reusing Facet's builtin `#[facet(alias = ...)]`.

Facet does already have a builtin `alias` attribute, but it is not a full match for the CLI feature we needed here:

- In Facet today, field alias storage is singular, not repeatable. `facet-core` currently stores `pub alias: Option<&'static str>` on fields, with a TODO noting that it should probably become a slice to support multiple aliases.
- Enum variants currently have rename support, but they do not have alias storage in `facet-core`, so builtin Facet aliasing does not cover subcommand aliases.
- The behavior in this PR is specifically CLI-facing: repeated aliases, subcommand aliases, help/completions visibility, unknown-input suggestions, and boolean negation through aliases like `--no-colour`.

So if this PR had tried to build on builtin Facet `alias`, we still would have needed additional `figue` machinery for:

- repeated aliases
- variant/subcommand aliases
- CLI-specific behavior around help, completions, suggestions, and negated bool flags

That is why the implementation here uses repeatable `#[facet(args::alias = ...)]` in `figue` rather than trying to route everything through Facet's current builtin alias support.

I do think there is a reasonable follow-up discussion for Facet itself: making builtin `alias` repeatable and adding variant alias storage. If that happens later, `figue` could potentially decide to honor it too.

PR #2420, opened 2026-07-01
vox: axum websocket transport + first-party ws connect on wasm

fasterthanlime:
## What

Make `ws`/`wss` a first-party transport for `vox` even in a Rust→wasm build, and show (and test in CI) a wasm websocket vox client talking to a Rust HTTP server powered by **axum**.

### `AxumWsLink` — new
`vox-websocket` gains an optional, native-only `axum` feature exposing `AxumWsLink`, which implements `Link`/`LinkTx`/`LinkRx` over `axum::extract::ws::WebSocket`. An existing axum HTTP app can now host a vox endpoint on a route alongside everything else. Framing is identical to the native tungstenite `WsLink` (one vox payload per binary frame); text frames are rejected, close/EOF map to `None`.

### Proven end to end
- New `axum-peer-server` binary: a real axum app (plain `GET /` + vox `GET /ws` upgrade) reusing `subject_rust::TestbedService`.
- New `ws-wasm-axum.spec.ts` Playwright test drives the existing Rust/Wasm client against it — 16/16 testbed methods over `ws://…/ws`.
- The wasm client now uses the **first-party** `vox::connect_lane("ws://…")` instead of the low-level link + initiator, so the test actually exercises the public path.

### Bug fix: `vox::connect("ws://…")` never worked on wasm
The wasm `WsLinkSource::next_link` handed the initiator a socket that was still in `CONNECTING` state (`WsLink::new(WebSocket::new(url))`), so the initiator's first `send` failed with `InvalidStateError: still in CONNECTING state`. The prior wasm test only passed because it used `WsLink::connect` (which awaits open) directly. `next_link` now awaits open via `WsLink::connect`; removed the dead `JsError`/`convert_js_err`/`DomException` path.

### CI
Added a wasm32 check of the `vox` facade with `--no-default-features --features runtime,transport-websocket`. The new Playwright spec runs in the existing `wasm-browser-tests` job.

## Testing
- `ws-wasm-axum.spec.ts`: 16/16 green (wasm `vox::connect_lane` → `AxumWsLink` → axum server).
- `ws-wasm.spec.ts` (bare `ws-peer-server`, shared client): green, no regression.
- `cargo clippy -p vox-websocket --features axum -p peer-server`: clean.
- `cargo check --target wasm32-unknown-unknown` for `vox` (+ws feature), `vox-websocket`, `wasm-browser-tests`: clean.
- `cd typescript && pnpm check`: pass. Native `vox-websocket` tests: 4/4.

### DIFF
--- a/.github/workflows/vox.yml
+++ b/.github/workflows/vox.yml
@@ -340,6 +340,9 @@ jobs:
       - name: Check vox (default features) for wasm32
         run: cargo check --target wasm32-unknown-unknown -p vox
 
+      - name: Check vox facade for wasm32 (websocket transport)
+        run: cargo check --target wasm32-unknown-unknown -p vox --no-default-features --features runtime,transport-websocket
+
       - name: Check vox-websocket for wasm32
         run: cargo check --target wasm32-unknown-unknown -p vox-websocket
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8506,8 +8506,10 @@ dependencies = [
 name = "wasm-browser-tests"
 version = "0.10.0-rc.1"
 dependencies = [
+ "futures-util",
  "spec-proto",
  "vox",
+ "vox-rt",
  "vox-types",
  "wasm-bindgen",
  "wasm-bindgen-futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -410,6 +410,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
 dependencies = [
  "axum-core",
+ "base64 0.22.1",
  "bytes",
  "futures-util",
  "http",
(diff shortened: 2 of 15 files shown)


fasterthanlime:
Added a browser-driven **soak/stress test** (`ws-wasm-axum-stress.spec.ts`): opens N concurrent wasm WebSocket connections (first-party `vox::connect_lane`) to the *same* axum server and hammers each with a rotating mix of request/response calls, error paths, and both stream directions (`sum`/`generate`) for a fixed duration. Each call is wrapped in a per-call timeout, so a wedged connection is reported as **stuck** rather than hanging the run; the summary reports connected / total_requests / errors / stuck / elapsed.

Local run — 8 connections × 30s:
```
connected=8/8, requests=390578, errors=0, stuck=0, elapsed=30119ms
```
(~13k req/s across the mixed workload, streams included, nothing stuck.)

The spec asserts all connections establish, zero errors, zero stuck, and a non-trivial request count. It runs in the existing `wasm-browser-tests` job and adds ~30s.

PR #2421, opened 2026-07-01
snark(weavy): best-first recovery search — kill the branch explosion

fasterthanlime:
## What

The recovering GLR parse drained its branch worklist **FIFO** (`VecDeque`) with only
per-key cost-dominance pruning. On any non-trivial input it speculatively ground
through thousands of cost-bearing dead-end branches even when a cheap (usually cost-0)
parse existed — **super-linear, exploding past `BranchStepLimit` around 16 KB**.

This swaps the worklist to a **min-`error_cost` `BinaryHeap`** (uniform-cost / best-first)
and stops expanding once a complete parse is accepted and the frontier's cheapest branch
costs more. The min-cost repair is found first; the costlier tail is pruned. The
finalization already selects the minimum `error_cost` parse, so **the accepted tree is
unchanged** — best-first only changes how fast we get there.

## Numbers (gingembre, `snark-ts-diff recover`)

| input | before | after |
|---|---|---|
| docs-base ×2 (8 KB) | 333 ms, max_live **1251** | **0.61 ms**, max_live 17 |
| ×4 (16 KB) | `BranchStepLimit` (explode) | **1.11 ms** |
| distinct samples (58 KB) | explode | **24 ms** |
| ×32 (128 KB) | — | **10 ms** |

Now ~linear (~14 KB/ms). Same accepted/failed/errors/missing in every case.

## Tests

`130/130` snark tests pass (recovery/parse/gingembre included).

## Note

Touches `lower/weavy.rs`, which is actively evolving for the stencil/JIT work — the
change is self-contained (heap + a 4-line early-stop) but flagging for PL-core review
so it lands cleanly alongside that.

### DIFF
--- a/snark/src/lower/weavy.rs
+++ b/snark/src/lower/weavy.rs
@@ -8,7 +8,7 @@
 use std::{
     borrow::Cow,
     cell::RefCell,
-    collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
+    collections::{BTreeMap, BTreeSet, BinaryHeap, HashMap},
     error::Error,
     fmt,
     sync::Arc,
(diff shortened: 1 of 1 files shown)


fasterthanlime:
Closing — not meant to go through PRs; integrating into PL core's branch instead.

PR #2422, opened 2026-07-01
snark playground: perf, phase timing, benchmark tooling, wasm-opt

fasterthanlime:
## What

Playground work on top of latest origin/main:

- **Per-phase prepare timing** (parse json / validate / normalize / compile tables / build plan) + live parse ms in the results dock.
- **Lazy embedded-language prepare**: only the selected grammar's tables build up front; embedded languages resolve on first use (select-gingembre 5.9 s -> ~0.3 s of table work).
- **Parse worker guard + emit dedup**: rebuild the session only when the content-hashed key changes; dedup the grammar.js->grammar.json emit. Same language no longer re-prepares (was preparing 2-3x on load).
- **React-gated result panels**: mount the body only when open, so a collapsed panel costs nothing to reconcile; heavy panels (s-expression, captures, layers) default closed. Kills the ~400 ms render on a page switch.
- **Leading-edge throttle** (one frame) replaces the 150 ms debounce; busy pill only for slow work.
- **One-click benchmark mode** + graphql size-ladder samples.
- **wasm-opt -O3** in the wasm build (3.68 MB -> 2.96 MB; better cold-start baseline).
- Adapted to main's snark API (by-reference Weavy, unconditional lowering).

## Tests

tsc clean; wasm builds; playground wasmFlow suite green.


fasterthanlime:
Closing — not meant to go through PRs; integrating into PL core's branch instead.

PR #2426, opened 2026-07-02
figue: add implementation links to help output

### LINKED ISSUE #2425: figue: let help output link command implementations
## Problem

When I run `--help` on a figue-powered CLI, the output explains the command surface but it does not help me jump to the implementation for that command.

This becomes more painful for larger CLIs with nested subcommands, where the fastest way to understand or modify behavior is to open the command handler directly.

## Current behavior

- `--help` shows usage, flags, config, and subcommands.
- `subcommand --help` shows help scoped to that subcommand.
- Neither root help nor subcommand help includes implementation references.

## Desired behavior

When `--help` is run for a command or subcommand, figue should be able to include implementation references for the resolved command target:

- a local source file path that can be clicked in supporting terminals/editors to jump to source
- a GitHub URL that can be opened in the browser to inspect the implementation online

This should work for:

- root command help
- nested subcommand help
- the specific command/subcommand that the help text was generated for, not just the top-level CLI type

## Why this matters

This makes help output more useful for contributors and operators working from source:

- faster debugging
- easier onboarding in large CLIs
- easier navigation from runtime surface to implementation

## Proposed shape

Figue should support opt-in help output that can append an `Implementation:` block to generated help text, with one or both of:

- local source path
- GitHub URL

The local path should be emitted as plain text in a way that terminals/editors can detect as clickable source.

The GitHub link should be subcommand-aware, so `foo bar --help` points at the implementation for `bar`, not only the root CLI type.

## Acceptance criteria

- `--help` for the root command can include a local implementation path
- `--help` for the root command can include a GitHub implementation URL
- `--help` for a subcommand can include the correct local implementation path for that subcommand
- `--help` for a subcommand can include the correct GitHub implementation URL for that subcommand
- the feature is opt-in so existing help output is unchanged unless configured

TeamDman:
## Summary

This PR adds opt-in implementation references to figue help output.

When enabled, generated `--help` text can append an `Implementation:` block with:

- a local source file path
- a GitHub URL for the resolved command/subcommand implementation

## What changed

- added help builder support for including implementation source paths
- added help builder support for including GitHub implementation URLs
- moved implementation footer rendering into help generation
- made subcommand help resolve the correct implementation target from Facet shape metadata
- added focused tests for root help and subcommand help

## Motivation

This addresses the gap described in #2425: help output currently explains the command surface, but does not help contributors jump to the command implementation.

## Validation

- `cargo test implementation_source --lib`

Closes #2425

### DIFF
--- a/figue/src/builder.rs
+++ b/figue/src/builder.rs
@@ -56,6 +56,7 @@
 use std::marker::PhantomData;
 use std::path::{Path, PathBuf};
 use std::string::String;
+use std::sync::Arc;
 
 use camino::Utf8PathBuf;
 use facet::Facet;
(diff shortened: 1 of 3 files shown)


TeamDman:
Here is a concrete way to see the new help output locally.

Command:

```powershell
cargo test implementation_source --lib -- --nocapture
```

Example output:

```text
test-app serve

Serve requests.

USAGE:
    test-app serve [OPTIONS]

OPTIONS:
        --port <U16>
            Port to listen on.


Implementation:
    figue\src\driver.rs
    https://github.com/example/teamy/blob/deadbeef/figue/src/driver.rs

test-app

USAGE:
    test-app [OPTIONS] <COMMAND>

OPTIONS:
    -h, --[no-]help
            Show help message and exit.
        --[no-]html-help
            Open HTML help in the browser and exit.
    -V, --[no-]version
            Show version and exit.
        --completions <bash,zsh,fish>
            Generate shell completions.
        --export-jsonschemas <DIR>
            Export JSON Schema files for all config roots.

COMMANDS:
    serve
            Serve requests.


Implementation:
    figue\src\driver.rs
    https://github.com/example/teamy/blob/deadbeef/figue/src/driver.rs
```

PR #2428, opened 2026-07-02
figue: add help pseudo-subcommands

### LINKED ISSUE #2427: figue: support help pseudo-subcommands and help list
## Problem

Figue currently requires `--help` to trigger help output.

That works, but it misses two workflows that are useful in larger CLIs with nested subcommands:

- typing `help` as a subcommand-like shorthand instead of only `--help`
- enumerating the full subcommand tree from a given command level

## Current behavior

- `myapp --help` works
- `myapp subcommand --help` works
- `myapp help` is treated as an unexpected argument unless the schema defines a real `help` subcommand
- there is no built-in way to print the full set of reachable subcommands from a given command level

## Desired behavior

When a CLI does **not** define a real `help` subcommand, figue should accept:

- `myapp help` as shorthand for `myapp --help`
- `myapp subcommand help` as shorthand for `myapp subcommand --help`

It should also support:

- `myapp help list` to print help for all reachable leaf subcommands
- `myapp help list --short` to print just the full command paths
- `myapp subcommand help list` and `myapp subcommand help list --short` for nested command groups

If a CLI **does** define a real `help` subcommand, that explicit subcommand should continue to win over the pseudo-help behavior.

## Why this matters

This gives users and contributors a fast way to inspect a command tree without manually running `--help` on every subcommand.

The `--short` form is especially useful for:

- scripting
- documentation generation
- quick audits of what a command surface exposes

## Acceptance criteria

- `help` acts like `--help` only when no real `help` subcommand exists at that level
- explicit schema-defined `help` subcommands still take precedence
- `help list` prints full help text for all reachable leaf subcommands from the current command level
- `help list --short` prints only the full command paths for those reachable leaf subcommands
- nested command groups support the same behavior
- invalid command paths still fall back to normal help/error behavior instead of panicking

TeamDman:
## Summary

This PR adds pseudo-help subcommand support to figue when no real `help` subcommand exists at the current command level.

It also adds recursive subcommand listing support via:

- `help list`
- `help list --short`

## What changed

- treat `help` as shorthand for `--help` when no explicit `help` subcommand exists
- preserve precedence for schema-defined real `help` subcommands
- add `help list` to render help for all reachable leaf subcommands from the current command level
- add `help list --short` to print only the full leaf command paths
- plumb the pseudo-help list mode through CLI parsing into the driver help short-circuit
- add focused tests for renderer behavior, shorthand help behavior, and explicit-help precedence

## Motivation

This addresses #2427 by making it much easier to inspect and enumerate a CLI's subcommand tree without manually invoking `--help` on each subcommand.

## Validation

- `cargo test help_list --lib`
- `cargo test help_word --lib`

Closes #2427

### DIFF
--- a/figue/src/driver.rs
+++ b/figue/src/driver.rs
@@ -32,7 +32,8 @@ use crate::dump::dump_config_with_schema;
 use crate::enum_conflicts::detect_enum_conflicts;
 use crate::env_subst::{EnvSubstError, RealEnv, substitute_env_vars};
 use crate::help::{
-    generate_help_for_subcommand_with_config_formats, generate_help_list_for_subcommand,
+    generate_help_for_subcommand_with_config_formats,
+    generate_help_list_for_subcommand_with_config_formats,
     generate_root_html_help_with_config_formats_and_anchor, html_help_anchor_for_subcommand_path,
     open_html_help_file, write_html_help_to_temp_file,
 };
@@ -263,11 +264,12 @@ impl<T: Facet<'static>> Driver<T> {
 
                 let config_file_extensions = self.config_file_extensions();
                 let text = if let Some(mode) = layers.cli.help_list_mode {
-                    generate_help_list_for_subcommand(
+                    generate_help_list_for_subcommand_with_config_formats(
                         &self.config.schema,
                         &subcommand_path,
                         &help_config,
                         mode,
+                        &config_file_extensions,
                     )
                 } else {
                     generate_help_for_subcommand_with_config_formats(
--- a/figue/src/driver.rs
+++ b/figue/src/driver.rs
@@ -32,7 +32,7 @@ use crate::dump::dump_config_with_schema;
 use crate::enum_conflicts::detect_enum_conflicts;
 use crate::env_subst::{EnvSubstError, RealEnv, substitute_env_vars};
 use crate::help::{
-    generate_help_for_subcommand_with_config_formats,
+    generate_help_for_subcommand_with_config_formats, generate_help_list_for_subcommand,
     generate_root_html_help_with_config_formats_and_anchor, html_help_anchor_for_subcommand_path,
     open_html_help_file, write_html_help_to_temp_file,
 };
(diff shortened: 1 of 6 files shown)


TeamDman:
Here are some concrete examples from the current focused tests, along with commands to reproduce them from the `figue/` crate directory.

Reproduction commands:

```powershell
cargo test help_word --lib
cargo test help_list --lib
```

More targeted commands:

```powershell
cargo test test_driver_help_word_alias_triggers_help_without_help_subcommand --lib -- --exact
cargo test test_driver_help_list_short_prints_leaf_commands --lib -- --exact
cargo test test_help_list_short_is_recursive_with_full_command_paths --lib -- --exact
cargo test test_help_list_full_is_recursive_for_leaf_subcommands --lib -- --exact
```

Concrete behavior examples:

1. `help` behaves like `--help` when there is no real `help` subcommand.

Example shape exercised by `test_driver_help_word_alias_triggers_help_without_help_subcommand`:

```text
test-app help
```

That path produces a `DriverError::Help` and the rendered help text still includes the normal builtin help surface, including:

```text
test-app
--[no-]help
```

2. `help list --short` prints full leaf command paths from the current command level.

Example exercised by `test_driver_help_list_short_prints_leaf_commands`:

```text
test-app db help list --short
```

Expected output:

```text
test-app db create
test-app db run
test-app db rollback
```

3. The renderer-level recursive short form works from the root as well.

Example exercised by `test_help_list_short_is_recursive_with_full_command_paths`:

```text
myapp help list --short
```

Expected output:

```text
myapp home open
myapp home show
myapp cache open
myapp cache show
```

4. The full form (`help list`) recursively renders leaf help pages rather than intermediate command groups.

`test_help_list_full_is_recursive_for_leaf_subcommands` asserts that the output contains:

```text
myapp home open
myapp home show
myapp cache open
myapp cache show
```

and intentionally does not include standalone intermediate sections like:

```text
myapp home
myapp cache
```

TeamDman:
the help pseudocommand doesn't show up under the `COMMANDS:` header in default behaviour.
There's also `--html-help` from upstream figue
Opportunity to make it so that `--html-help` is instead `--help html` is instead `help html` to transform things into a more command-oriented structure where the user could then `help help` to see all the help that is available for help as the help surface grows

PR #2469, opened 2026-07-09
Add optional bstr Facet support

### LINKED ISSUE #2468: Add first-party Facet support for bstr::BString and bstr::BStr
# Issue Draft: Add first-party Facet support for bstr::BString and bstr::BStr

## Summary

`bstr::BString` currently cannot be used in `#[derive(Facet)]` structs without a local opaque/proxy workaround. Projects that store command output or other byte-oriented text need `BString` because process output can contain invalid UTF-8, but the data is still naturally represented as a byte list in formats like JSON.

It would be useful for `facet` to provide an optional `bstr` feature with first-party `Facet` implementations for:

- `bstr::BString`
- `bstr::BStr`

## Current behavior

A struct containing `BString` fails to derive/use `Facet` directly unless the field is marked opaque and routed through a project-local proxy type.

Example workaround:

```rust
use bstr::BString;
use facet::Facet;

#[derive(Facet)]
struct CommandOutput {
    #[facet(opaque, proxy = BStringJsonProxy)]
    stdout: BString,
}

#[derive(Facet)]
#[facet(transparent)]
struct BStringJsonProxy(Vec<u8>);
```

That proxy exists only to expose the `BString` bytes as a `Vec<u8>` so JSON can serialize them as a byte array.

## Desired behavior

With a `bstr` feature enabled, this should work without a proxy:

```rust
use bstr::BString;
use facet::Facet;

#[derive(Facet)]
struct CommandOutput {
    stdout: BString,
}
```

Expected JSON shape:

```json
{"stdout":[104,105,255]}
```

This preserves invalid UTF-8 bytes and matches the existing `Vec<u8>` proxy behavior.

## Proposed shape

`BString` can be represented as an opaque user type with list semantics over `u8`, similar to the existing `bytes::Bytes` / `bytes::BytesMut` implementations:

- `Type::User(UserType::Opaque)` for nominal third-party type identity
- `Def::List(ListDef< u8 >)` for reflection, serialization, and deserialization behavior

`BStr` can be represented as the borrowed/unsized byte-slice counterpart, serializing as a byte array as well.

## Why this matters

This avoids every downstream project having to define local `BString` proxy types and makes byte-string fields compose normally inside maps, structs, and other derived `Facet` types.

## Acceptance criteria

- Add optional `bstr` feature to `facet` / `facet-core`.
- Implement `Facet` for `bstr::BString`.
- Implement `Facet` for `bstr::BStr`.
- Serialize `BString` and `&BStr` as JSON byte arrays.
- Deserialize JSON byte arrays into `BString`.
- Preserve invalid UTF-8 bytes across JSON round-trips.

## Disclaimer

This issue draft was written with LLM assistance from OpenAI GPT-5.5 via Codex.

TeamDman:
# PR Draft: Add optional bstr Facet support

## Summary

Adds an optional `bstr` feature with first-party `Facet` implementations for `bstr::BString` and `bstr::BStr`.

`BString` is reflected as an opaque user type with list semantics over `u8`, matching the existing pattern used by byte-oriented third-party types such as `bytes::Bytes` / `bytes::BytesMut`. This lets serializers treat it as byte data while preserving its nominal type identity.

Fixes #2468

## Changes

- Add workspace dependency on `bstr`.
- Add `facet-core/bstr` and top-level `facet/bstr` feature flags.
- Include `bstr` in `facet`'s `all-impls` feature.
- Implement `Facet` for `bstr::BString` as a `Def::List` of `u8`.
- Implement `Facet` for `bstr::BStr` as the borrowed byte-slice counterpart.
- Add facet-json integration tests for byte-array serialization and invalid UTF-8 round-trips.

## Behavior

`BString` now works directly in derived structs without an opaque proxy:

```rust
use bstr::BString;
use facet::Facet;

#[derive(Facet)]
struct CommandOutput {
    stdout: BString,
}
```

JSON output uses byte arrays:

```json
{"stdout":[104,105,255]}
```

This preserves invalid UTF-8 bytes and matches the behavior downstream users previously had to implement with a `Vec<u8>` proxy.

## Tests

Run locally:

```text
cargo check -p facet-core --features bstr
cargo check -p facet --features bstr
cargo test -p facet-json --test main bstr
```

The facet-json test filter covers:

- `BString` serializes as a byte array.
- `BString` round-trips invalid UTF-8 bytes.
- `&BStr` serializes as a byte array.
- `BString` fields in derived structs work without proxies.

## Branch

Prepared from `origin/main` on branch:

```text
feat/bstr-facet
```

Current commit:

```text
a19dd9fa549ecd071f8f63c785644295f0cae9ea Add bstr facet support
```

## Disclaimer

This PR description draft was written with LLM assistance from OpenAI GPT-5.5 via Codex.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -948,6 +948,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5cee35f73844aa3014bb606320a6c1f010249dbdf43342fe54b5a4f6a8ed4b79"
+dependencies = [
+ "memchr",
+ "serde_core",
+]
+
+[[package]]
 name = "btparse"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2302,6 +2312,7 @@ name = "facet-core"
 version = "0.50.0-rc.5"
 dependencies = [
  "autocfg",
+ "bstr",
  "bytes",
  "bytestring",
  "camino",
@@ -2439,6 +2450,7 @@ name = "facet-json"
 version = "0.50.0-rc.5"
 dependencies = [
  "axum-core",
+ "bstr",
  "codspeed-divan-compat",
  "compact_str",
  "copypatch",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -393,6 +393,7 @@ blake3 = "^1"
 bolero = "^0.13.4"
 brotli = "8"
 bytes = { version = "^1.11.0", default-features = false }
+bstr = { version = "^1.12.3", default-features = false, features = ["alloc"] }
 bytestring = { version = "^1.4.0", default-features = false }
 cbindgen = "^0.28"
 cc = "^1.0.87"
@@ -604,3 +605,5 @@ styx-tree = { path = "styx-tree" }
 styx-wasm = { path = "styx-wasm" }
 tree-sitter-styx = { path = "tree-sitter-styx" }
 strid-examples = { path = "strid-examples" }
+
+
--- a/facet-core/Cargo.toml
+++ b/facet-core/Cargo.toml
@@ -45,6 +45,8 @@ url = ["alloc", "dep:url"]
 jiff02 = ["alloc", "dep:jiff"]
 # Provide Facet trait implementations for bytes::Bytes
 bytes = ["alloc", "dep:bytes"]
+# Provide Facet trait implementations for bstr::BString and bstr::BStr
+bstr = ["alloc", "dep:bstr"]
 # Provide Facet trait implementations for bytestring::ByteString
 bytestring = ["alloc", "dep:bytestring"]
 # Provide Facet trait implementations for compact_str::CompactString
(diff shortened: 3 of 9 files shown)


theaddonn:
Thanks for tackling this @TeamDman — the lossless byte-array round-trip is great, and it matches what `bstr`'s own serde impls do (`serialize_bytes`, which serde_json renders as `[…]`).

One design concern from a downstream angle — I'm porting an NBT/SNBT codec ([nbtx](https://github.com/bedrock-crustaceans/bedrockrs-nbt)) onto facet and this issue is my blocker: reflecting `BString`/`BStr` as `Def::List<u8>` makes them **structurally identical to `Vec<u8>`/`&[u8]`**. A format then has no way to tell a byte-*string* from a byte-*list*, so it can't opt to render valid-UTF-8 content as a real string. For text formats that's the whole point of a byte string — a `Value::String(BString)` needs to emit `"text"` in SNBT (and a quoted string in human-readable output generally), not a numeric array or a `[B;…]` literal.

`bstr`'s serde keeps that door open via the distinct `serialize_bytes` hint (which bincode/msgpack special-case, and which a format *could* map to a string when valid UTF-8); `Def::List<u8>` collapses that distinction. Would an opaque/scalar representation that exposes the bytes but preserves the nominal "byte string" identity — closer to how `String`/`str` are scalars rather than `List<char>` — be viable? That keeps your lossless default while letting each format decide bytes-vs-string.

Not blocking; the PR already unblocks the derive case for me. Just flagging before the representation sets. Could this get another review? cc @fasterthanlime 🙏

TeamDman:
@coolbear-bot were you able to address the above comment regarding the underlying representation and the desire to differentiate string-like `Vec<u8>` versus pure `Vec<u8>`?

PR #2471, opened 2026-07-09
Respect proxies when building figue arg schemas

### LINKED ISSUE #2470: figue struct CLI arg schema should respect field and type proxies
## Summary

`figue` rejects struct-shaped fields in CLI args unless they are flattened. That is correct for nested argument structs, but it currently rejects fields even when the field or type has a Facet proxy to a supported single-value CLI type such as `String`.

This comes up for CLI positional fields whose Rust type is a newtype or structured wrapper that should parse from one token. The field can already declare `#[facet(proxy = String)]`, but `figue` schema construction uses the raw field shape and rejects it before the proxy shape is considered.

## Reproduction

```rust
use std::str::FromStr;
use uuid::Uuid;

#[derive(facet::Facet)]
struct Args {
    #[facet(figue::positional, proxy = String)]
    id: WorkItemQueryId,
}

#[derive(facet::Facet)]
struct WorkItemQueryId(Uuid);

impl FromStr for WorkItemQueryId {
    type Err = uuid::Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        Ok(Self(Uuid::parse_str(s)?))
    }
}

impl TryFrom<String> for WorkItemQueryId {
    type Error = uuid::Error;

    fn try_from(value: String) -> Result<Self, Self::Error> {
        value.parse()
    }
}

impl From<&WorkItemQueryId> for String {
    fn from(value: &WorkItemQueryId) -> Self {
        value.0.to_string()
    }
}
```

## Current Output

```text
The application panicked (crashed).
Message:  CLI schema should be valid: Error: struct fields in args must use #[facet(flatten)]
   ╭─[ <unknown>:6:5 ]
   │
 3 │ struct AzureDevOpsWorkItemQueryInvokeArgs {
   │        ─────────────────┬────────────────
   │                         ╰────────────────── defined at crates\entrypoint\src\cli\command\azure_devops\work_item_query\invoke\azure_devops_work_item_query_invoke_cli.rs:9
   │
 6 │     id: AzureDevOpsWorkItemQueryId,
   │     ───────────────┬──────────────
   │                    ╰──────────────── this field is a struct type
 7 │ }
   │ ┬
   │ ╰── end of definition
───╯

Location: crates\entrypoint\src\entrypoint.rs:32
```

The message is directionally correct for nested CLI args, but the schema builder should not fail here when the field has a proxy to a supported single-value CLI type.

## Desired Behavior

A struct-shaped field with `#[facet(proxy = String)]` at the CLI field level should build a CLI schema using the proxy shape.

A struct-shaped type with type-level `#[facet(proxy = String)]` should also build a CLI schema using the proxy shape.

If neither flattening nor a supported proxy applies, the diagnostic should say:

```text
The application panicked (crashed).
Message:  CLI schema should be valid: Error: struct fields in args are not valid unless one of these applies:
- add #[facet(flatten)] to the CLI field to expose nested argument fields
- add #[facet(proxy = T)] to the CLI field or type definition, where T is a supported single-value CLI type: bool, string/char, integer, float, or a unit-variant enum
   ╭─[ <unknown>:6:5 ]
   │
 3 │ struct AzureDevOpsWorkItemQueryInvokeArgs {
   │        ─────────────────┬────────────────
   │                         ╰────────────────── defined at crates\entrypoint\src\cli\command\azure_devops\work_item_query\invoke\azure_devops_work_item_query_invoke_cli.rs:9
   │
 6 │     id: AzureDevOpsWorkItemQueryId,
   │     ───────────────┬──────────────
   │                    ╰──────────────── this field has a struct-shaped Facet schema
 7 │ }
   │ ┬
   │ ╰── end of definition
───╯

Location: crates\entrypoint\src\entrypoint.rs:32
```

## Notes

Supported proxy targets for CLI args are the existing `figue` leaf categories: bool, string/char, integer, float, or a unit-variant enum.

## LLM Assistance Disclaimer

This issue draft was written with LLM assistance from OpenAI GPT-5 Codex.

TeamDman:
## Summary

This updates `figue` arg schema construction to use Facet proxy shapes before deciding whether a CLI argument is a struct or a supported leaf value.

With this change, struct-shaped fields can be accepted as single CLI values when either of these applies:

- the CLI field has `#[facet(proxy = T)]`
- the field type has type-level `#[facet(proxy = T)]`

The proxy target `T` still has to be a supported single-value CLI type: bool, string/char, integer, float, or a unit-variant enum.

## Motivation

The previous schema logic used the raw field shape:

```rust
value_schema_from_shape(field.shape(), &field_ctx)
```

That meant a CLI field like this still failed schema construction even though it explicitly supplied a scalar proxy:

```rust
#[facet(figue::positional, proxy = String)]
pub id: AzureDevOpsWorkItemQueryId,
```

The raw type is struct-shaped, but the CLI representation is meant to be a single `String` token.

## Behavior

The schema builder now resolves `field.proxy` and type-level `shape.proxy` before classifying arg values.

If a struct-shaped arg still has no usable flattening or proxy, the diagnostic now says:

```text
struct fields in args are not valid unless one of these applies:
- add #[facet(flatten)] to the CLI field to expose nested argument fields
- add #[facet(proxy = T)] to the CLI field or type definition, where T is a supported single-value CLI type: bool, string/char, integer, float, or a unit-variant enum
```

## Testing

Added schema tests covering:

- unflattened/unproxied struct CLI field fails with the expected diagnostic
- struct CLI field with a field-level proxy builds successfully
- struct CLI field with a type-level proxy builds successfully

Ran:

```text
cargo test -p figue schema::tests::test_struct_field --lib
cargo check -p figue
```

## Related Issue

Closes #2470

## LLM Assistance Disclaimer

This PR draft was written with LLM assistance from OpenAI GPT-5 Codex.

### DIFF
--- a/figue/src/schema/from_schema.rs
+++ b/figue/src/schema/from_schema.rs
@@ -1116,13 +1116,9 @@ fn arg_level_from_fields_with_prefix(
         if matches!(value, ValueSchema::Struct { .. }) {
             return Err(SchemaError::new(
                 field_ctx.clone(),
-                "struct fields in args must use #[facet(flatten)]",
+                "struct fields in args are not valid unless one of these applies:\n- add #[facet(flatten)] to the CLI field\n- add #[facet(transparent)] to the type definition if it is a newtype over a scalar",
             )
-            .with_primary_label("this field is a struct type")
-            .with_label(
-                field_ctx.clone(),
-                "add #[facet(flatten)] to include its fields at this level",
-            ));
+            .with_primary_label("this field has a struct-shaped Facet schema"));
         }
 
         #[allow(clippy::nonminimal_bool)]
--- a/figue/src/schema/from_schema.rs
+++ b/figue/src/schema/from_schema.rs
@@ -358,10 +358,25 @@ fn leaf_schema_from_shape(
     }
 }
 
+fn value_schema_from_field(
+    field: &Field,
+    ctx: &SchemaErrorContext,
+) -> Result<ValueSchema, SchemaError> {
+    if let Some(proxy) = field.proxy {
+        return value_schema_from_shape(proxy.shape, ctx);
+    }
+
+    value_schema_from_shape(field.shape(), ctx)
+}
+
 fn value_schema_from_shape(
     shape: &'static Shape,
     ctx: &SchemaErrorContext,
 ) -> Result<ValueSchema, SchemaError> {
+    if let Some(proxy) = shape.proxy {
+        return value_schema_from_shape(proxy.shape, ctx);
+    }
+
     match shape.def {
         Def::Option(opt) => Ok(ValueSchema::Option {
             value: Box::new(value_schema_from_shape(opt.t, ctx)?),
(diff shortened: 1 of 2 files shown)


TeamDman:
Working on fixing the commit history on this, oversight led to this not being based off origin/main

PR #2477, opened 2026-07-13
vox: add authenticated Iroh transport

fasterthanlime:
## Summary

Add `vox-iroh`, an authenticated Vox transport over Iroh QUIC bidirectional streams.

- map one Iroh bidirectional stream to Vox `Link`/`LinkSource`
- expose the cryptographically verified remote `EndpointId` as transport-authenticated public-key evidence
- keep Vox connection policy responsible for accepting or rejecting that identity
- frame messages with bounded cancel-safe writes and exact reads
- reserve a versioned ALPN for protocol compatibility
- emit Vox observer connection lifecycle events
- support direct and relay-only connectivity through the same listener/client API

## Production consumer

`checks.vixen.rs` now uses this transport for its outbound fleet protocol. A real Debian 13 agent on Yasuo authenticated a deployed scheduler through Iroh, and the scheduler persisted the enrolled EndpointId/heartbeat in Dibs.

## Proof

- direct path transmits framed Vox traffic and asserts both peer EndpointIds
- forced-relay path transmits the same traffic and identity evidence
- untrusted peer identity is rejected before service use in both directions
- rejected handshakes close gracefully so the peer receives the typed decline before QUIC teardown
- cancellation-safe frame writes and oversize rejection are covered at the stream layer
- ALPN and observer events are certified
- strict Clippy, formatting, and Dodeca validation passed on the published head


fasterthanlime:
Follow-up for the macOS `vix-interpreter` budget failure:

- profiled the 10M-iteration interpreter path with Stax; the hot dispatch cloned and dropped every `Op`
- `a8f53e59d` now executes scalar/control ops through a borrowed fast path and keeps the existing cloned path for structural/hosted ops
- the declared 5s budget is unchanged
- strict interpreter certificate: 5/5 green locally (3.187–4.735s)
- Weavy: 231/231 green
- rung-050 budget suite: 18/18 green
- strict Clippy, fmt, diff-check, and Vox Dodeca validation: green

PR #2483, opened 2026-07-16
vix typed primitives — 02: runtime::primitive core

cramt:
**Phase 2 of 6.** Stacked on #2482 (base `vix-prim-01-spec`).

The `vix::runtime::primitive` module: primitive descriptor / id / `MemoPolicy`; the `Primitive` trait + `EffectCtx` + ticket + completion; `PrimitiveSet` and the `register_function` typed adapter; the taxon→vir validator (lossless subset only); and facet↔`FramedNode` conversion (aggregates are identity trees with **empty resident bytes** — never phon-serialized, per the weavy ABI).

Pure unit tests; no compiler/scheduler wiring yet (the stack is one deliverable — core merges as part of the wired stack).

---
_Stacked PR — review after #2482._

PR #2484, opened 2026-07-16
vix typed primitives — 03: compiler surface

cramt:
**Phase 3 of 6.** Stacked on #2483 (base `vix-prim-02-core`).

The compiler learns registered primitive names from a **manifest** derived from the `PrimitiveSet` (`set.compiler_manifest()` — descriptors projected to vir types + effect ids only; **no handlers cross into the compiler**, preserving the runtime→compiler boundary). Registered names resolve and type-check via the facet-derived schemas through a generalized `where`-call; adds VIR `EffectId`, `EffectKind::Effect`, `Op::EffectRequest`, and typed where-call diagnostics (wrong/missing/extra field, type mismatch).

Effect identity in the IR is `vir::EffectId` (a content hash), never a runtime `PrimitiveId`.

---
_Stacked PR — review after #2483._

### DIFF
--- a/docs/superpowers/plans/2026-07-15-vix-prim-03-compiler.md
+++ b/docs/superpowers/plans/2026-07-15-vix-prim-03-compiler.md
@@ -1,0 +1,665 @@
+# Vix Typed Primitives — Phase 03: Compiler Surface Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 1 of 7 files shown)


fasterthanlime:
Superseded by #2492 and #2495. We retained the generic primitive trait/descriptor, typed request and response values, explicit EffectCtx authority, demand-owned tickets, typed completion, memo policy, and the no-pure-hostcalls boundary. We did not merge the captured-closure registration surface, the alternate demand-layer EffectRequest partition, or the synchronous-v1 scheduler shape: production uses concrete Primitive types, Op::InvokePrimitive lowered to verified Weavy host-call yields, the unified completion inbox, and receipt-backed memoization. Closing this draft stack; its history remains available for reference.

PR #2485, opened 2026-07-16
vix typed primitives — 04: VIR partitioning + lowering

cramt:
**Phase 4 of 6.** Stacked on #2484 (base `vix-prim-03-compiler`).

Partitioning cuts at `Op::EffectRequest`: the request subgraph becomes its own pure value island, and the consuming island records a **generic effect edge** (`EffectEdge { primitive: EffectId, request: ValueIslandId }`) — one edge shape for all primitives, not per-primitive variants. The `LoweringArtifact` carries the effect response as a precomputed **realized value-input** binding (`effect_inputs`), so the scheduler finds the request island warm. Structural only — phase 04 never resolves, memoizes, or dispatches.

---
_Stacked PR — review after #2484._


fasterthanlime:
Superseded by #2492 and #2495. We retained the generic primitive trait/descriptor, typed request and response values, explicit EffectCtx authority, demand-owned tickets, typed completion, memo policy, and the no-pure-hostcalls boundary. We did not merge the captured-closure registration surface, the alternate demand-layer EffectRequest partition, or the synchronous-v1 scheduler shape: production uses concrete Primitive types, Op::InvokePrimitive lowered to verified Weavy host-call yields, the unified completion inbox, and receipt-backed memoization. Closing this draft stack; its history remains available for reference.

PR #2638, opened 2026-08-07
feat(facet-macros-impl): add `list(shape_type)` attr-grammar payload kind

### LINKED ISSUE #2637: Feature request: support a repeated/list attribute payload in `define_attr_grammar!`
Downstream motivation: [nbtx wants a per-field attribute that names a *set* of acceptable alternate types](https://github.com/bedrock-crustaceans/nbtx/issues/19) for lenient NBT tag-width decoding, e.g.

```rust
#[facet(nbtx::lenient_width(f32, i32))]
pub findable: i32,
```

That specific syntax — one attribute call whose payload is a comma-separated list of type tokens — doesn't parse against `define_attr_grammar!` today. Every variant payload kind the grammar currently supports (`VariantKind` in `facet-macros-impl/src/attr_grammar/make_parse_attr.rs`, roughly lines 315-363: `Unit`, `Newtype`, `NewtypeStr`, `NewtypeOptionChar`, `NewtypeI64`, `NewtypeUsize`, `Struct(Ident)`, `ArbitraryType`, `MakeT`, `Predicate`, `Validator`, `FnPtr`, `ShapeType`, `Arbitrary`, `OptionalStr`) is a *single* item, never a repetition — there's no `Vec<shape_type>`/`Vec<&'static str>`-style kind anywhere.

Concretely, in `dispatch_attr.rs`:
- The `Newtype`/`Arbitrary` kinds wrap the parenthesized tokens bodily in `Some(#inner)` (`dispatch_attr.rs` ~639-710, ~718-770) — `Some(f32, f64, i32)` doesn't compile, `Some` takes one argument.
- `shape_type` expands to `<#inner as ::facet::Facet>::SHAPE` (~907-962), which needs `#inner` to be a single type in type position — `f32, f64, i32` isn't valid there either.
- The `Struct(Ident)` kind *is* multi-item at the call site (`column(name = "...", primary_key)` sets two fields in one call), but `parse_field_type` (~764-782) hard-restricts a struct payload's field types to exactly five spellings — `bool`, `&'static str`, `Option<&'static str>`, `Option<bool>`, `Option<char>` — so it can't hold a `shape_type` field or a list either.

nbtx's own workaround (filed in the linked issue) is to declare a small fixed-field struct payload — `struct LenientWidth { i8: bool, i16: bool, i32: bool, i64: bool, f32: bool, f64: bool }` — which recovers the desired call-site syntax via bare-identifier-as-bool-flag, but only works because NBT happens to have a small closed set of numeric wire widths to enumerate as named fields ahead of time. That trick doesn't generalize to an open-ended list (e.g. "one of an arbitrary set of type names" or "an arbitrary set of string literals"), which seems like a fairly natural thing for an attribute grammar to want to express in general, not just for this one NBT use case.

**Ask:** a new grammar item kind along the lines of `List(shape_type)` (and maybe `List(&'static str)`/others, if that's not too much extra surface) that parses a comma-separated list inside the attribute's parens and generates a `&'static [&'static Shape]` (or similarly slice-typed) field, rather than requiring every payload to be exactly one item or a fixed pre-declared struct shape. Happy to be pointed at the right place in the macro if this is something a PR from outside would be workable for — the grammar/dispatch code is dense but the existing single-item kinds gave a decent map of where a new one would need to plug in (declaration parsing, `analyze_variant_payload`, the per-kind expression generator in `dispatch_attr.rs`, and whatever downstream code reads the resulting field back off `Shape`/`Field`/`Variant::attributes`).

theaddonn:
Fixes #2637.

`define_attr_grammar!` had single-item payload kinds for everything (`Newtype`, `ShapeType`, `Predicate`, etc.) but nothing for a repeated payload, so an attribute whose call-site value is a comma-separated list of types (nbtx's motivating use case) had no way to be declared.

This adds a new `list(shape_type)` variant-payload kind:

```rust
pub enum Attr {
    LenientWidth(list(shape_type)),
}
```

which generates `LenientWidth(&'static [&'static Shape])` and accepts a bare comma-separated list of types at the call site, matching the syntax requested in the issue:

```rust
#[facet(nbtx::lenient_width(f32, i32))]
```

## Design notes

- Unlike bare `shape_type`, which bypasses the generated enum and stores `&'static Shape` directly via `Attr::new_shape` (to support the `proxy_shape()` O(1) fast path), `list(shape_type)` is wrapped in the generated enum, following the same convention as `Newtype`/`ArbitraryType`/`FnPtr`/etc. This is simpler and required no changes to `facet-core`: `&'static [&'static Shape]` already implements `Facet` for free by composing the existing reference (`&'a T: Facet<'a>`) and slice (`[T]: Facet<'a>`) blanket impls with `Shape: Facet<'static>`.
- Both call-site shapes are handled: the derive macro's `#[facet(ns::attr(...))]` path (where `facet-macro-parse` has already stripped the outer parens by the time tokens reach the dispatcher) and a direct `__parse_attr!(attr(...))` invocation (where the parens survive).
- **Known limitation**, documented on `split_top_level_commas`: the list is split on top-level commas without tracking `<>` depth, since angle brackets aren't a real token-tree delimiter in `proc_macro2`. A comma inside a nested generic type in the list (e.g. `HashMap<K, V>`) would currently mis-split. This didn't seem worth the complexity for the motivating use case (a list of simple types), but flagging it explicitly in case it should block merge or is worth a follow-up.
- Scope: only `list(shape_type)` is implemented, not a fully generic `List(T)` combinator over every payload kind, per the issue's own framing of `shape_type` as the concrete ask.

## Testing

- New variant added to the `facet-testattrs` test grammar: `LenientWidth(list(shape_type))`.
- New integration test (`facet/tests/integration/extension_attr_list_shape_type.rs`) derives a struct with `#[facet(testattrs::lenient_width(f32, i32))]` and asserts the attribute decodes to the expected two-element `&[&Shape]`, including pointer-identity checks against `<f32 as Facet>::SHAPE` / `<i32 as Facet>::SHAPE`.
- `cargo test -p facet -p facet-testattrs -p facet-macros-impl` and `cargo clippy -p facet-macros-impl -p facet-testattrs --all-targets -- -D warnings` are clean.
- Manually verified error paths: empty list (`lenient_width()`), and malformed lists (leading/doubled/adjacent commas, e.g. `lenient_width(f32,,i32)`) all produce friendly `compile_error!`s rather than leaking through to raw rustc syntax errors.


theaddonn:
@coolbear-bot please review this pr

fasterthanlime:
> @coolbear-bot please review this pr

I'm afraid cool bear is not *that* automated yet :P

theaddonn:
awww sad, I cant prompt inject the bot to get my malicious code inside facet :(

fasterthanlime:
> awww sad, I cant prompt inject the bot to get my malicious code inside facet :(

well, I'm sure you can, just not that way.

I'm looking at it, griefs are:

  * no "unhappy path" test (we have 'slow' UI tests for compile errors)
  * no support for `<` / `>` generic types, unsynn has stuff for that,w e support it elsewhere. gonna see how feasible it is to support it day 1

theaddonn:
Thanks for the review! Both are fair and I think both are doable:

* **Unhappy-path test**: didn't realize the `slow-tests`-gated UI harness (`facet/tests/integration/compile_tests.rs` + `facet/tests/compile_tests/*.rs`) existed under that name when I looked — I'll add a case there for the "requires at least one type" / malformed-list error paths.
* **`<`/`>` generics in the list**: agreed this is worth doing day one rather than punting. Looks like `facet-macro-types::AngleTokenTree` / `VerbatimUntil<Comma>` (already used elsewhere, e.g. for field-type parsing) is exactly the right tool — `facet-macros-impl` already depends on `facet-macro-types`, it's just not pulled into `attr_grammar` yet. I'll swap my hand-rolled `split_top_level_commas` for that instead of tracking angle-bracket depth by hand, so `list(shape_type)` items like `HashMap<K, V>` split correctly.

Pushing an update shortly.

theaddonn:
Pushed both fixes:

* **Generic types in the list** now work correctly, including nested generics (`HashMap<u8, HashMap<u16, u32>>` etc.) — swapped the hand-rolled comma splitter for `facet_macro_types::VerbatimUntil<Comma>` parsed via `CommaDelimitedVec<ListShapeTypeItem>`, mirroring the exact pattern `facet-macros-impl/src/function/func_params.rs` already uses for function-parameter lists. Angle brackets aren't a real token-tree `Group`, so `VerbatimUntil`'s `AngleTokenTree` treats `<...>` as opaque and stops splitting at the right comma. Added a test exercising `#[facet(testattrs::lenient_width(f32, HashMap<u8, u8>))]` to prove the internal comma isn't mis-split.
* **Unhappy-path tests** added under the existing `slow-tests` UI harness (`facet/tests/integration/compile_tests.rs` + new fixtures in `facet/tests/compile_tests/`): one for the empty-list case (`attr()`), one for a malformed list (doubled comma, `attr(f32,, i32)`), each asserting the friendly `compile_error!` text actually shows up in `cargo build`'s stderr.

Existing error messages/behavior for the plain non-generic case are unchanged. All of `cargo test -p facet --test main`, `cargo test -p facet --features slow-tests --test main -- compile_tests`, and `cargo clippy -D warnings` are green.

fasterthanlime (approved):

PR #2641, opened 2026-08-09
fix(facet-value): Properly deserialize tuple structs in a way that can round-trip them

rlabrecque:
The `deserialize_value_into` dispatcher only routed `StructKind::Tuple` (native tuples like `(T0, T1)`) to `deserialize_tuple`, letting `StructKind::TupleStruct` (like `struct CName(Name)`) fall through to `deserialize_struct` which expects an object and rejects the array produced by serialization.

Added `test_roundtrip_tuple_struct` to cover both single-field and nested tuple structs.

Note that I'm a little surprised by this; seems like it would not be a terribly uncommon thing :? The new test does not function without the deserialize.rs change though -- and no other tests break sooooo this seems to work?

### DIFF
--- a/facet-value/src/deserialize.rs
+++ b/facet-value/src/deserialize.rs
@@ -338,7 +338,7 @@ fn deserialize_value_into<'facet>(
     // Priority 3: Check the Type for structs and enums
     match &shape.ty {
         Type::User(UserType::Struct(struct_def)) => {
-            if struct_def.kind == StructKind::Tuple {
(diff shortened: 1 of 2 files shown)


fasterthanlime:
> letting `StructKind::TupleStruct` (like `struct CName(Name)`) fall through to `deserialize_struct` which expects an object and rejects the array produced by serialization.

I'm not entirely surprised by this: the problem with those two tuple structs that you can't name the inside, so... I tend to avoid them.

Will review soon!

fasterthanlime (approved):

PR #2648, opened 2026-08-19
fix: enum variant renaming in flatten structs

### LINKED ISSUE #2239: Enum variant renaming does not work in flatten structs
**Reproduce**
```rs
use facet_pretty::FacetPretty;

#[derive(facet::Facet)]
struct User {
    user_id: String,
    #[facet(flatten)]
    banner: Banner,
}

#[derive(facet::Facet)]
#[facet(tag = "bannerType")]
#[repr(C)]
enum Banner {
    #[facet(rename = "color")]
    Color {
        #[facet(rename = "bannerColor")]
        color: String,
    },
    #[facet(rename = "customImage")]
    CustomImage {
        #[facet(rename = "bannerUrl")]
        url: String
    },
}

fn main() {
    let banner: Banner = facet_json::from_str(r#"{
        "bannerType": "color",
        "bannerColor": "123456"
    }"#).unwrap();
    println!("[direct] {}", banner.pretty());
    let u: User = facet_json::from_str(r#"{
        "user_id": "424242",
        "bannerType": "color",
        "bannerColor": "123456"
    }"#).unwrap();
    println!("[flatten] {}", u.pretty());
}

```

**Output**

```
[direct] Banner::Color {
  color: "123456",
}

thread 'main' (1186205) panicked at src/main.rs:36:10:
called `Result::unwrap()` on an `Err` value: DeserializeError { span: [53..60), path: banner, kind: invalid value: expected tag value 'Color', got 'color' }
```

Like use it directly work, but not within another struct via `#[facet(flatten)]`

**Expected**

```
[direct] Banner::Color {
  color: "123456",
}
[flatten] User {
  user_id: "424242",
  banner: Banner::Color {
    color: "123456",
  }
}
```

ruanpetterson:
Fix #2239

cc: @CyanChanges

### DIFF
--- a/facet-format/src/deserializer/path_navigator.rs
+++ b/facet-format/src/deserializer/path_navigator.rs
@@ -188,11 +188,20 @@ impl<'input, const BORROW: bool> PathNavigator<'input, BORROW> {
             );
 
             if current_path == vs_fields {
+                let effective_variant_name = match wip.shape().ty {
+                    facet_core::Type::User(facet_core::UserType::Enum(enum_def)) => enum_def
+                        .variants
+                        .iter()
+                        .find(|variant| variant.name == vs.variant_name)
+                        .map(facet_core::Variant::effective_name)
+                        .unwrap_or(vs.variant_name),
+                    _ => vs.variant_name,
+                };
                 trace!(
                     "open_segment: selecting variant '{}' at path {:?}",
-                    vs.variant_name, current_path
+                    effective_variant_name, current_path
                 );
-                wip = wip.select_variant_named(vs.variant_name)?;
+                wip = wip.select_variant_named(effective_variant_name)?;
                 break;
             }
         }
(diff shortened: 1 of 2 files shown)


facet-format/src/deserializer/path_navigator.rs
```diff
@@ -188,11 +188,20 @@ impl<'input, const BORROW: bool> PathNavigator<'input, BORROW> {
             );
 
             if current_path == vs_fields {
+                let effective_variant_name = match wip.shape().ty {
+                    facet_core::Type::User(facet_core::UserType::Enum(enum_def)) => enum_def
+                        .variants
+                        .iter()
+                        .find(|variant| variant.name == vs.variant_name)
```
fasterthanlime, line 195: not happy with the O(n) here, pondering whether it's worth merging as-is and optimize later with tables/type plans etc.

